PR #12, opened 2018-09-18
Fix service account application instructions

joeholley:


joeholley:
Changes from a WIP branch snuck in; will fix and re-submit PR.

PR #14, opened 2018-09-20
Update to reflect incomplete state of documenation

joeholley:


joeholley:
Pushed from wrong branch.  Need more coffee.

PR #26, opened 2018-10-03
Add redis auth and move connection to redishelpers

voidpirate:
Changes proposed

- Add support for redis authentication. Currently it assumes the keys for `redis.user` and `redis.password` are set in `matchmaker_config.json`. If this needs to be moved to look at env vars or similar, can do.

- Refactored redis connection pool to live in `redishelpers` (avoids duplication) and updated `backendapi`, `frontendapi`, and `mmforc` entry points to use this helper method.

- Misc clean up, such as; removed unreferenced packages and removed trailing whitespace.

### DIFF
--- a/cmd/backendapi/main.go
+++ b/cmd/backendapi/main.go
@@ -25,13 +25,12 @@ import (
 	"errors"
 	"os"
 	"os/signal"
-	"time"
 
 	"github.com/GoogleCloudPlatform/open-match/cmd/backendapi/apisrv"
 	"github.com/GoogleCloudPlatform/open-match/config"
 	"github.com/GoogleCloudPlatform/open-match/internal/metrics"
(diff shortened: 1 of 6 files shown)


joeholley:
Thanks a bunch for submitting a PR!  Nice dry code here.

In this case, we need to keep it relatively easy to have per-component pool settings.  Although all components are currently using the same pool settings, we expect them to have separate scaling requirements under load.  We'll need to preserve the ability to tweak the settings on a per-program basis once we've gotten feedback from production usage and run more comprehensive scale load tests.  

Maybe we could add some more changes to work with your approach?  The `matchmaker_config.json` file could hold the current used values as defaults, then we could add equivalent env vars to `config/config.go:envMappings` so these values could easily be overridden in the Kubernetes manifests on a per-pod basis?  Thoughts?

PR #27, opened 2018-10-03
Timestamp Indexing

calebatwd:
Yet another attempt to get my commit authorship accepted by the Google Cloud CLA.

Add timestamp indexing and de-indexing as a first class index for all match requests.

To use the timestamp index, match functions can us `ZRANGE` now to better segment filters. The short-term goal is to allow for the simplest way to run the same function multiple times at once.

I didn't debate the name too long, so we can come up with something more descriptive if there are good ideas. I didn't think starttime, qtime, createddate or any of the other db-toned names fit once you move into the match function context.

### DIFF
--- a/internal/statestorage/redis/playerq/playerq.go
+++ b/internal/statestorage/redis/playerq/playerq.go
@@ -20,6 +20,7 @@ package playerq
 import (
 	"encoding/json"
 	"strings"
+	"time"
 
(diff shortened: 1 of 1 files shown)


calebatwd:
We signed it?

joeholley:
Sorry for the CLA woes.  Approving!

PR #30, opened 2018-10-18
Update matchmaker_config from json to yaml

voidpirate:
Changes:
- Updated `matchmaker_config` to be defined in yaml, updated `matchmaker_config.json` symlinks to use yaml file and updated references in docs.
- Misc doc updates: trailing spaces removal, typos and removed note about adding redis auth since that was done in https://github.com/GoogleCloudPlatform/open-match/pull/26, I forgot to update the docs.

**This will also be a breaking change for anyone that has overridden default values in their configuration file.**


joeholley:
Some comments about this landed in the slack development channel, but basically: having the code support both would be the preference.  With Viper I think that's about as easy as just removing the 
`v.SetConfigType("json")`
[line without replacing it](https://github.com/spf13/viper#reading-config-files).  Looking at the source, if you don't specify a file type, it should just read the first file it finds at the given path with the given base filename and any supported extension:
[`// SupportedExts are universally supported extensions.
var SupportedExts = []string{"json", "toml", "yaml", "yml", "properties", "props", "prop", "hcl"}`](https://github.com/spf13/viper/blob/master/viper.go#L268)

PR #36, opened 2018-11-16
Fixing minor typo in apisrv.go

ReDucTor:


joeholley:
For this case, 'Open' is the name of the function, and has to be the first word of the comment in order to make the golang linters happy.  I'll reword it in v.0.2.0 to something clearer.

joeholley:
commit b57dd3e

PR #38, opened 2018-11-18
Add logging config to support json and level selection for logrus

calebatwd:
Adds a simple log config section in matchmaker_config.json to add level selection and format selection (text or json). This change also adds a logging package that's now being called from each of the components' inits to configure logging once when each process starts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ Example MMFs are provided in Golang and C#.
 
 ### Structured logging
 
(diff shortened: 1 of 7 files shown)


joeholley:
Yep looks good.  Will merge after we cut the v.0.2.0 release

joeholley:
This is in the 0.3.0 release

PR #39, opened 2018-11-20
020rc1 discussion

joeholley:
Discussion thread for changes in v0.2.0 RC1 with eyes on making RC2.


README.md
```diff
@@ -24,7 +28,7 @@ Open Match is designed to support massively concurrent matchmaking, and to be sc
 * **Component** &mdash; One of the discrete processes in an Open Match deployment. Open Match is composed of multiple scalable microservices called 'components'.
 * **Roster** &mdash; A list of all the players in a match.
 * **Profile** &mdash; The json blob containing all the parameters used to select which players go into a roster.
-* **Match Object** &mdash; A json blob to contain the results of the matchmaking function. Sent with an empty roster section to the backend API from your game backend and then returned with the matchmaking results filled in.
+* **Match Object** &mdash; A protobuffer message format that contains the Profile and the results of the matchmaking function. Sent to the backend API from yoru game backend with an empty roster  and then returned from your MMF with the matchmaking results filled in.
```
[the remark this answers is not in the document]
joeholley: Nice catch, thanks!

internal/statestorage/redis/playerq/playerq.go
```diff
@@ -119,43 +112,38 @@ func Delete(redisConn redis.Conn, playerID string) (err error) {
...
 	check(err, "")
 	return
 }
 
-// Unindex a player without deleting their JSON object representation from
-// state storage.
-func Unindex(redisConn redis.Conn, playerID string) (err error) {
-	results, err := Retrieve(redisConn, playerID)
+// Deindex a BLARG without deleting there JSON object representation from
+// state storage.  Unindexing is done in two stages: first the BLARG is added to an ignore list, which 'atomically' removes them from consideration. A Goroutine is then kicked off to 'lazily' remove them from any field indicies that contain them.
```
calebatwd: blarg blarg blarg

cmd/backendapi/apisrv/apisrv.go
```diff
@@ -115,112 +119,144 @@ func (s *backendAPI) CreateMatch(c context.Context, p *backend.Profile) (*backen
...
-			thisError = thisError + " - did the evaluator write a valid JSON match object?"
-			stats.Record(fnCtx, BeGrpcErrors.M(1))
-			return mo, errors.New(errString + thisError)
+			// ok is false if watchChan has been closed by redispb.Watcher()
+			newMO.Error = newMO.Error + "; channel closed - was the context cancelled?"
+		} else {
+			// 'ok' was true, so properties should contain the results from redis.
+			// Do basic error checking on the returned JSON
+			if !gjson.Valid(profile.Properties) {
+				newMO.Error = "retreived properties json was malformed"
```
calebatwd: The matchobject error gets override here. Is there a grpc oriented way of throwing an error instead?

calebatwd: Could also do the append like above.

cmd/backendapi/apisrv/apisrv.go
```diff
@@ -286,7 +328,7 @@ func (s *backendAPI) DeleteMatch(ctx context.Context, mo *backend.MatchObject) (
 		beLog.WithFields(log.Fields{
 			"error":     err.Error(),
 			"component": "statestorage",
-		}).Error("Statestorage error")
+		}).Error("State storage error")
 
 		stats.Record(fnCtx, BeGrpcErrors.M(1))
 		return &backend.Result{Success: false, Error: err.Error()}, err
```
calebatwd: Is there a valid way to cancel without triggering the error stat?

cmd/backendapi/apisrv/apisrv.go
```diff
@@ -305,7 +347,11 @@ func (s *backendAPI) DeleteMatch(ctx context.Context, mo *backend.MatchObject) (
 func (s *backendAPI) CreateAssignments(ctx context.Context, a *backend.Assignments) (*backend.Result, error) {
 
 	// TODO: make playerIDs a repeated protobuf message field and iterate over it
```
calebatwd: todo done? :)

cmd/mmlogicapi/apisrv/apisrv.go
```diff
@@ -0,0 +1,659 @@
...
+
+	if count == 0 {
+		err = errors.New("filter applies to no players")
+		mlLog.Error(err.Error())
+		return nil, err
+	} else if count > 500000 {
+		// 500,000 results is an arbitrary number; OM doesn't encourage
+		// patterns where MMFs look at this large of a pool.
+		mlLog.Warn("filter applies to too many players, ignoring")
+		return nil, nil
```
calebatwd: I feel like Warn should be Error if we don't intend on doing the thing that was asked for

cmd/mmlogicapi/apisrv/apisrv.go
```diff
@@ -0,0 +1,659 @@
...
+			out = out[:len(out)-1]
+		}
+	}
+
+	return out
+}
+
+// Functions for getting or setting player IDs to/from rosters
+// Probably should get moved to an internal module in a future version.
+func getPlayerIdsFromRoster(r *mmlogic.Roster) []string {
```
calebatwd: Helper inside internal?

api/protobuf-spec/mmlogic.proto
```diff
@@ -0,0 +1,74 @@
...
+  //        matched and how long the filters took to run, which will be sent out
+  //        the backend api along with your match results.
+  // OUTPUT: a Result message with a boolean success value and an error string
+  // if an error was encountered
+  rpc CreateProposal(messages.MatchObject) returns (messages.Result) {}
+
+  // Player listing and filtering functions
+  //
+  // RetrievePlayerPool gets the list of players that match every Filter in the
+  // PlayerPool, and then removes all players it finds in the ignore list.  It
```
calebatwd: , excluding players in any configured ignore lists.

`removes` threw me off until I read the implementation :) Sounded like it was taking action on the pool

api/protobuf-spec/backend.proto
```diff
@@ -1,76 +1,56 @@
...
+  //  - properties
+  //  - error. Empty if no error was encountered
+  //  - rosters, if you choose to fill them in your MMF. (Recommended)
+  //  - pools, if you used the MMLogicAPI in your MMF. (Recommended, and provides stats)
+  rpc CreateMatch(messages.MatchObject) returns (messages.MatchObject) {} 
   // Continually run MMF and stream matchobjects that fit this profile until
-  // client closes the connection.
-  rpc ListMatches(Profile) returns (stream MatchObject) {}
+  // client closes the connection.  Same inputs/outputs as CreateMatch.
+  rpc ListMatches(messages.MatchObject) returns (stream messages.MatchObject) {}
```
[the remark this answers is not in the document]
calebatwd: Otherwise it would cause some eyebrow raising when someone gets 200 match functions all hitting back at once even though it's expected :)

api/protobuf-spec/messages.proto
```diff
@@ -0,0 +1,90 @@
...
+// PlayerPools are defined by a set of 'hard' filters, and can be filled in
+// with the players that match those filters.
+//
+// PlayerPools contain a number of fields, but many gRPC calls that take a
+// PlayerPool as input only require a few of them to be filled in.  Check the
+// gRPC function in question for more details.
+message PlayerPool{
+    string name = 1;                // Arbitrary developer-chosen, human-readable string.
+    repeated Filter filters = 2;    // Filters are logical AND-ed (a player must match every filter).
+    Roster roster = 3;              // Roster of players that match all filters.
```
calebatwd: If this were a separate contract, it would clean things up quite a bit:
PlayerPoolDefinition {name, filters}
PlayerPoolPopulation {players, stats}

examples/backendclient/profiles/testprofile.json
```diff
@@ -1,14 +1,51 @@
...
+            { "name": "silverRanking", "attribute": "mmr.rating", "maxv": "1250", "minv": "950" },
+            { "name": "supportRole", "attribute": "role.support", "maxv": "2147483647" }
+          ]
+        }
+     ],
+     "rosters": [
+        {
+          "name": "red",
+          "players": [
+            { "pool": "defaultPool" },
```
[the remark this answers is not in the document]
calebatwd: After a few reads, this still confuses me a bit, but maybe it's just the example. I think there are three different concepts at play here:
1. A team config - a way to tell my custom match function how to organize and draft my teams
2. A roster - the players that end up in the team-shapes with any metadata they came with... maybe some team data as well
3. The players in the match - so that open match can put them on ignore lists automatically

I get the backfill perspective for something this prescriptive... like: "I already have some players, tell my match function to give a few more"... maybe there's an optimization for open match here, but could still just be custom function though?
**#3** seems like the important one for open match. I get that a developer may want to re-use what #3 is providing... which looks like a landing zone for sticking in teams, but then we have to teach open match what common ways developers may want to construct those teams... e.g. subteams (squads), conditional rule-teams.

Suggestions welcome!

README.md
```diff
@@ -187,28 +207,37 @@ Open Match is in active development - we would love your help in shaping its fut
...
+- [ ] Redis watch should be unified to watch a hash and stream updates.  The code for this is written and validated but not committed yet. We don't want to support two redis watcher code paths, so the backend watch of the match object should be switched to unify the way the frontend and backend watch keys.  The backend part of this is in but the frontend part is in another branch and will be committed later. 
+
+## Instrumentation / Metrics / Analytics
+- [ ] Instrumentation of MMFs is in the planning stages.  Since MMFs are by design meant to be completely customizable (to the point of allowing any process that can be packaged in a Docker container), metrics/stats will need to have an expected format and formalized outgoing pathway.  Currently the thought is that it might be that the metrics should be written to a particular key in statestorage in a format compatible with opencensus, and will be collected, aggreggated, and exported to Prometheus using another process.
+- [ ] [OpenCensus tracing](https://opencensus.io/core-concepts/tracing/) will be implemented in an upcoming version.
+- [ ] Read logrus logging configuration from matchmaker_config.json.
```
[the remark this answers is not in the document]
joeholley: Yep, will get it in with the final 0.2.0, and thanks!

README.md
```diff
@@ -187,28 +207,37 @@ Open Match is in active development - we would love your help in shaping its fut
...
+
+- [ ] Player/Group records generated when a client enters the matchmaking pool need to be removed after a certain amount of time with no activity. When using Redis, this will be implemented as a expiration on the player record.
+## CI / CD / Build
+- [ ] We plan to host 'official' docker images for all release versions of the core components in publicly available docker registries soon.
+- [ ] CI/CD for this repo and the associated status tags are planned.
+- [ ] Golang unit tests will be shipped in an upcoming version.
+- [ ] A full load-testing and e2e testing suite will be included in an upcoming version.
+
+## Will not Implement 
+- [X] Match profiles should be able to define multiple MMF container images to run, but this is not currently supported. This enables A/B testing and several other scenarios.
```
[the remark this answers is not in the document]
joeholley: LGTM, updating.

api/protobuf-spec/backend.proto
```diff
@@ -1,76 +1,56 @@
...
+  //  - properties
+  //  - [optional] roster, any fields you fill are available to your MMF.
+  //  - [optional] pools, any fields you fill are available to your MMF.
+  // OUTPUT: MatchObject message with these fields populated:
+  //  - id
+  //  - properties
+  //  - error. Empty if no error was encountered
+  //  - rosters, if you choose to fill them in your MMF. (Recommended)
+  //  - pools, if you used the MMLogicAPI in your MMF. (Recommended, and provides stats)
+  rpc CreateMatch(messages.MatchObject) returns (messages.MatchObject) {}
```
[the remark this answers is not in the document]
joeholley: In protobuf messages, an unpopulated field doesn't even exist in the serialized message, so I think it's less of a change than you might think.  Pretty much the only difference between something like 
`backend.Profile(id='123', properties:'{my:json}')`
and
`backend.MatchObject(id='123', properties:'{my:json}')`
is _actually just the name of the object_, from the standpoint of the code populating the object (in fact, as mentioned above, in most languages you could just do something like `type Profile = backend.MatchObject` and not even have to use the different message name in your calling code..  However, for the code receiving and manipulating the object (i.e. Open Match), it makes a significant difference, especially as we continue to 'promote' frequently used parts of the profile properties into message fields so they can be operated on by Open Match.

api/protobuf-spec/backend.proto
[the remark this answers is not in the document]
joeholley: If anything, that might be another reason to remove the evaluator sooner rather than later.  I'm not sure it really fits the paradigm as well once we've chosen not to allow that additional level of 'fan-out' within a profile (e.g. specifying multiple MMF images in a single profile).

api/protobuf-spec/backend.proto
```diff
@@ -1,76 +1,56 @@
...
-message ConnectionInfo{
-    string connection_string = 1;   // Passed by the matchmaker to game clients without modification. 
-}
-
-message Assignments{
-    Roster roster = 1;
-    ConnectionInfo connection_info = 2;
+  // INPUT: MatchObject message with the 'id' field populated. 
+  // (All other fields are ignored.)
+  rpc DeleteMatch(messages.MatchObject) returns (messages.Result) {}
```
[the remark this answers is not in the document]
joeholley: In proto3, all fields are 'optional' by default, and leaving a field empty basically costs you nothing (it's omitted completely from the serialized protobuf, and returned as an 'empty' field in the language-native format by the proto3 library of your language when it is accessed).

api/protobuf-spec/messages.proto
```diff
@@ -0,0 +1,90 @@
...
+// Data structure to hold details about a player
+message Player{
+  message Attribute{                
+    string name = 1;                // Name should match a Filter.attribute field. 
+    int64 value = 2;
+  }
+  string id = 1;                    // By convention, a UUID 
+  string properties = 2;            // By convention, a JSON-encoded string
+  string pool = 3;                  // Optionally used to specify the PlayerPool in which to find a player. 
+  repeated Attribute attributes= 4; // Attributes of this player.
```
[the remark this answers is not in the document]
joeholley: Not quite; unless someone writes a MMF where they put all the values from all the custom indexes, OM only currently populates the **players attribute values from the indexes that were queried to make a pool**, and even then only if using the GetPlayerPool() MMLogic API call.  For example: if your players were indexed on 12 attributes "A" through "L" when coming in the front door, but you call GetPlayerPool() and send it a pool with 3 filters: a value between 0 and 100 for "A", "D", and "F", you will receive a Roster of all the players that match, and each of those players will have an "A", "D", and "F" attribute, with the value the GetPlayerPool() call saw for that player when processing the filters.

api/protobuf-spec/mmlogic.proto
```diff
@@ -0,0 +1,74 @@
...
+  //        backend along with your error.
+  //      - [optional] rosters, anything you put here is returned to the
+  //        backend along with your error.
+  //      - [optional] pools, set to the output from the 'GetPlayerPools' call,
+  //        will populate the pools with stats about how many players the filters
+  //        matched and how long the filters took to run, which will be sent out
+  //        the backend api along with your match results.
+  // OUTPUT: a Result message with a boolean success value and an error string
+  // if an error was encountered
+  rpc CreateProposal(messages.MatchObject) returns (messages.Result) {}
```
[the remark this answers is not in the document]
joeholley: Yeah, this is a good contract.  PR?

api/protobuf-spec/mmlogic.proto
```diff
@@ -0,0 +1,74 @@
...
+
+  // Player listing and filtering functions
+  //
+  // RetrievePlayerPool gets the list of players that match every Filter in the
+  // PlayerPool, and then removes all players it finds in the ignore list.  It
+  // combines the results, and returns the resulting player pool.
+  rpc GetPlayerPool(messages.PlayerPool) returns (stream messages.PlayerPool) {}
```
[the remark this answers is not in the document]
joeholley: That's not correct; each time you call it it will start streaming from page 1 again.  You need to stay connected to it until it closes the stream to get all the pages.

To be clear: this is not paging because people only want partial results sometimes, this is paging because you cannot send all the results in one message (it is basically just chunking the response).  Proper usage should always stay connected to get all the pages every time it is called. (I can make that clearer in the documentation)

cmd/mmforc/main.go
```diff
@@ -242,110 +240,91 @@ func main() {
...
 	}
+	mmfuncLog.Info("Attempting to create mmf k8s job")
+
+	// Kick off k8s job
+	envvars := []apiv1.EnvVar{
+		{Name: "MMF_PROFILE_ID", Value: profID},
+		{Name: "MMF_PROPOSAL_ID", Value: propID},
+		{Name: "MMF_REQUEST_ID", Value: moID},
+		{Name: "MMF_ERROR_ID", Value: resultsID},
+		{Name: "MMF_TIMESTAMP", Value: timestamp},
```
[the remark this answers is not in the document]
joeholley: :D

internal/statestorage/redis/playerq/playerq.go
joeholley: Yep, dev code, fixing.  Nice catch.

cmd/backendapi/apisrv/apisrv.go
joeholley: Hrm.  At that point maybe better to make Error a repeated string message field.  I'd rather people be able to 'loop' through the errors, trying to identify an atomic string that has a specific failure meaning than having to do do some kind of fuzzy/substring matching within a concatenated string. This could also allow us to raise errors from further down the stack? 
 Thoughts?

cmd/backendapi/apisrv/apisrv.go
joeholley: nice catch

cmd/backendapi/apisrv/apisrv.go
joeholley: I'm not sure that cancelling a deletion is a case we are expecting?  Can you elaborate a bit?

cmd/mmlogicapi/apisrv/apisrv.go
joeholley: Yep.

cmd/mmlogicapi/apisrv/apisrv.go
joeholley: Probably a 'companion' golang source file for working with protobuf messages that lives in internal/pb but isn't auto-generated from protoc, I think.

api/protobuf-spec/messages.proto
joeholley: We're going to need to discuss this further.  I'll hit you up in chat for now and we can take it back to a public thread once we've firmed up the approach.

RELEASE.md
```diff
@@ -0,0 +1,30 @@
...
+
+
+### Roadmap 
+ - It has become clear from talking to multiple users that the software they write to talk to the Backend API needs a name.  'Backend API Client' is technically correct, but given how many APIs are in Open Match and the overwhelming use of 'Client' to refer to a Game Client in the industry, we're currently calling this a 'Director', as its primary purpose is to 'direct' which profiles are sent to the backend, and 'direct' the resulting MatchObjects to game servers.  Further discussion / suggestions are welcome.
+ - We'll be entering the design stage on longer-running MMFs before the end of the year. We'll get a proposal together and on the github repo as a request for comments, so please keep your eye out for that.
+ - Match profiles providing multiple MMFs to run isn't planned anymore.  Just send multiple copies of the profile with different MMFs specified via the backendapi. 
+ - Evaluators will be examined for removal in an upcoming version. There's a good change this logic should live in the Directory
```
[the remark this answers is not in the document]
joeholley: Yeah, interested in other user's expected workflows too.  One of the advantages of moving the Evaluator logic into the Director is that the Director may have additional information about the backend (where free servers are, or the size of pools other than the ones considered for a particular match) that an evaluator may not have easy access to.

api/protobuf-spec/backend.proto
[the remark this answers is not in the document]
calebatwd: At some point you've to fan-in though? Unless you can guarantee that all match functions create deterministically different sets? I guess you could not care that players are in multiple matches, and that a dev has some kind of build a match, then continuously munge it type of workflow? Some kind of supported fan-in seems useful. Maybe we can break that out from the evaluator concept.

cmd/backendapi/apisrv/apisrv.go
calebatwd: Is errors reserved for open match errors or is it available to match functions?
Either way, having errors as a repeated string probably makes more sense. I know in C# There's innerexceptions, but i don't think we need nearly that much wrapping here to get the errors communicated.

cmd/backendapi/apisrv/apisrv.go
calebatwd: Ah, this is delete. Nevermind, totally get it!

cmd/mmlogicapi/apisrv/apisrv.go
```diff
@@ -438,8 +433,13 @@ func (s *mmlogicAPI) applyFilter(c context.Context, filter *mmlogic.Filter) (map
 	} else if count > 500000 {
 		// 500,000 results is an arbitrary number; OM doesn't encourage
```
calebatwd: Do you think capping is better than not doing it at all?

joeholley: I did a couple different approaches to this during development, and I think from the perspective of OM being a tool used by matchmakers, that most of the time you'd prefer to know you had all the results or nothing? Otherwise the MMF might make a match that it would otherwise reject as not high enough quality if it had the whole matching player pool?  Ultimately I think that we just need to make it clear that folks should be constructing pools that they can fully retrieve, which ultimate is a subjective evaluation about how long you're willing to wait?  Like, you can actually retrieve a pool of this size just fine, it just takes several seconds.  If your matchmaking is okay with waiting that long, you could just adjust this number higher or even comment this branch out entirely?  We just don't want a situation where OM lets you 'accidentally' shoot yourself in the foot WRT performance or incomplete filter results.  Also compounded as either of those situations would be easy for inexperienced users to misinterpret as 'Open Match is broken'.  Thoughts?

PR #47, opened 2018-12-15
Update to use Xid instead of UUID

### LINKED ISSUE #32: Evaluate switching various object IDs from UUIDv4 to xid
https://github.com/rs/xid

This is an internal representation thing - shouldn't have any impact outside of the core components, or in any custom functions/evaluators/backend client/frontend clients.  Will drop the string size for the unique IDs from 36 characters to 20 characters, a pretty significant memory savings in Redis.  We'll also be passing around less data this way, and it should help with the 64-character Job name limit in Kubernetes when kicking off evaluators and MMFs as well.

scrypter:
close #32

### DIFF
--- a/api/protobuf-spec/frontend.proto
+++ b/api/protobuf-spec/frontend.proto
@@ -19,5 +19,5 @@ message Group{
 }
 
 message PlayerId {
-    string id = 1;          // By convention, a UUID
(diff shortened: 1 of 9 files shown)


joeholley:
I'll merge this into the 0.3.0 RC in the next couple days.

PR #55, opened 2018-12-30
v0.3.0 RC2 discussion

joeholley:
Discussion thread for changes in v0.3.0 RC1 with eyes on making RC2.

Updates available in the [change log](https://github.com/GoogleCloudPlatform/open-match/blob/030wip/CHANGELOG.md).


cmd/frontendapi/apisrv/apisrv.go
```diff
@@ -152,149 +130,107 @@ func (s *frontendAPI) DeleteRequest(c context.Context, g *frontend.Group) (*fron
...
+			feLog.WithFields(log.Fields{
+				"assignment": a.Assignment,
+				"playerid":   a.Id,
+				"status":     a.Status,
+				"error":      a.Error,
+			}).Info("updating client")
+			assignmentStream.Send(&a)
+			stats.Record(fnCtx, FeGrpcStreamedResponses.M(1))
+			// Reset timeout.
+			timeoutChan = time.After(time.Duration(s.cfg.GetInt("api.frontend.timeout")) * time.Second)
```
calebatwd: Behavior callout. Overuse of the status field could mean infinitely connected clients. Do we want an actual _disconnect players after x minutes no matter what_ timeout in addition to the _touched_ timeout?

calebatwd: It's kind of a protect you from yourself sort of thing, so maybe not.

cmd/mmlogicapi/apisrv/apisrv.go
```diff
@@ -166,7 +165,7 @@ func (s *mmlogicAPI) CreateProposal(c context.Context, prop *mmlogic.MatchObject
 	}
 
 	// Write all non-id fields from the protobuf message to state storage.
-	err := redispb.MarshalToRedis(c, prop, s.pool)
+	err := redispb.MarshalToRedis(c, s.pool, prop, s.cfg.GetInt("redis.expirations.matchobject"))
```
calebatwd: A pattern I've noticed from the redispb signatures... by sticking with the wire-format as the data model, any augmentations or decorations to those models going down to the database require parameter-centric signatures. If we end up with more than 4 or 5 parameters, it may be worth investigating a proper datamodel pattern. Cross that bridge when we come to it.

internal/statestorage/redis/playerindices/playerindices.go
```diff
@@ -0,0 +1,273 @@
...
+//     "maps.sunsetvalley",
+//     "mmr.rating",
+//   ]
+//
+// For now, OM reads your 'config/matchmaker_config.(json|yaml)' file for a
+// list of indices, which it monitors using the golang module Viper
+// (https://github.com/spf13/viper).
+// In a full deployment, it is expected that you don't manage the config file
+// directly, but instead put the contents of that file into a Kubernetes
+// ConfigMap.  Kubernetes will write those contents to a file inside your
```
calebatwd: Managing live indices via kubernetes configmaps makes sense to start, and at least it avoids requiring another data story. I imagine we'll want some kind of live-configiness native to open match that doesn't rely on routing configs through kubernetes, but I think we can come back around to that when we need it.

internal/statestorage/redis/playerindices/playerindices.go
```diff
@@ -0,0 +1,273 @@
...
+			//v := gjson.Get(player.Properties, regexp.QuoteMeta(attribute))
+			v := gjson.Get(player.Properties, attribute)
+
+			// If this attribue wasn't provided in the JSON, continue to the
+			// next attribute to index.
+			if !v.Exists() {
+				iLog.WithFields(log.Fields{"attribute": attribute, "value": v.Raw}).Debug("Couldn't find index in JSON: ", player.Properties)
+				continue
+			} else if -9223372036854775808 <= v.Int() && v.Int() <= 9223372036854775807 {
+				// value contains a valid unsigned 64-bit integer
```
calebatwd: My gjson + golang types is a little rough, but would this be a signed int? I definitely think it makes sense as a signed int64 and not a uint64, so the code is right, but the comment doesn't match.

internal/statestorage/redis/redispb/matchobject.go
```diff
@@ -0,0 +1,143 @@
...
+	if len(pbMap) == 0 {
+		return errors.New("matchobject key does not exist")
+	}
+
+	// Put values from redis into the MatchObject message
+	pb.Error = pbMap["error"]
+	pb.Properties = pbMap["properties"]
+
+	// TODO: Room for improvement here.
+	if j := pbMap["pools"]; j != "" {
```
calebatwd: 👍

internal/statestorage/redis/redispb/matchobject.go
```diff
@@ -0,0 +1,143 @@
...
+				// Cleanup
+				close(watchChan)
+				return
+			default:
+				//results, err = Retrieve(ctx, pool, key)
+				results = om_messages.MatchObject{Id: pb.Id}
+				err = UnmarshalFromRedis(ctx, pool, &results)
+				if err != nil {
+					moLog.Debug("No new results")
+					time.Sleep(2 * time.Second) // TODO: exp bo + jitter
```
calebatwd: If we make this a `rate` parameter or a `rateChan` parameter, then the caller can describe the jitter and frequency for the queries.

internal/statestorage/redis/redispb/player.go
```diff
@@ -0,0 +1,177 @@
...
+				// Player stopped asking for updates; clean up
+				close(watchChan)
+				return
+			default:
+				// Update the player's 'accessed' timestamp to denote they haven't disappeared
+				err := playerindices.Touch(ctx, pool, pb.Id)
+				if err != nil {
+					// Not fatal, but this error should be addressed.  This could
+					// cause the player to expire while still actively connected!
+					pwLog.Error("State storage error:", err.Error())
```
calebatwd: Nit. "State storage error" in the playerWatcher doesn't differential well from some of the redispb errors

internal/statestorage/redis/redispb/player.go
```diff
@@ -0,0 +1,177 @@
...
+				// latest version in string form so we can compare it easily in
+				// future loops.
+				//
+				// If we decide to watch other message fields for updates,
+				// they will need to be added here.
+				//
+				// This can be made much cleaner if protobuffer reflection improves.
+				curResults := fmt.Sprintf("%v%v%v", results.Assignment, results.Status, results.Error)
+				if prevResults == curResults {
+					pwLog.Debug("No new results, backing off")
```
calebatwd: Nit. "backing off" implies some kind of change in the wait period. I would be confused if I saw this in the logs. Maybe. "No new watcher results".

internal/statestorage/redis/redispb/player.go
```diff
@@ -0,0 +1,177 @@
...
+
+// PlayerWatcher makes a channel and returns it immediately.  It also launches an
+// asynchronous goroutine that watches a redis key and returns updates to
+// that key on the channel.
+//
+// The pattern for this function is from 'Go Concurrency Patterns', it is a function
+// that wraps a closure goroutine, and returns a channel.
+// reference: https://talks.golang.org/2012/concurrency.slide#25
+//
+// NOTE: this function will never stop querying Redis during normal operation! You need to
```
calebatwd: I suppose there's an underlying behavior in that the available redis pool could potentially bottleneck watchers if concurrency gets too high. Curious what that might look like.

internal/statestorage/redis/redispb/shared.go
```diff
@@ -104,7 +103,12 @@ func MarshalToRedis(ctx context.Context, pb proto.Message, pool *redis.Pool) (er
 		//  something like parseTag() in src/encoding/json/tags.go
 		//field := strings.ToLower(pbInfo.Type().Field(i).Tag.Get("json"))
 		field := strings.ToLower(pbInfo.Type().Field(i).Name)
-		value := gjson.Get(jsonMsg, field)
+		value := ""
+		//value, err = strconv.Unquote(gjson.Get(jsonMsg, field).String())
+		value = gjson.Get(jsonMsg, field).String()
```
calebatwd: I've noticed a lot of things stop working when the json field keys are of varying case. Is there a reason to restrict everything too lower?

internal/statestorage/redis/playerindices/playerindices.go
```diff
@@ -0,0 +1,273 @@
...
+	diLog := piLog.WithFields(log.Fields{"playerID": playerID})
+
+	// Connect to redis
+	redisConn := rPool.Get()
+	defer redisConn.Close()
+
+	// Get the list of indices to delete
+	indices, err := Retrieve(cfg)
+	// Look for previously configured indices
+	indices = append(indices, RetrievePrevious(cfg)...)
```
calebatwd: cool

cmd/frontendapi/apisrv/apisrv.go
```diff
@@ -152,149 +130,107 @@ func (s *frontendAPI) DeleteRequest(c context.Context, g *frontend.Group) (*fron
...
+				"error":     err.Error(),
+				"component": "statestorage",
+				"playerid":  p.Id,
+			}).Error("State storage error")
+
+			// Count errors for metrics
+			errTag, _ := tag.NewKey("errtype")
+			fnCtx, _ := tag.New(ctx, tag.Insert(errTag, "watch_timeout"))
+			stats.Record(fnCtx, FeGrpcErrors.M(1))
+			return err
```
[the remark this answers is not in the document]
joeholley: Yeah, definitely doable.  Left a note in the code until we decide and solidify some error states we want to reply with

internal/statestorage/redis/redispb/player.go
joeholley: Good point, updated with better message

internal/statestorage/redis/redispb/player.go
joeholley: Yeah, possible. My reading of the redigo docs leads me to believe that the pool.Get() should block until a connection is available.  If that becomes the case, there are a couple of approaches, but probably the best choice for these high-load 'watch' processes would be a read-only redis slave and a new k8s service name that points at it, and then coding the ability to have the watch processes use that slave while everything else uses the master.

internal/statestorage/redis/redispb/matchobject.go
joeholley: [Issue 59](https://github.com/GoogleCloudPlatform/open-match/issues/59)

internal/statestorage/redis/redispb/shared.go
joeholley: It's not ideal but probably okay to leave it this way until post 1.0 or someone is sufficiently motivated to take a stab at another approach

joeholley:
Okay, updates from @calebatwd 's review are in, this is now considered RC2.  Plan to cut the release Friday unless we hit a showstopper.

joeholley:
Looks good, releasing

PR #70, opened 2019-02-05
Fix build and test breakages

jeremyje:
Fixing the build, vet, fmt errors in open-match.

### DIFF
--- a/internal/statestorage/redis/ignorelist/ignorelist_test.go
+++ b/internal/statestorage/redis/ignorelist/ignorelist_test.go
@@ -14,6 +14,6 @@ func TestCreate(t *testing.T) {
 	redisConn.GenericCommand("ZADD").Expect("ok")
 	err := Create(redisConn, ilid, pids)
 	if err != nil {
-		t.Error(err)
+		t.Fatal(err)
 	}
 }
--- a/internal/statestorage/redis/ignorelist/ignorelist_test.go
+++ b/internal/statestorage/redis/ignorelist/ignorelist_test.go
@@ -6,12 +6,14 @@ import (
 	"github.com/rafaeljusto/redigomock"
 )
 
-func TestCreate(*testing.T) {
-	pids := "test1 test2 test3"
+func TestCreate(t *testing.T) {
+	pids := []string{"test1", "test2", " test3"}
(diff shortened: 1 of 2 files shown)


internal/statestorage/redis/ignorelist/ignorelist_test.go
```diff
@@ -6,12 +6,14 @@ import (
...
+func TestCreate(t *testing.T) {
+	pids := []string{"test1", "test2", " test3"}
 	ilid := "testil"
 	redisConn := redigomock.NewConn()
+
+	redisConn.GenericCommand("ZADD").Expect("ok")
 	err := Create(redisConn, ilid, pids)
 	if err != nil {
-		panic(err.Error())
+		t.Error(err)
```
jkowalski: t.Fatal() ?

test/cmd/frontendclient/main.go
```diff
@@ -194,7 +194,7 @@ func waitForResults(resultsChan chan frontend.Player, stream frontend.Frontend_G
 				break
 			}
 			log.Println("Error encountered")
-			log.Println("Error reading stream for GetAssignments(_) = _, %v", err)
+			log.Printf("Error reading stream for GetAssignments(_) = _, %v\n", err)
```
jkowalski: nit: log.Printf() does not need a trailing newline

test/cmd/frontendclient/main.go
```diff
@@ -194,7 +194,7 @@ func waitForResults(resultsChan chan frontend.Player, stream frontend.Frontend_G
 				break
 			}
 			log.Println("Error encountered")
-			log.Println("Error reading stream for GetAssignments(_) = _, %v", err)
+			log.Printf("Error reading stream for GetAssignments(_) = _, %v", err)
```
calebatwd: nit/curious. I haven't seen placeholders like the "*_*" before to visualize the input and right-side? Is that a standard pattern/does it work with the logging library in some interesting way? (I realize it's carry-over from the previous code... just curious :) )

jeremyje: Not sure probably not worth looking into just because this code is probably going to die.

PR #71, opened 2019-02-08
Update git ignore to ignore goland ide files

calebatwd:
Goland is a free golang ide from jetbrains. This prevents metadata files from accidentally getting checked into source.


jeremyje:
LGTM

PR #72, opened 2019-02-08
Log noise reduction and unmarshall err swallowing bug

calebatwd:
This change reduces the level of several state-storage related logs.
It also fixes a small error where the redis key for a given matchobject to unmarshal from the db gets swallowed. Not finding the key is expected, however the result is a wrapping log line that simply reads "No new results," without logging the key the log was looking for.


internal/statestorage/redis/redispb/matchobject.go
```diff
@@ -129,7 +131,7 @@ func Watcher(ctx context.Context, pool *redis.Pool, pb om_messages.MatchObject)
 				results = om_messages.MatchObject{Id: pb.Id}
 				err = UnmarshalFromRedis(ctx, pool, &results)
 				if err != nil {
-					moLog.Debug("No new results")
+					moLog.Debug("No new results ", err)
```
jeremyje: Would this be Debugf("... %s")?

calebatwd: I'll keep that in mind. I know there's a bunch of log format improvements to chase down.

calebatwd: Fixed :)

joeholley:
To keep with the format of the other logs, errors should be put as 'fields' in the logrus logging format.  Should help us categorize and count errors in logs.  So, something like this:
```
moLog.WithFields(log.Fields{
  "error": err.Error(),
}).Debug("no new results")
```

jeremyje:
This change conflicts with the cmd/ move PR.

jeremyje:
Ping, there's still conflicts with this PR against master.

jeremyje:
Discussed on Slack @Laremere will fix this in a different PR.

PR #79, opened 2019-02-20
Move application logic from cmd/ to internal/app/

jeremyje:
Moving application logic to conform to https://github.com/golang-standards/project-layout.
This is a first stage refactor to consolidate duplicate logic.


ihrankouski:
Btw it feels to me that we'd better move `internal/pb` to be `pkg/pb` - to allow others just import that when writing OM client apps (even MMFs) in golang, what do you think?

jeremyje:
> Btw it feels to me that we'd better move `internal/pb` to be `pkg/pb` - to allow others just import that when writing OM client apps (even MMFs) in golang, what do you think?

That could be done in a different PR. I'm not 100% sure if OM is ready to have a semi-stable API yet. I think @joeholley should make that call. If yes, I can do that next.

joeholley:
> Btw it feels to me that we'd better move `internal/pb` to be `pkg/pb` - to allow others just import that when writing OM client apps (even MMFs) in golang, what do you think?

Probably doable using API versioning ala kubernetes?

PR #81, opened 2019-02-21
Helm charts for open-match

jeremyje:


jeremyje:
One thing about the Helm chart is that Prometheus is running but is not configured yet. I have pointers in the code to add that later.

install/helm/open-match/config/matchmaker_config.yaml
```diff
@@ -0,0 +1,98 @@
...
+    backoff: "[2 32] *2 ~0.33 <30"
+  frontend: 
+    hostname: om-frontendapi
+    port: 50504
+    backoff: "[2 32] *2 ~0.33 <300"
+  mmlogic: 
+    hostname: om-mmlogicapi
+    port: 50503
+  functions:
+    port: 50502
```
calebatwd: Interesting, I don't recall seeing 50502 assignments anywhere. I know knative defaults to 8080, which is what we're using as well.

jeremyje: This config is a copy paste of the the matchmaker_config.yaml in other parts of the repository.

jeremyje:
Ping @joeholley for review.

PR #82, opened 2019-02-21
Update .gcloudignore to include .gitignore's filters so that Cloud Bu…

jeremyje:
…ild packages don't upload binaries.

### DIFF
--- a/.gcloudignore
+++ b/.gcloudignore
@@ -12,10 +12,4 @@
 # below:
 .git
 .gitignore
-
-# Don't include binaries in the build.
-cmd/backendapi/backendapi
-cmd/frontendapi/frontendapi
-cmd/mmforc/mmforc
-cmd/mmlogicapi/mmlogicapi
-examples/backendclient/backendclient
+#!include:.gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,7 @@
 populations
 
 # local config files
-*.json
+#*.json
 
 # Discarded code snippets
 build.sh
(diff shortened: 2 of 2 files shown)


ihrankouski:
One thing regarding this is that current .gitignore has dangerous line: https://github.com/GoogleCloudPlatform/open-match/blob/040wip/.gitignore#L28
This may lead to broken images in theory (and I recall it was the case once in the past)

jeremyje:
> One thing regarding this is that current .gitignore has dangerous line: https://github.com/GoogleCloudPlatform/open-match/blob/040wip/.gitignore#L28
> In future this may lead to broken images in theory (and I recall it was the case once in the past).

That's strange why .json files would get excluded but I did remove all of the json configs in https://github.com/GoogleCloudPlatform/open-match/commit/1847f795719460da17eb8552d36f1e7d9178296a

joeholley:
> > One thing regarding this is that current .gitignore has dangerous line: https://github.com/GoogleCloudPlatform/open-match/blob/040wip/.gitignore#L28
> > In future this may lead to broken images in theory (and I recall it was the case once in the past).
> 
> That's strange why .json files would get excluded but I did remove all of the json configs in [1847f79](https://github.com/GoogleCloudPlatform/open-match/commit/1847f795719460da17eb8552d36f1e7d9178296a)

This is so that developers can maintain local copies of all the build and deploy `json` files without concern about them getting included in commits, since almost every file with a `json` or `yaml` extension requires updating with developer-specific registries and image names when doing local development.  Open to suggestions but needs a reasonably easy solution.

jeremyje:
I think the ideal solution here is to have some presubmit (Cloud Build) script that would verify that files are good.
We could also change the deployment system to have a designated experimental deployment directory and ignore the files in there.

How feasible is this?

jeremyje:
I got bit by the *.json ignore rule. Is there specific directories that we can reduce this filter to? I'd propose we remove *.json and see what specific directories/files are problematic and filter those out specifically.

PR #93, opened 2019-03-01
Add Prometheus service discovery annotations to the Open Match servers.

jeremyje:
This touches on https://github.com/GoogleCloudPlatform/open-match/issues/46 where we should move towards Prometheus service discovery.

### DIFF
--- a/install/yaml/02-open-match.yaml
+++ b/install/yaml/02-open-match.yaml
@@ -26,6 +26,10 @@ spec:
       app: openmatch
       component: backend
   template:
+    annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/port: 9555
+        prometheus.io/path: "/metrics"
     metadata:
       labels:
         app: openmatch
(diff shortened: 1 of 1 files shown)


joeholley:
I'm not exactly sure what this does without reading through all the documentation, could you provide a quick summary?

jeremyje:
> I'm not exactly sure what this does without reading through all the documentation, could you provide a quick summary?

Prometheus by default will scan Kubernetes for pods that it can scape. These annotations are what it looks for to determine it should scrape data. 

https://www.weave.works/docs/cloud/latest/tasks/monitor/configuration-k8s/#per-pod-prometheus-annotations

After that simply deploying a prometheus instance to your cluster allows it to collect metrics data.

joeholley:
I'm pretty sure this is handled by the coreOS Prometheus Operator we're using, and it's [ServiceMonitor resource](https://coreos.com/tectonic/docs/latest/tectonic-prometheus-operator/user-guides/application-monitoring.html)?
[](https://github.com/GoogleCloudPlatform/open-match/blob/4ed6d275a3c21cc98592dc082251fcc51b45996c/install/yaml/03-prometheus.yaml#L2-L18)

jeremyje:
I'm proposing that we support regular prometheus as well because it's much simpler to deploy and avoids using CRDs which have this ordering issue.

PR #96, opened 2019-03-04
Remote executable bit from files that are not executable.

jeremyje:


jeremyje:
> I think the two protoc scripts are probably supposed to be executable, but the rest all looks fine?

Sorry about that. Also these files are sorta obsolete now with the Makefile.

PR #98, opened 2019-03-04
Update the Dockerfiles to build completely from local source and to CI.

jeremyje:
With this change you'll need to build from the root of the repository so all the code is in context.

```bash
docker build -f Dockerfile.base-build -t open-match-base-build .
docker build -f cmd/frontendapi/Dockerfile -t gcr.io/$PROJECT/open-match-frontendapi:0.4.0 .
```

Example Build: https://console.cloud.google.com/cloud-build/builds/ed7c33d9-d082-4382-83e4-463190b832f5?project=open-match-build


cmd/backendapi/Dockerfile
```diff
@@ -1,14 +1,6 @@
...
-# TODO: Set these to the final branch name after RC approval.
-RUN svn export https://github.com/GoogleCloudPlatform/open-match/branches/040wip/internal
-RUN svn export https://github.com/GoogleCloudPlatform/open-match/branches/040wip/config
-WORKDIR /go/src/github.com/GoogleCloudPlatform/open-match/cmd/backendapi
-COPY . .
-RUN go get -d -v
+FROM open-match-base-build as builder
+
+WORKDIR /go/src/github.com/GoogleCloudPlatform/open-match/cmd/backendapi/
 RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo .
```
ihrankouski: When developing locally, is there any way to avoid re-building of two images (base & cmd) without interventions to Dockerfile? Because all the logic is under `/internal/` now...

jeremyje: When vendoring/go mod is introduced we can avoid it because the base image is just initializing and downloading the dependencies. And Open Match has a lot of dependencies. If you look at the base Dockerfile that's all it's doing.

Pulling the dependencies for each image (there's 10 of them) in Cloud Build takes so long that it was timing out (10 minutes). It was also flaky because the script was downloading all the dependencies for Open Match at one point more then 20 times! (10 docker, 10 binaries, 1 for test).

One thing I don't think we can get rid of is where the context of the Dockerfile's build starts. It'll need to start at the root of the repository, not at the main file. The reason is there's the dependencies, the main file, and files in other directories like internal/, and later on pkg/.

jeremyje: I forgot to mention there is a way to build the Dockerfile without the whole repository context. You can prebuild the binary first and just copy that over so there's no building in the Dockerfile. This is pretty common practice as well and might be better since it aligns well with Skaffold.

I do have a branch where I take this approach but it's a major change that would take quite a few PRs and some buy in to get in. Example: https://github.com/jeremyje/open-match/blob/buildsys/cmd/backendapi/Dockerfile

jeremyje: Moving this discussion to a proposal bug, https://github.com/GoogleCloudPlatform/open-match/issues/111

cloudbuild.yaml
```diff
@@ -43,62 +43,64 @@
...
+  waitFor: ['Docker Image: open-match-base-build']
+
+- id: 'Docker Image: Evaluator Simple'
+  name: gcr.io/cloud-builders/docker
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/openmatch-evaluator-simple:${_OM_VERSION}-${SHORT_SHA}', 'examples/evaluators/golang/simple']
+  waitFor: ['Docker Image: open-match-base-build']
+
+- id: 'Docker Image: openmatch-cs-func-simple'
+  name: gcr.io/cloud-builders/docker
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/openmatch-cs-func-simple:${_OM_VERSION}-${SHORT_SHA}', 'examples/functions/golang/manual-simple']
```
sawagh: You mean: 'examples/functions/csharp/simple'

cloudbuild.yaml
```diff
@@ -43,62 +43,64 @@
...
+  waitFor: ['Docker Image: open-match-base-build']
+
+- id: 'Docker Image: openmatch-go-func-simple'
+  name: gcr.io/cloud-builders/docker
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/openmatch-go-func-simple:${_OM_VERSION}-${SHORT_SHA}', 'examples/functions/golang/manual-simple']
+  waitFor: ['Docker Image: open-match-base-build']
+
```
sawagh: Does building mechanism for php / python function containers stay the same? Dont see them included in the steps here.

cmd/mmforc/Dockerfile
```diff
@@ -1,28 +1,9 @@
...
-RUN svn export https://github.com/GoogleCloudPlatform/open-match/branches/040wip/config
+FROM open-match-base-build as builder
 
 WORKDIR /go/src/github.com/GoogleCloudPlatform/open-match/cmd/mmforc/
-COPY . .
-RUN go get -d -v 
 RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo .
 
-FROM gcr.io/distroless/static
+FROM gcr.io/distroless/static
```
sawagh: super nit :) Extra space at line end? (Likewise for the rest of the places this has been added in this file)

examples/evaluators/golang/simple/Dockerfile
```diff
@@ -1,11 +1,10 @@
...
+FROM open-match-base-build as builder
+
+WORKDIR /go/src/github.com/GoogleCloudPlatform/open-match/examples/evaluators/golang/simple/
 RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo .
 
-FROM gcr.io/distroless/static
-COPY --from=builder /go/src/github.com/GoogleCloudPlatform/open-match/examples/evaluators/golang/simple .
+FROM gcr.io/distroless/static  
+COPY --from=builder /go/src/github.com/GoogleCloudPlatform/open-match/examples/evaluators/golang/simple/simple .
+COPY --from=builder /go/src/github.com/GoogleCloudPlatform/open-match/config/matchmaker_config.yaml ./config/
```
sawagh: Is this really needed?

sawagh: Ah - I remember seeing this go away in a later review. (just a symlink right)

cloudbuild.yaml
```diff
@@ -275,21 +277,19 @@ artifacts:
...
-#- 'gcr.io/$PROJECT_ID/openmatch-clientloadgen:${_OM_VERSION}-${SHORT_SHA}'
-#- 'gcr.io/$PROJECT_ID/openmatch-frontendclient:${_OM_VERSION}-${SHORT_SHA}'
-#substitutions:
-#    _OM_VERSION: 0.4.0
+- 'gcr.io/$PROJECT_ID/openmatch-backendapi:${_OM_VERSION}-${SHORT_SHA}'
+- 'gcr.io/$PROJECT_ID/openmatch-frontendapi:${_OM_VERSION}-${SHORT_SHA}'
+- 'gcr.io/$PROJECT_ID/openmatch-mmforc:${_OM_VERSION}-${SHORT_SHA}'
+- 'gcr.io/$PROJECT_ID/openmatch-mmlogicapi:${_OM_VERSION}-${SHORT_SHA}'
+- 'gcr.io/$PROJECT_ID/openmatch-evaluator-simple:${_OM_VERSION}-${SHORT_SHA}'
+- 'gcr.io/$PROJECT_ID/openmatch-cs-func-simple:${_OM_VERSION}-${SHORT_SHA}'
```
sawagh: I do not see any Dockerfile change to the C# sample Dockerfile in this CL. I am not familiar with the build mechanism for it - but seems different from the rest of these. Do we know if it builds and is still a valid example?

cloudbuild.yaml
jeremyje: https://github.com/GoogleCloudPlatform/open-match/pull/129/files

cloudbuild.yaml
jeremyje: Done.

examples/evaluators/golang/simple/Dockerfile
jeremyje: Added a TODO for this.

cloudbuild.yaml
jeremyje: The build happens as part of Docker Image: openmatch-cs-func-simple.
This section is to upload to the build artifacts (docker images and binaries) to gcr.io and GCS.

cmd/backendapi/Dockerfile
jeremyje: If we want to revisit this we can but I'm going to submit the PR since we don't have any coverage at all for image builds.

PR #102, opened 2019-03-10
Fix typos in comments, set and playerindices

watsonso:


internal/set/set.go
```diff
@@ -20,7 +20,7 @@ limitations under the License.
 
 package set
 
-// Intersection returns the interection of two sets.
+// Intersection returns the interaction of two sets.
```
joeholley: You were right, it's a typo; but it should say 'intersection' not 'interaction' :)
https://en.wikipedia.org/wiki/Intersection_(set_theory)

joeholley: Otherwise looks good, push a new commit with that fixed and I'll get it merged!

watsonso: @joeholley Thank you for your review. Fixed!

PR #103, opened 2019-03-11
Add Hugo/Docsy website with port of Markdown documentation

jeremyje:
In alignment with Agones, we'll use Hugo/Docsy as the templating engine for the Open Match website.

This PR is mean to baseline a working site but not meant to be a completely working or final website.

To stand up a local server:

```bash
make run-dev-site
```


jeremyje:
https://github.com/GoogleCloudPlatform/open-match/issues/108 is the issue for this PR.

jeremyje:
I'd prefer sooner so we can get CI support for this as well which will ease updating the website.

PR #106, opened 2019-03-13
Go Modules

jeremyje:
This change will pin Open Match dependencies to predictable versions. This will allow for more reproducible builds since go get pulls code from repository HEAD, with the exception of aliased repos like gopkg.in.

After looking at go modules and vendoring it turns out the vendor/ directory is not touched if you build with go mod.

The recommended way is to just check in the go.mod and go.sum files to pin your dependencies to a known good checkout and download the dependencies one time.

The dependency output will be copied to $GOPATH/pkg/mod.

This change will set the minimum golang version for Open Match to 1.11. 1.12 is the latest version, 1.13 will turn go mod on by default.

For now you'll need to add this to your ~/.bashrc
```bash
export GO111MODULE=on
```


jeremyje:
https://github.com/GoogleCloudPlatform/open-match/issues/91 is the related issue for this.

PR #119, opened 2019-03-22
Generate install/yaml/ files from Helm chart templates

ihrankouski:
Trying to minimize the number of k8s files that are created manually.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,11 +56,11 @@ TOOLCHAIN_DIR = $(BUILD_DIR)/toolchain
 TOOLCHAIN_BIN = $(TOOLCHAIN_DIR)/bin
 PROTOC := $(TOOLCHAIN_BIN)/protoc
 PROTOC_INCLUDES := $(TOOLCHAIN_DIR)/include/
-GCP_PROJECT_ID =
+GCP_PROJECT_ID ?=
 GCP_PROJECT_FLAG = --project=$(GCP_PROJECT_ID)
 OM_SITE_GCP_PROJECT_ID = open-match-site
 OM_SITE_GCP_PROJECT_FLAG = --project=$(OM_SITE_GCP_PROJECT_ID)
-REGISTRY = gcr.io/$(GCP_PROJECT_ID)
+REGISTRY ?= gcr.io/$(GCP_PROJECT_ID)
 TAG := $(VERSION)
 ALTERNATE_TAG := dev
 GKE_CLUSTER_NAME = om-cluster
@@ -68,7 +68,7 @@ GCP_REGION = us-west1
 GCP_ZONE = us-west1-a
 EXE_EXTENSION =
 LOCAL_CLOUD_BUILD_PUSH = # --push
-KUBECTL_RUN_ENV = --env='REDIS_SERVICE_HOST=$$(OPEN_MATCH_REDIS_MASTER_SERVICE_HOST)' --env='REDIS_SERVICE_PORT=$$(OPEN_MATCH_REDIS_MASTER_SERVICE_PORT)'
+KUBECTL_RUN_ENV = --env='REDIS_SERVICE_HOST=$$(OM_REDIS_MASTER_SERVICE_HOST)' --env='REDIS_SERVICE_PORT=$$(OM_REDIS_MASTER_SERVICE_PORT)'
 GCP_LOCATION_FLAG = --zone $(GCP_ZONE)
 # Flags to simulate behavior of newer versions of Kubernetes
 KUBERNETES_COMPAT = --no-enable-basic-auth -no-issue-client-certificate --enable-ip-alias
@@ -81,6 +81,9 @@ TILLER = $(TOOLCHAIN_BIN)/tiller
 MINIKUBE = $(TOOLCHAIN_BIN)/minikube
 KUBECTL = $(TOOLCHAIN_BIN)/kubectl
 SERVICE = default
+NAMESPACE ?= open-match
+OPEN_MATCH_NAME ?= open-match
+REDIS_NAME = om-redis
 
 ## Make port forwards accessible outside of the proxy machine.
 PORT_FORWARD_ADDRESS_FLAG = --address 0.0.0.0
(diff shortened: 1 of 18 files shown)


jeremyje:
Thanks for doing this. Really appreciate it.

install/yaml/02-open-match.yaml
```diff
@@ -1,60 +1,153 @@
...
+kind: ConfigMap
+metadata:
+  name: om-configmap
+  namespace: default
+  labels:
+    app: open-match
+    component: mmlogic
+    chart: open-match-0.4.0
+    release: openmatch
+    heritage: Tiller
```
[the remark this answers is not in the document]
jeremyje: To clarify, chart and heritage should be removed in the manual deployment.

install/yaml/02-open-match.yaml
```diff
@@ -1,60 +1,153 @@
...
+data:
+  matchmaker_config.yaml: "# kubectl create configmap om-configmap --from-file=config/matchmaker_config.yaml\ndebug:
```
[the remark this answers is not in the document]
ihrankouski: You mean to make kubectl read the file and include its content into configmap? No, it's not possible. We have to use tools like Helm or Kustomize to achieve this.

Well, now I think it's better to not generate this configmap as a part of static YAML installation and rather make users create it manually with `kubectl create configmap --from-file`...

install/yaml/02-open-match.yaml
```diff
@@ -135,157 +256,216 @@ spec:
...
+        component: mmforc
+        chart: open-match-0.4.0
+        release: openmatch
+        heritage: Tiller
     spec:
       containers:
-      - name: om-mmlogic
-        image: gcr.io/matchmaker-dev-201405/openmatch-mmlogicapi:0.4
+      - name: om-mmforc
+        image: "ihrankouski/openmatch-mmforc:0.4"
```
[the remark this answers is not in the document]
ihrankouski: Sure, accidentally committed my local messing.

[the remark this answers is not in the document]
ihrankouski: done, thanks

install/yaml/02-open-match.yaml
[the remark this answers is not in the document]
ihrankouski: what about "release" label?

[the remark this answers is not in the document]
ihrankouski: removed that as well

Makefile
```diff
@@ -46,9 +46,9 @@ TOOLCHAIN_BIN = $(TOOLCHAIN_DIR)/bin
 TOOLCHAIN_GOLANG_DIR = $(TOOLCHAIN_DIR)/golang
 PROTOC := $(TOOLCHAIN_BIN)/protoc
 PROTOC_INCLUDES := $(TOOLCHAIN_DIR)/include/
-GCP_PROJECT_ID = "Set $GCP_PROJECT_ID in your bashrc."
+GCP_PROJECT_ID ?= open-match-public-images
```
jeremyje: Please revert this line. This will cause privileged devs to accidentally overrwrite the public images during testing.

install/helm/open-match/templates/backendapi.yaml
```diff
@@ -42,20 +39,15 @@ metadata:
...
   selector:
     matchLabels:
       app: {{ template "openmatch.name" . }}
       component: backend
-      chart: {{ template "openmatch.chart" . }}
-      release: {{ .Release.Name }}
-      heritage: {{ .Release.Service }}
+      {{- include "openmatch.chartmeta" (set . "indent" 6) }}
   template:
-    namespace: {{ .Release.Namespace }}
```
jeremyje: What namespace is this going into? Or was this unncessary?

ihrankouski: Pod will be created in the namespace that is specified at [`.metadata.namespace` of Deployment definition](https://github.com/GoogleCloudPlatform/open-match/pull/119/files/12a44bcf2b6e6ee12e269388dc460a432b873b09#diff-ff3081741227225fc85e625cbc4c142aR38). That was not valid k8s config actually, kubectl complained about that.

ihrankouski: BTW it seems to me that it's better to have namespace-agnostic k8s configs and let anybody pass the desired namespace with "-n" arg.
- People may want to install Open-match in any namespace they want, and we'd better not make them do replaces in the file.
- The YAML output that Redis chart produces does not contain a namespace in resources definitions, so the components will be installed in "default" unless other is specified with `kubectl --namespace XXX apply -f ...`. On the other hand Open-match components expect to find Redis service by its name that currently is automatically substituted by Kubernetes (`$(OM_REDIS_MASTER_SERVICE_HOST)`) - in the same namespace. Of course it's possible to make Open-match talk to Redis in different namespace but again not without of editing k8s files.

install/yaml/02-open-match.yaml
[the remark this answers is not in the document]
ihrankouski: Well kustomize is a part of kubectl now: https://kubernetes.io/blog/2019/03/25/kubernetes-1-14-release-announcement/#notable-kubectl-updates

install/helm/open-match/templates/backendapi.yaml
ihrankouski: So I've removed "namespace" from Open-match k8s YAML. However one thing that is still coupled to a hardcoded namespace is a Service Accounts: "mmf-sa" in case of Open-match, and similar ones for Prometheus and Grafana. Not sure about how to deal with it...

jeremyje: I'm fine having a namespace as a parameter but using default is not desirable.

ihrankouski: You can install resources in any namespace with `kubectl --namespace XXX apply -f file.yaml` if it's not specified. In contrast if namespace is set explicitly in YAML then kubectl will complain in that case.

jeremyje:
We are getting close to cutting 0.4.0. I think this will be the last PR to make it in. Can we prioritize this? It looks like it's still in work in progress.

jeremyje:
Please switch to target the master branch.

ihrankouski:
> Please switch to target the master branch.

Rebased.

Makefile
```diff
@@ -269,35 +272,93 @@ print-chart: build/toolchain/bin/helm$(EXE_EXTENSION)
...
+		--set openmatch.config.install=false \
+		--set openmatch.backendapi.install=false \
+		--set openmatch.frontendapi.install=false \
+		--set openmatch.mmlogicapi.install=false \
+		--set openmatch.mmforc.install=false \
+		--set prometheus.enabled=false \
+		--set grafana.enabled=true \
+		install/helm/open-match > install/yaml/04-grafana-chart.yaml
+
+set-redis-password:
```
jeremyje: Add a TODO here to remove this and just have it as a property of the chart.

jeremyje: This is very minor. I'd rather get this in now. I'll add the TODO.

PR #120, opened 2019-03-22
Remove symlinks to config files because they are mounted via ConfigMaps.

jeremyje:
There's still a symlink remaining on the helm chart. Not sure how to get rid of that one but suggestions welcome.

### DIFF
--- a/cmd/backendapi/matchmaker_config.yaml
+++ b/cmd/backendapi/matchmaker_config.yaml
@@ -1,1 +1,0 @@
-../../config/matchmaker_config.yaml
--- a/cmd/frontendapi/matchmaker_config.yaml
+++ b/cmd/frontendapi/matchmaker_config.yaml
@@ -1,1 +1,0 @@
-../../config/matchmaker_config.yaml
--- a/cmd/mmforc/matchmaker_config.yaml
+++ b/cmd/mmforc/matchmaker_config.yaml
@@ -1,1 +1,0 @@
-../../config/matchmaker_config.yaml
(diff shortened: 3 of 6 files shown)


ihrankouski:
I don't think it's possible to get rid of that symlink because the rendering of chart happens is actually done by Tiller, and Helm client only sends the chart to Tiller during `helm install`.

ihrankouski:
What if we move matchmaker_config.yaml from `/config/` to `/install/helm/open-match/config/`?
It doesn't make much sense to have it stored out of the configmap since 'base' Docker image does not include this file any more. And there's not so much difference which file to edit to customize configuration. (Assuming that `make gen-install` will create a separate file for configmap in `/install/yaml`).

Moreover it feels to me that all the config properties should be a separate params to Open-match Helm chart (like `openmatch.config.defaultImages.evaluator` or `openmatch.config.playerIndices`). Then these keys could be used in configmap Helm template to produce the output. Thus we'd have both static configmap installation file and a flexibility to override default values during `helm install` without edits of any files.

jeremyje:
I think splitting up the configuration has floated around but nothing has materialized from it. I've created issue, https://github.com/GoogleCloudPlatform/open-match/issues/130.

ihrankouski:
Also "defaultImages" in matchmaker_config.yaml should have the same version as in chart & install/yaml/. Either we need manage this with tags like "latest" or render these values with Helm template.

ihrankouski:
I'd like to take #130 which will probably need the merge of configuration from several sources which will likely require some changes in config.go. Maybe it's a time to merge this PR?

PR #126, opened 2019-03-24
Backport GDC work to 040 wip branch.

joeholley:
There will still be some work to cut the 040 RC, but this includes:

- Support for MMFs run as k8s services instead of jobs, called from the backend API (not MMForc)
 - New pbs for MMF gRPC API and messages to support it
- Support for Evaluator run as a k8s service instead of being called from MMForc.
- An example matchmaker using the "Director" pattern
- A few other odds and ends.  Check the changelog


CHANGELOG.md
```diff
@@ -6,7 +6,14 @@
...
  - This paves the way for CI/CD in an upcoming version.  
  - This paves the way for public images in an upcoming version!
-
+- The code to read pools from your `MatchObject` `Properties` JSON blob when you pass it into the Backend API and make Open Match 'native' protobuf `PlayerPool` messages out of them for use with the MMLogic API is deprecated and will be removed in a future version. When that happens, if you want to use the MMLogic API, you will be required to populate the MatchObjects's `pools` field yourself.
+- The same with the code that makes `Rosters` messages out of your `Properties` JSON blob.  You'll have to populate the `Rosters` field yourself to make use of the MMLogic API in upcoming versions.
+- Running MMFs as kubernetes `jobs` is deprecated.  This pattern and the `MMForc` component will be removed in a future version.
+- The Backend API no longer prepends the timestamp as part of the key under which results are stored in Redis. This simplifies and shortens the keys a bit.  Existing evaluators will need to be updated to parse the new key format when they read from the `proposalq`.
+- Evaluator sample in golang got a pretty big overhaul.
```
sawagh: The future direction being considered is to not store the proposals in redis but just pass them in memory to the evaluator. The current change that moves the evaluator into a separate harness assuming it can read proposals out of the db - which may not work in future. 

Given that Open Match will trigger evaluation passing in in-memory proposals, we can either move back to k8s job for evaluator (which may not have scaling concerns) - or for consistency - if we use the harness approach - the evaluator may implement an interface over which Open Match will trigger evaluation. 

We can either leave the k8s mechanism for evaluator as is - or - create an issue to have the harness implement a method to trigger evaluation in 0.5.

api/protobuf-spec/messages.proto
```diff
@@ -20,6 +20,31 @@ message MatchObject{
...
   repeated PlayerPool pools = 5;        // 'Hard' filters, and the players who match them.  
   string status = 6;                    // Resulting status of the match function
+  string name = 7;                      // Arbitrary developer-chosen, human-readable string.  Optional.
+
+}
+
+message MmfSpec {
+    string name = 1;  // Arbitrary developer-chosen, human-readable string.  Optional.
+    string host = 2;  // Host or DNS name for service providing this MMF.  Must be resolve-able by the backend API.
+    int32 port = 3;   // Port number for service providing this MMF.
```
jeremyje: I think we'll want to revisit this. There might be a security concern here. It's probably better to configure the routing via ConfigMap. For now just add a TODO.

api/protobuf-spec/messages.proto
```diff
@@ -20,6 +20,31 @@ message MatchObject{
...
+    enum Type {
+        // Generic serving types.
+        GRPC = 0;
+        REST = 1;
+        // Knative types will be KNATIVEREST, KNATIVEGRPC, etc.
```
jeremyje: I'd remove this comment.

cmd/evaluator/main.go
```diff
@@ -0,0 +1,27 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import "github.com/GoogleCloudPlatform/open-match/internal/app/evaluator"
+
+func init() {
```
jeremyje: Do not use init() functions. e2e tests cannot control the execution of the code. See the other binaries for an example.

examples/evaluators/golang/simple/main.go
```diff
@@ -26,41 +26,52 @@ package main
...
 	"github.com/gobs/pretty"
 	"github.com/gomodule/redigo/redis"
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 )
 
+var (
+	lgr    *log.Logger
+	cfg    *viper.Viper
+	err    error
```
jeremyje: Having err as a global is a major pitfall because the err can be updated anywhere and give unpredictable results. This is particularly nasty with goroutines.

examples/evaluators/golang/simple/main.go
```diff
@@ -26,41 +26,52 @@ package main
...
 	// Connect to redis
-	pool, err := redishelpers.ConnectionPool(cfg)
-	if err != nil {
-		lgr.Fatal(err)
-	}
-	defer pool.Close()
-
+	// As per https://www.iana.org/assignments/uri-schemes/prov/redis
+	// redis://user:secret@localhost:6379/0?foo=bar&qux=baz // redis pool docs: https://godoc.org/github.com/gomodule/redigo/redis#Pool
+	pool := redishelpers.ConnectionPool(cfg)
```
jeremyje: Please revert this back. This method returns pool, err now. This line does not concern itself with the credentials so it's unclear why this comment is necessary. If anything the comment should live in redishelpers.

examples/evaluators/golang/simple/main.go
```diff
@@ -72,57 +83,107 @@ func main() {
 
 	start := time.Now()
 
+	//proposedMatchIds  []string,
+	//overloadedPlayers map[string][]int
+	//overloadedMatches map[int][]int
+	//approvedMatches   map[int]bool
```
jeremyje: I think this comment is unnecessary, IDEs will tell devs what the types are. This will go out of date very quickly.

examples/evaluators/golang/simple/main.go
```diff
@@ -72,57 +83,107 @@ func main() {
...
+	lgr.Info("overloadedMatchList")
 	pretty.PrettyPrint(overloadedMatchList)
-	fmt.Println("approvedMatchList")
+	lgr.Info("approvedMatchList")
 	pretty.PrettyPrint(approvedMatchList)
 
+	// Choose matches to approve from among the overloaded ones.
 	approved, rejected, err := chooseMatches(overloadedMatchList)
 	approvedMatchList = append(approvedMatchList, approved...)
+	approvedPlayers := make([]string, 0)
```
jeremyje: approvedPlayers := []string{}

examples/evaluators/golang/simple/main.go
```diff
@@ -72,57 +83,107 @@ func main() {
...
 		// The match object was already written by the MMF, just change the
 		// name to what the Backend API (apisrv.go) is looking for.
 		proposedID := proposedMatchIds[proposalIndex]
 		// Incoming proposal keys look like this:
-		// proposal.1542600048.80e43fa085844eebbf53fc736150ef96.testprofile
+		// proposal.80e43fa085844eebbf53fc736150ef96.testprofile
 		// format:
-		// "proposal".timestamp.unique_matchobject_id.profile_name
+		// "proposal".unique_matchobject_id.profile_name
 		values := strings.Split(proposedID, ".")
```
jeremyje: Add a precondition to guarantee that the name format is accurate before reading fields based on index. Otherwise the error here is going to be a cryptic error.

examples/functions/golang/serving-mmlogic-simple/main.go
```diff
@@ -0,0 +1,135 @@
...
+		"component": "function_service",
+	}
+	fnLog = log.WithFields(fnLogFields)
+
+	// Viper config management setup
+	cfg = viper.New()
+	err = errors.New("")
+)
+
+func init() {
```
jeremyje: Fold this into Run().

A later pull request will canonicalize server initialization for 2 reasons, simplified e2e server bootstraping and adding gRPC secured mode.

For example: https://github.com/jeremyje/open-match/blob/alignment/internal/app/backendapi/backendapi.go.

examples/matchmaker/mo/mo.go
```diff
@@ -0,0 +1,222 @@
...
+	tiers    []string
+)
+
+type poolArray []*pb.PlayerPool
+type ftrArray []*pb.Filter
+type rosterArray []*pb.Roster
+
+/*
+//example of how to use this module
+func main() {
```
jeremyje: Consider converting this to a golang example test file. 
https://blog.golang.org/examples

examples/matchmaker/mo/mo.go
```diff
@@ -0,0 +1,222 @@
...
+			log.Println("all Done")
+			return
+		}
+	}
+}
+*/
+
+// GenerateMatchObjects
+func GenerateMatchObjects(moChan chan *pb.MatchObject) {
+	//func GenerateMatchObjects(moChan chan *pb.MatchObject, doneChan chan bool) {
```
jeremyje: Please remove dead/commented code if it's not relevant.

api/protobuf-spec/messages.proto
```diff
@@ -20,6 +20,31 @@ message MatchObject{
...
+}
+
+message MmfSpec {
+    string name = 1;  // Arbitrary developer-chosen, human-readable string.  Optional.
+    string host = 2;  // Host or DNS name for service providing this MMF.  Must be resolve-able by the backend API.
+    int32 port = 3;   // Port number for service providing this MMF.
+    enum Type {
+        // Generic serving types.
+        GRPC = 0;
+        REST = 1;
```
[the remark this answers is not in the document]
joeholley: Unity already has support for REST, though?  It's in @calebatwd 's previous branches.  We can leave it out for 0.4.0 if we don't want to try to get Unity to use MMFSpec tho I guess

api/protobuf-spec/messages.proto
```diff
@@ -20,6 +20,31 @@ message MatchObject{
   repeated Roster rosters = 4;          // Rosters of players.  
   repeated PlayerPool pools = 5;        // 'Hard' filters, and the players who match them.  
   string status = 6;                    // Resulting status of the match function
+  string name = 7;                      // Arbitrary developer-chosen, human-readable string.  Optional.
+
+}
+
+message MmfSpec {
+    string name = 1;  // Arbitrary developer-chosen, human-readable string.  Optional.
```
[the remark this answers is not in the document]
joeholley: Thoughts were around naming it for logging and metrics recording, yes.  Without this, using something that procedurally generates service names (like knative) could introduce [cardinality issues with prometheus](https://prometheus.io/docs/practices/naming/#labels).  Having 'name' there from the beginning will allow us to standardize on using it from the get-go.

api/protobuf-spec/messages.proto
```diff
@@ -20,6 +20,31 @@ message MatchObject{
   repeated Roster rosters = 4;          // Rosters of players.  
   repeated PlayerPool pools = 5;        // 'Hard' filters, and the players who match them.  
   string status = 6;                    // Resulting status of the match function
+  string name = 7;                      // Arbitrary developer-chosen, human-readable string.  Optional.
```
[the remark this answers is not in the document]
joeholley: I see human-readable names for these reusable objects as pretty fundamental - if I were, for example, building some kind of admin interface for my matchmaker, one of the things I would definitely want would be to be able to dump out info about every matchobject it has sent or received.  Dumping out the actual object would be huge and unreadable, so as a developer I'd elect to name the objects according to what they contain, and then display the list of names.

Also, having optional fields in protobufs doesn't have a significant cost, but forcing someone to maintain their own fork of a core component to add something as basic as a name is very significant IMHO.

api/protobuf-spec/messages.proto
```diff
@@ -98,13 +123,14 @@ message Assignments{
 
 // The message for passing in the per-request identifiers
 // to a matchmaking function; used so it knows which records to
-// write/update in state storage.
+// write/update in state storage.  Internal to Open Match; this is populated by
+// the Backend API and sent to your gRPC-served MMF.
 message Request {
```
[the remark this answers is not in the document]
joeholley: This is just temp, @jeremyje  is going to overhaul all of these post MR

api/protobuf-spec/messages.proto
```diff
@@ -98,13 +123,14 @@ message Assignments{
...
-    string profile_id = 1;
-    string proposal_id = 2;
-    string request_id = 3;
-    string error_id = 4;
-    string timestamp = 5;
+    string profile_id = 1;  // Developer-chosen profile name, a copy of the MatchObject name.
+    string request_id = 2;  // Per-request generated unique ID, by convention XID.
```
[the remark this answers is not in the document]
joeholley: Both are used now but feel free to deprecate once we've moved all the request details to the call context

api/protobuf-spec/messages.proto
```diff
@@ -98,13 +123,14 @@ message Assignments{
...
+    string proposal_id = 3; // Intermediate result proposal ID. by convention, the string 'proposal.' prepended to the result_id. 
+    string result_id = 4;   // Final result ID. By convention, request_id.profile_id
+    int64 timestamp = 5;    // Reserved for future use.
```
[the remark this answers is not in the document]
joeholley: Yeah this doesn't need to be here, will remove

examples/backendclient/profiles/testprofile.json
```diff
@@ -1,6 +1,8 @@
 {
-  "imagename":"gcr.io/open-match-public-images/openmatch-mmf-py3-mmlogic-simple:dev",
+  "imagename":"gcr.io/matchmaker-dev-201405/openmatch-mmf-py3-mmlogic-simple:dev",
```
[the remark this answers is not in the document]
joeholley: mmfs as jobs is marked as deprecated but not removed in 0.4.0, unless we want to make that leap.  Would prefer to give people a version with both jobs and services to have time to migrate if they have existing MMFs?

examples/functions/golang/serving-mmlogic-simple/apisrv/harness.go
```diff
@@ -0,0 +1,198 @@
...
+		"component": "function_mmlogic_harness",
+	}
+	mmfLog = log.WithFields(mmfLogFields)
+)
+
+// Step 2 - Talk to Redis.
+// This example uses the MM Logic API in OM to read/write to/from redis.
+
+// mmfRun is used to kick off the served MMF.
+func mmfRun(ctx context.Context, fnArgs *api.Arguments, cfg *viper.Viper, mmlogic api.MmLogicClient) error {
```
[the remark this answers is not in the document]
joeholley: Yeah, issue in for this already at https://github.com/GoogleCloudPlatform/open-match/issues/125

internal/app/backendapi/apisrv/apisrv.go
```diff
@@ -150,67 +152,133 @@ func (s *backendAPI) CreateMatch(c context.Context, profile *pb.MatchObject) (*p
...
-	// Queue the request ID to be sent to an MMF
-	_, err = redishelpers.Update(ctx, s.pool, s.cfg.GetString("queues.profiles.name"), requestKey)
-	if err != nil {
-		beLog.WithFields(log.Fields{
-			"error":     err.Error(),
-			"component": "statestorage",
-		}).Error("State storage failure to queue profile")
+	// Figure out what kind of MMF we are running and kick it off
+	switch beRequest.Mmfspec.Type {
+	case backend.MmfSpec_K8SJOB:
```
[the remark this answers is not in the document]
joeholley: See above comment

CHANGELOG.md
joeholley: Yeah what was discussed at some point was to put an API in front on evaluator when we're ready to get proposals out of redis.  MMFs would then call the evaluator API to send proposals to it when they are finished.  Best to let Unity drive that in a future version based on their experiences, I suspect.

api/protobuf-spec/messages.proto
joeholley: The expectation is that hosts and ports for served MMFs are running using ClusterIP and can't be accessed by anything running outside the cluster, so I think it's probably fine?

examples/evaluators/golang/simple/main.go
joeholley: fixed

examples/evaluators/golang/simple/main.go
joeholley: changed + two more places

examples/evaluators/golang/simple/main.go
joeholley: added

examples/matchmaker/mo/mo.go
joeholley: done

examples/matchmaker/mo/mo.go
joeholley: deleted

api/protobuf-spec/messages.proto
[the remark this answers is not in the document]
joeholley: updated proto

IIRC the name of the API was from @calebatwd 's branch @ https://github.com/GoogleCloudPlatform/open-match/compare/grpc-functions#diff-80dc8d1cc002a937b20888d5853a1c63R8 - may want to run the rename past him

api/protobuf-spec/messages.proto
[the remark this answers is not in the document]
joeholley: Useful to add to MMF logs/traces to aid in troubleshooting but I can take them out if we agree they aren't needed.

cmd/evaluator/main.go
joeholley: sorted

examples/functions/golang/serving-mmlogic-simple/main.go
joeholley: thanks for the example, sorted

sawagh:
Since this PR moves stuff from 'gdcbackports' to '040wip' - which is now locked since we are in master, I am removing this PR. I am separately working on moving changes from gdcbackports to master for 0.5

PR #129, opened 2019-03-25
Move Dockfiles to build C#, Golang, PHP, and Python3 MMFs.

jeremyje:


Makefile
```diff
@@ -130,13 +133,28 @@ push-mmlogicapi-image: build-mmlogicapi-image
 push-backendclient-image: build-backendclient-image
 	docker push $(REGISTRY)/openmatch-backendclient:$(TAG)
 
+push-mmf-cs-mmlogic-simple-image: build-mmf-cs-mmlogic-simple-image
+	docker push $(REGISTRY)/openmatch-mmf-cs-mmlogic-simple:$(TAG)
```
sawagh: A naming suggestion - why do we have mmlogic in the example function names?  - a simple cs mmf example can just be mmf-cs-simple?

sawagh: I ll rename these in a separate CL! for now - checkin.

PR #141, opened 2019-04-01
fix for evaluator error

joeholley:
Error regression from changes in the previous commits.

### DIFF
--- a/internal/app/evaluator/harness.go
+++ b/internal/app/evaluator/harness.go
@@ -71,7 +71,8 @@ func initializeApplication() {
 	log.AddHook(metrics.NewHook(EvaluatorLogLines, KeySeverity))
 
 	// Viper config management initialization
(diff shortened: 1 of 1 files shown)


internal/app/evaluator/harness.go
```diff
@@ -71,7 +71,8 @@ func initializeApplication() {
 	log.AddHook(metrics.NewHook(EvaluatorLogLines, KeySeverity))
 
 	// Viper config management initialization
-	cfg, err := config.Read()
+	var err error
```
jeremyje: How does this fix the problem this probably produces the same code in Golang.

sawagh: Wont cfg, err := config.Read() only set a local variable 'cfg' whereas 	cfg, err = config.Read() actually update the global cfg which is the intent in initializeApplication ?

PR #142, opened 2019-04-01
Fix dependency issues in the build.

jeremyje:
This change addresses https://github.com/GoogleCloudPlatform/open-match/issues/143.

 * Fix build dependencies
 * Add README stickers
 * Switch back to '.' from source since some shells don't like it.

Because we don't have CI coverage for make yet here's the command I ran in a GCE VM to verify the build.

```bash
sudo rm -rf workspace/ \
  && export GOPATH=$HOME/workspace \
  && export GO111MODULE=on \
  && mkdir -p workspace/src/github.com/GoogleCloudPlatform \
  && cd workspace/src/github.com/GoogleCloudPlatform/ \
  && git clone https://github.com/jeremyje/open-match.git \
  && cd open-match \
  && git checkout -b fixdeps origin/fixdeps \
  && make install-toolchain \
  && make install-npm build/site/ \
  && make all build-images -j8
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # Open Match
 
+[![GoDoc](https://godoc.org/github.com/GoogleCloudPlatform/open-match?status.svg)](https://godoc.org/github.com/GoogleCloudPlatform/open-match)
+[![Go Report Card](https://goreportcard.com/badge/github.com/GoogleCloudPlatform/open-match)](https://goreportcard.com/report/github.com/GoogleCloudPlatform/open-match)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -183,10 +187,18 @@ The easiest way to build Open Match is to use the Makefile. Before you can use t
 
 ```bash
```
sawagh: Separately, could you add a note to what need to be the steps for OS X?

jeremyje: You basically have to download the dependencies from the websites. I'll add links to the websites.

PR #152, opened 2019-04-03
Mini Match

jeremyje:
This PR does a few things.
* Removes many of the global variables that were causing subtle bugs like errors potentially crossing grpc handler paths that may or may not be happening because of Golang variable shadowing semantics.
* Adds e2e test harness for bringing up an in-memory Open Match cluster and then sending requests to it.
* Mini Match binary that can be used for testing MMF, Evaluator, Director logic outside of Open Match, like C# in isolation.


rmb938:
Should there be some documentation (even if it's just basic) on how to use mini match?

jeremyje:
It's not a functional service yet. It's not deployable either at this point.

rmb938:
Ah I see

cloudbuild.yaml
```diff
@@ -81,7 +86,7 @@ steps:
   name: gcr.io/cloud-builders/docker
   args: ['build', '-t', 'gcr.io/$PROJECT_ID/openmatch-mmf-cs-mmlogic-simple:${_OM_VERSION}-${SHORT_SHA}', '.']
   dir: 'examples/functions/csharp/simple'
-  waitFor: ['Docker Image: open-match-base-build']
+  waitFor: ['-']
```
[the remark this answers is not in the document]
jeremyje: Only the go builds need to wait because go the go get cache. You can look at each docker file to see who depends on it.

jeremyje:
I've incorporated your feedback but some of the PR has broken off into smaller PRs. After the other 2 are submitted I think the remaining code should be submitted as a unit.

internal/app/minimatch/minimatch.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// CreateServerParams creates the configuration and prepares the binding for serving handler.
+func CreateServerParams() []*serving.ServerParams {
+	// Add a hook to the logger to auto-count log lines for metrics output thru OpenCensus
+	log.AddHook(metrics.NewHook(frontend.FeLogLines, frontend.KeySeverity))
+	// Add a hook to the logger to auto-count log lines for metrics output thru OpenCensus
+	log.AddHook(metrics.NewHook(backend.BeLogLines, backend.KeySeverity))
+	// Add a hook to the logger to auto-count log lines for metrics output thru OpenCensus
+	log.AddHook(metrics.NewHook(mmlogic.MlLogLines, mmlogic.KeySeverity))
```
Laremere: Aren't these hooks already added by the calls to CreateServerParams?

test/e2e/minimatch/minimatch_test.go
```diff
@@ -0,0 +1,89 @@
+package frontend
```
Laremere: Is this the right package?

test/e2e/minimatch/minimatch_test.go
```diff
@@ -0,0 +1,89 @@
...
+	defer closer()
+	mm.Start()
+	defer mm.Stop()
+
+	feClient, err := mm.GetFrontendClient()
+	if err != nil {
+		t.Fatalf("cannot obtain frontend client, %s", err)
+	}
+	if feClient == nil {
+		t.Fatalf("cannot get frontend client, %v", feClient)
```
Laremere: What's the value of logging nil?

test/e2e/minimatch/minimatch_test.go
```diff
@@ -0,0 +1,89 @@
...
+		player := createFakePlayer()
+		//lastPlayer = player
+		_, err = feClient.CreatePlayer(ctx, &pb.CreatePlayerRequest{
+			Player: player,
+		})
+		if err != nil {
+			t.Errorf("request error, %s", err)
+		}
+	}
+	/*
```
Laremere: Please include some context of where this is going in a TODO.

PR #154, opened 2019-04-03
Fix 'make create-gke-cluster'

Laremere:
It is missing a dash on one of the arguments, which breaks things.


jeremyje:
```bash
git fetch upstream
git rebase -i upstream/master
```

You'll want to squash the commits to just have 1 change.

Issue #130, opened 2019-03-25
Proposal: Split up matchmaker_config.yaml


jeremyje:
The matchmaker_config.yaml has grown to be a very large file and Open Match is a set up different servers with different context. We should consider splitting up the configuration so that each server is only provided the context necessary to run properly.

This should improve reliability because it'll force users to have multiple ConfigMaps for each configuration so if there's a bad config change it'll only impact a subset of server types instead of all of them.

ihrankouski:
I believe the ConfigMap could still remain single, but instead it will contain several entries in "data" section. Some pieces of configuration should be shared by services, some are independent, isn't it?

joeholley:
Thought about this for a bit.  There are definitely some per-component configs, where every component can have different values for these without issues. In fact, in many situations it would be preferred.  In particular, having the logging config and debug flags on a 'per component' basis would be nice.  

However, the majority of the parameters in the configs are there specifically for sharing - so that the values aren't baked into the code and when a change needs to be made, every component gets the same values (and hopefully avoid a situation where multiple components could accidentally have different values). To be honest, it would probably be a good idea to isolate the ones of those that we don't want beginners accidentally changing, like service names  and port numbers.  I still wouldn't want them baked directly into the code, but maybe move them back to being in a config location built into the container images by default.  We could then update the OM config loading module to 'override' the defaults with values from a configmap if it exists.  That way expert users (people deploying OM to a K8s cluster with many other apps and potential port conflicts, for example) could configure those sorts of things, but beginners aren't exposed to all that extra complexity that they probably don't need to mess with at all.  Some other things, like the list of properties to index, need to be kept in sync between the component writing the index (FE API) and the component deleting the index (BE API), but also should absolutely be modified by someone coding a matchmaker against OM.

Right now based on 0.4.0 WIP branch, and leaving out things that are getting deprecated and should be removed in future versions:

I see the first group ('per component' config) as containing:
```
debug: true
logging:
  level: debug
  format: text
  source: true
redis:
  pool:
    maxIdle: 3
    maxActive: 0
    idleTimeout: 60
  queryArgs:
    count: 10000
  results:
    pageSize: 10000
```
The second group ('we should hide this but allow overrides for experts' config):
```
api: 
  backend: 
    hostname: om-backendapi
    port: 50505
  frontend: 
    hostname: om-frontendapi
    port: 50504
  mmlogic: 
    hostname: om-mmlogicapi
    port: 50503
  functions:
    port: 50502
[...]

PR #158, opened 2019-04-05
Split up matchmaker_config.yaml

ihrankouski:
This proposes a solution to #130.

The configuration of core Open-match components ('services') is split into 3 files:
- /config/component_config.yaml - contains a per-component configuration
- /config/openmatch_config.yaml - configuration that must be shared by all components
- /config/openmatch_constants.yaml - configuration that must be shared but is more dangerous to modify (so it's built into docker images)

File "openmatch_config.yaml" is also mounted to MMF and Evaluator Job's containers.

Original "matchmaker_config.yaml" is removed along with all its symlinks.

Also moved config package under internal/.

Also I placed the contents of "component_config.yaml" and "openmatch_config.yaml" inside the values.yaml file of Helm chart. So in theory one could even override some parts of Open-match configuration at chart installation time (didn't test how it works yet).

Note that configuration watching and reloading is not implemented at the moment. (Because of merging of several files in Viper and usage of "subPath")


config/openmatch_constants.yaml
```diff
@@ -0,0 +1,39 @@
+# A copy of this file is built into Open-match components' Docker images.
```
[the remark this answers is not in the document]
ihrankouski: On the other side if we always keep all configuration in ConfigMap it will be easier to get config reloading back...

test/cmd/clientloadgen/main.go
```diff
@@ -23,15 +23,15 @@ import (
...
+	"github.com/GoogleCloudPlatform/open-match/internal/config"
 	"github.com/GoogleCloudPlatform/open-match/test/cmd/clientloadgen/player"
 	"github.com/GoogleCloudPlatform/open-match/test/cmd/clientloadgen/redis/playerq"
 
 	"github.com/gomodule/redigo/redis"
 )
 
 func main() {
-	cfg, err := config.Read()
+	cfg, err := config.Read("config.yaml")
```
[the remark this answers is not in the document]
ihrankouski: Makes sense. 'clientloadgen' is just a tool to flood the state storage with dummy data, and it needed it's own configuration (that is not related to core components' params). I agree, having a convention of ['config.yaml' + ReadConfig()] would be sufficient to make pkg /internal/config more reusable, will add that.

examples/backendclient/main.go
```diff
@@ -138,6 +137,7 @@ func main() {
 				break
 			}
 
+			match := listResponse.Match
```
jeremyje: Can you sync? This was fixed.

examples/functions/python3/mmlogic-simple/harness.py
```diff
@@ -28,13 +29,13 @@
...
 # Establish grpc channel and make the API client stub
-api_conn_info = "%s:%d" % (os.environ["OM_MMLOGICAPI_SERVICE_HOST"],os.environ["OM_MMLOGICAPI_SERVICE_PORT"])
+api_conn_info = "%s:%s" % (os.environ["OM_MMLOGICAPI_SERVICE_HOST"],os.environ["OM_MMLOGICAPI_SERVICE_PORT"])
 with  grpc.insecure_channel(api_conn_info) as channel:
     mmlogic_api = mmlogic_pb2_grpc.MmLogicStub(channel)
 
     # Step 3 - Read the profile written to the Backend API.
     # Get profile from redis
-    profile_pb = mmlogic_api.GetProfile(mmlogic.MatchObject(id=os.environ["MMF_PROFILE_ID"]))
+    profile_pb = mmlogic_api.GetProfile(mmlogic.GetProfileRequest(match = pb.MatchObject(id=os.environ["MMF_PROFILE_ID"]))).match
```
jeremyje: Split this into a separate change. Much appreciated fix.

ihrankouski: done #193

examples/backendclient/main.go
ihrankouski: done

examples/backendclient/main.go
```diff
@@ -138,6 +138,7 @@ func main() {
 			}
 			match := listResponse.Match
 
+			match := listResponse.Match
```
jeremyje: Bad merge?

test/cmd/clientloadgen/Dockerfile
```diff
@@ -17,8 +17,9 @@ FROM open-match-base-build as builder
 WORKDIR /go/src/github.com/GoogleCloudPlatform/open-match/test/cmd/clientloadgen/
 RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o clientloadgen . 
 
-FROM gcr.io/distroless/static  
+FROM alpine
```
jeremyje: Why?

test/cmd/clientloadgen/Dockerfile
```diff
@@ -17,8 +17,9 @@ FROM open-match-base-build as builder
...
 COPY --from=builder /go/src/github.com/GoogleCloudPlatform/open-match/test/cmd/clientloadgen/clientloadgen .
+COPY --from=builder /go/src/github.com/GoogleCloudPlatform/open-match/test/cmd/clientloadgen/config.yaml ./
```
jeremyje: Please don't add configs to the images.

examples/backendclient/main.go
ihrankouski: ooops...

internal/config/merged_config.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, l := range layers {
+		m := l.AllSettings()
+		cfg.MergeConfigMap(m)
+	}
+	return cfg
+}
+
+// Wrapper struct that implements View interface
+// and delegates to other viper.Viper instance
+type wrapperView struct {
```
jeremyje: Why is this necessary if you have the merge() function?

internal/config/view.go
```diff
@@ -34,6 +34,7 @@ type View interface {
 	GetBool(string) bool
 	GetDuration(string) time.Duration
 	GetStringMap(string) map[string]interface{}
+	AllSettings() map[string]interface{}
```
jeremyje: AllSettings is only used against concrete `*viper.Viper` so this isn't necessary. If you need it because of tests I'd prefer a type assertion case to viper.Viper there since you already know you're testing against that class to begin with.

jeremyje:
Those should be my last comments on this change.

internal/config/merged_config.go
ihrankouski: I didn't find a function to replace all contents of a Viper instance. That means that if we use the same instance upon layers' reloads, then that instance might become polluted with some old data from previous reads.

internal/config/view.go
ihrankouski: Makes sense.

ihrankouski:
Well `go test` reports a data race: https://console.cloud.google.com/cloud-build/builds/c2a4b933-405e-4440-b52a-e6d21c872c6c?project=open-match-build#step_15
I'm not sure about how to deal with that because the issue seems to originate inside Viper, similar issues:
https://github.com/spf13/viper/issues/174
https://github.com/spf13/viper/issues/268

internal/config/view.go
ihrankouski: Done, thanks!

jeremyje:
Were you able to get rid of the race condition?
I think for the purpose of the change we should break out the file move first and then add the split if the race condition still is there.

jeremyje:
I think the easiest thing to do is bring back the wrapper and add a lock to use each method. We don't read from the config often enough for it to hurt perf.

jeremyje:
Merge conflicts. CI via Make just got submitted should provide better regression coverage.

ihrankouski:
Okay I'm still fighting with data races. What's weird is that "go1.12.1 darwin/amd64" does not report races on this branch for me any more (after my attempt to fix races). However I sometimes can reproduce it on clean instances in GCE.
I'll open a new PR with only files move.

jeremyje:
https://github.com/GoogleCloudPlatform/open-match/issues/213

jeremyje:
There's a lot of conflicts here. Is this change still going in?

ihrankouski:
I'm closing this, let's try to get #204 working at first.

PR #171, opened 2019-04-09
Disable the PHP example during the CI build.

jeremyje:
PHP image build is very expensive because it has to download all the debian build tools like clang to build protobuf and grpc C++ libraries. This alongside all the other build artifacts takes a very long time.


jeremyje:
https://github.com/GoogleCloudPlatform/open-match/issues/172

jeremyje:
This reduced the build by 4 minutes.

Issue #160, opened 2019-04-09
Adding transcoding mappings to gRPC


yfei1:
Following the transcoding [guideline](https://cloud.google.com/endpoints/docs/grpc/transcoding), Open Match should support protocol transcoding such that client can access gRPC by using JSON/HTTP requests. This simplifies client side testing and provides flexibility to access service resources.

yfei1:
Resolved by #183.

PR #183, opened 2019-04-10
REST Implementation

yfei1:
This resolves issue #160


Makefile
```diff
@@ -452,18 +452,24 @@ build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION):
...
 		-I $(CURDIR) -I $(PROTOC_INCLUDES) \
+		-I $(GOPATH)/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis -I $(GOPATH)/src \
 		--go_out=plugins=grpc:$(REPOSITORY_ROOT)
 
-php-protos: examples/functions/php/mmlogic-simple/proto/
 examples/functions/php/mmlogic-simple/proto/:
 	mkdir -p examples/functions/php/mmlogic-simple/proto/
 	$(PROTOC) api/protobuf-spec/messages.proto \
 		-I $(CURDIR) -I $(PROTOC_INCLUDES) \
+		-I $(GOPATH)/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis -I $(GOPATH)/src \
```
jeremyje: This regresses the effort to move way from Go workspace setup.

What files are necessary here? Much of them are in $(PROTOC_INCLUDES) directory. Is there something missing there? https://github.com/grpc-ecosystem/grpc-gateway/tree/master/third_party/googleapis

If anything overlay the source repositories data on top of the include directory. You can download the zip of the repository. This would be a separate build target.

Something like this should work.

```Makefile
build/toolchain/bin/protoc$(EXE_EXTENSION):
	... APPEND THE FOLLOWING LINES ...
	mkdir -p $(TOOLCHAIN_DIR/googleapis-temp/
	curl -o $(TOOLCHAIN_DIR)/googleapis-temp/googleapis.zip -L https://github.com/googleapis/googleapis/archive/master.zip
	(cd $(TOOLCHAIN_DIR/googleapis-temp/); unzip -o googleapis.zip)
	cp -rf $(TOOLCHAIN_DIR)/googleapis-temp/googleapis-master/google/rpc/ $(TOOLCHAIN_DIR)/protoc/includes/google/
	rm -rf $(TOOLCHAIN_DIR/googleapis-temp)
```

api/protobuf-spec/mmlogic.proto
```diff
@@ -89,14 +92,21 @@ service MmLogic {
   //        the backend api along with your match results.
   // OUTPUT: a Result message with a boolean success value and an error string
   // if an error was encountered
-  rpc CreateProposal(CreateProposalRequest) returns (CreateProposalResponse) {}
+  rpc CreateProposal(CreateProposalRequest) returns (CreateProposalResponse) {
+    option (google.api.http) = {
+      put: "/v1/logic/match-proposals/{match.id}/{match.error}"
```
jeremyje: This shouldn't have field parameters in the URL. Everything should go in the body. Typically creates do not have identifiers in them.

Makefile
jeremyje: Looking at the rest of the code /rpc/ is not referenced anywhere. I'd only add this Makefile dependency if the /rpc/ directory is necessary through transitive include.

yfei1: Adding REST proxy via gRPC gateway requires 'google/api/annotations.proto' file. You are right, I'll update the commit such that it only changes the dependency part.

api/protobuf-spec/mmlogic.proto
yfei1: Quoted from [gRPC HTTP Rules](https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#httprule):

"NOTE: the field paths in variables and in the body must not refer to repeated fields or map fields."

"The name of the request field whose value is mapped to the HTTP body, or * for mapping all fields not captured by the path pattern to the HTTP body. NOTE: the referred field must not be a repeated field and must be present at the top-level of request message type."

Would it means the repeated types are not allowed in the HTTP request body? The CreateProposalRequest messages contains a MatchObject type, which includes a repeated PlayerPool and repeated Roster. Is there any other workarounds to this limitation?

Makefile
```diff
@@ -361,7 +361,7 @@ set-redis-password:
 		printf "apiVersion: v1\nkind: Secret\nmetadata:\n  name: $(REDIS_NAME)\n  namespace: $(NAMESPACE)\ndata:\n  redis-password: $$REDIS_PASSWORD\n" | \
 		$(KUBECTL) replace -f - --force
 
-install-toolchain: build/toolchain/bin/protoc$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION) build/toolchain/bin/kubectl$(EXE_EXTENSION) build/toolchain/bin/helm$(EXE_EXTENSION) build/toolchain/bin/minikube$(EXE_EXTENSION) build/toolchain/bin/skaffold$(EXE_EXTENSION)  build/toolchain/bin/hugo$(EXE_EXTENSION) build/toolchain/python/
+install-toolchain: build/toolchain/bin/protoc$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION) build/toolchain/bin/kubectl$(EXE_EXTENSION) build/toolchain/bin/helm$(EXE_EXTENSION) build/toolchain/bin/minikube$(EXE_EXTENSION) build/toolchain/bin/skaffold$(EXE_EXTENSION)  build/toolchain/bin/hugo$(EXE_EXTENSION) build/toolchain/python/ build/toolchain/bin/protoc-gen-grpc-gateway
```
jeremyje: Add $(EXE_EXTENSION) because of Windows.

Makefile
```diff
@@ -452,15 +452,26 @@ build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION):
...
+	mkdir -p $(TOOLCHAIN_DIR)/googleapis-temp/
+	curl -o $(TOOLCHAIN_DIR)/googleapis-temp/googleapis.zip -L \
+		https://github.com/googleapis/googleapis/archive/master.zip
+	(cd $(TOOLCHAIN_DIR)/googleapis-temp/; unzip -o googleapis.zip)
+	cp -rf $(TOOLCHAIN_DIR)/googleapis-temp/googleapis-master/google/api/ \
+		$(PROTOC_INCLUDES)/google/api
+	rm -rf $(TOOLCHAIN_DIR)/googleapis-temp
+	$(GO) get github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway
+	$(GO) install github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway
+	mv $(GOPATH)/bin/protoc-gen-grpc-gateway$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-grpc-gateway$(EXE_EXTENSION)
```
jeremyje: Use mv $(GO_BIN)/

Makefile
```diff
@@ -490,10 +501,15 @@ internal/pb/matchfunction.pb.go: internal/pb/messages.pb.go
 mmlogic-simple-protos: examples/functions/python3/mmlogic-simple/api/protobuf_spec/messages_pb2.py examples/functions/python3/mmlogic-simple/api/protobuf_spec/mmlogic_pb2.py
 
 examples/functions/python3/mmlogic-simple/api/protobuf_spec/%_pb2.py: api/protobuf-spec/%.proto build/toolchain/python/
-	. build/toolchain/python/bin/activate && python3 -m grpc_tools.protoc -I $(CURDIR) -I $(PROTOC_INCLUDES) --python_out=examples/functions/python3/mmlogic-simple/ --grpc_python_out=examples/functions/python3/mmlogic-simple/ $< && deactivate
+	. build/toolchain/python/bin/activate && \
```
jeremyje: I think it's easier to read if the && is the start of each next line. That way you know that the line is a continuation from above instead of it's own command.

api/protobuf-spec/backend.proto
```diff
@@ -57,16 +58,31 @@ service Backend {
   //  - error. Empty if no error was encountered
   //  - rosters, if you choose to fill them in your MMF. (Recommended)
   //  - pools, if you used the MMLogicAPI in your MMF. (Recommended, and provides stats)
-  rpc CreateMatch(CreateMatchRequest) returns (CreateMatchResponse) {} 
+  rpc CreateMatch(CreateMatchRequest) returns (CreateMatchResponse) {
+    option (google.api.http) = {
+      put: "/v1/backend/matches"
+      body: "match"
```
jeremyje: Use `body: "*"`. Just to keep things consistent. It's not clear how many fields a top level request will have in the new API so for future compatibility I'd just use `body: "*"` here.

Issue #164, opened 2019-04-09
Makefile should detect and reject sudo invocations.


jeremyje:
Docker has a strange setup where if you don't do the [docker post-install instructions](https://docs.docker.com/install/linux/linux-postinstall/) devs are inclined to use sudo for the Docker commands.

The problem with this is that sudo/root typically do not have (and shouldn't have) the authentication token to talk to gcr.io or any GCP service via gcloud. So while sudo unblocks you from building the image you won't be able to upload it. A simple fix would be to detect sudo and notify devs that there's a doc to correct the setup.

jeremyje:
I'm rolling this back in https://github.com/GoogleCloudPlatform/open-match/pull/188.
If you run as sudo you get a cryptic message.

PR #187, opened 2019-04-10
Detect sudo invocations of Makefile #164

yfei1:
Resolve issue #164

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,15 @@
 ## make delete-gke-cluster
 ##
 # http://makefiletutorial.com/
+define newline
+
+
+endef
+
+ifeq ($(shell whoami),root)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -32,6 +32,15 @@
 # make delete-gke-cluster
 #
 ## http://makefiletutorial.com/
+define n
```
jeremyje: What's this for?

yfei1: It is used to define a newline in the Makefile to improve error message readability. I look around but could not find a second way to do it.

jeremyje:
You need to sync this PR.

Makefile
jeremyje: I think the diff is off here. Bad merge?

PR #188, opened 2019-04-10
CI with Makefile

jeremyje:
https://github.com/GoogleCloudPlatform/open-match/issues/137


cloudbuild.yaml
```diff
@@ -47,209 +47,98 @@
...
-  waitFor: ['Docker Image: open-match-base-build']
-
-- id: 'Docker Image: openmatch-mmf-go-mmlogic-simple'
-  name: gcr.io/cloud-builders/docker
-  args: ['build', '-t', 'gcr.io/$PROJECT_ID/openmatch-mmf-go-mmlogic-simple:${_OM_VERSION}-${SHORT_SHA}', '-f', 'examples/functions/golang/manual-simple/Dockerfile', '.']
-  waitFor: ['Docker Image: open-match-base-build']
+- id: 'Build: Clean'
+  name: open-match-build
+  args: ['make', 'clean']
+  waitFor: ['Docker Image: open-match-build']
```
yfei1: Would it make sense to push the open-match-build image to gcr and use it directly instead of building the base environment image per PR? Or we can use [Kaniko cache](https://cloud.google.com/cloud-build/docs/kaniko-cache) to rebuild the base image with expiration time in case any build time dependencies have an update.

cloudbuild.yaml
```diff
@@ -47,209 +47,98 @@
 #    https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/go
 
 steps:
-- id: 'Docker Image: open-match-base-build'
+- id: 'Docker Image: open-match-build'
   name: gcr.io/cloud-builders/docker
-  args: ['build', '-t', 'open-match-base-build', '-f', 'Dockerfile.base-build', '.']
+  args: ['build', '--build-arg', 'GCP_PROJECT_ID=$PROJECT_ID', '-t', 'open-match-build', '-f', 'Dockerfile', '.']
```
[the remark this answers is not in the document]
yfei1: Correct me if I am wrong :) Could we use something like docker build -t open-match-build - < Dockerfile since it seems like the base build does not requires any context?

cloudbuild.yaml
jeremyje: Kaniko cache looks like it would make much more sense here.

jeremyje: I hit https://github.com/GoogleContainerTools/kaniko/issues/477 immediately.

yfei1: I think it would be ideal to have a caching system for the CI base image. I tried to build just the base image on my workstation yesterday and it took me ~10.5 minutes.

jeremyje: I removed the C++ build stuff just now that should improve it significantly.

PR #190, opened 2019-04-10
Update install yaml files

jeremyje:
https://github.com/GoogleCloudPlatform/open-match/issues/161

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ LOCAL_CLOUD_BUILD_PUSH = # --push
 KUBECTL_RUN_ENV = --env='REDIS_SERVICE_HOST=$$(OM_REDIS_MASTER_SERVICE_HOST)' --env='REDIS_SERVICE_PORT=$$(OM_REDIS_MASTER_SERVICE_PORT)'
 GCP_LOCATION_FLAG = --zone $(GCP_ZONE)
 # Flags to simulate behavior of newer versions of Kubernetes
-KUBERNETES_COMPAT = --no-enable-basic-auth --no-issue-client-certificate --enable-ip-alias
+KUBERNETES_COMPAT = --no-enable-basic-auth --no-issue-client-certificate --enable-ip-alias --metadata disable-legacy-endpoints=true --enable-autoupgrade
 GO111MODULE = on
 PROMETHEUS_PORT = 9090
 GRAFANA_PORT = 3000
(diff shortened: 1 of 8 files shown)


install/yaml/02-open-match.yaml
```diff
@@ -204,7 +204,7 @@ spec:
     spec:
       containers:
       - name: om-backend
-        image: "gcr.io/open-match-public-images/openmatch-backendapi:0.4.0"
+        image: "gcr.io/jeremyedwards-gaming-dev/openmatch-backendapi:0.4.0-fc52ef6"
```
yfei1: Is there a specific reason that we are not using the public images for all open match components?

Makefile
```diff
@@ -631,24 +646,32 @@ clean-nodejs:
...
 
 run-clientloadgen: build/toolchain/bin/kubectl$(EXE_EXTENSION)
-	$(KUBECTL) run om-clientloadgen --rm --restart=Never --image-pull-policy=Always -i --tty --image=$(REGISTRY)/openmatch-clientloadgen:$(TAG) --namespace=$(NAMESPACE) $(KUBECTL_RUN_ENV)
+	$(KUBECTL) run om-clientloadgen --rm --restart=Never --image-pull-policy=Always -i --tty --image=$(REGISTRY)/openmatch-clientloadgen:$(TAG) --namespace=$(OPEN_MATCH_KUBERNETES_NAMESPACE) $(KUBECTL_RUN_ENV)
 
 proxy-grafana: build/toolchain/bin/kubectl$(EXE_EXTENSION)
 	echo "User: admin"
 	echo "Password: openmatch"
-	$(KUBECTL) port-forward --namespace $(NAMESPACE) $(shell $(KUBECTL) get pod --namespace $(NAMESPACE) --selector="app=grafana,release=$(OPEN_MATCH_NAME)" --output jsonpath='{.items[0].metadata.name}') $(GRAFANA_PORT):3000 $(PORT_FORWARD_ADDRESS_FLAG)
+	$(KUBECTL) port-forward --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) $(shell $(KUBECTL) get pod --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) --selector="app=grafana,release=$(OPEN_MATCH_CHART_NAME)" --output jsonpath='{.items[0].metadata.name}') $(GRAFANA_PORT):3000 $(PORT_FORWARD_ADDRESS_FLAG)
```
yfei1: Would it make more sense to use something like $(GRAFANA_EXTERNAL_PORT):$(GRAFANA_INTERNAL_PORT) to avoid using the port number 3000 and 9090 directly? This could make the file more self-explained.

jeremyje: I'll do this in a different PR. I fixed the chart names and some namespaces because it's getting out hand for now.

Issue #180, opened 2019-04-10
Improvements to Makefile build commands


yfei1:
Currently, the Makefile 'go get' the required packages and then move the installed binaries directly to 'build/toolchain/bin' directory. In this case, users with required packages installed (e.g. protoc-gen-go) before will have to manually re-install their packages. We should figure out a way to mitigate this issue.

jeremyje:
I looked around for a workaround for this. I don't think there's a reasonable one that doesn't cause potential disruption. You can't simply delete the files because they have been already placed there by other project.

yfei1:
Resolved om #191

PR #191, opened 2019-04-11
Fix build artifacts

yfei1:
This fix issue #180. go build is able to download a temporary source package and build the binaries into -pkgdir directory, which isolates the build environment from users' environment.

Also fixed the REPOSITORY_ROOT variable. This variable used to contains a trailing forward slash e.g. 'src/', which makes BUILD_DIR turns to an invalid path 'src//build'.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -448,7 +448,8 @@ build/toolchain/bin/protoc$(EXE_EXTENSION):
 	rm $(TOOLCHAIN_DIR)/protoc-temp.zip $(TOOLCHAIN_DIR)/readme.txt
 
 build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION):
-	cd build/toolchain/bin && $(GO) build -pkgdir . github.com/golang/protobuf/protoc-gen-go
+	mkdir -p $(TOOLCHAIN_BIN)
+	cd $(TOOLCHAIN_BIN) && $(GO) build -pkgdir . github.com/golang/protobuf/protoc-gen-go
 
 build/toolchain/bin/protoc-gen-grpc-gateway$(EXE_EXTENSION):
 	mkdir -p $(TOOLCHAIN_DIR)/googleapis-temp/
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -448,8 +448,7 @@ build/toolchain/bin/protoc$(EXE_EXTENSION):
 	rm $(TOOLCHAIN_DIR)/protoc-temp.zip $(TOOLCHAIN_DIR)/readme.txt
 
 build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION):
-	$(GO) get github.com/golang/protobuf/protoc-gen-go
-	mv $(GO_BIN)/protoc-gen-go$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION)
+	cd build/toolchain/bin && $(GO) build -pkgdir . github.com/golang/protobuf/protoc-gen-go
```
jeremyje: cd $(REPOSITORY_ROOT)/build/

Makefile
```diff
@@ -448,8 +448,7 @@ build/toolchain/bin/protoc$(EXE_EXTENSION):
 	rm $(TOOLCHAIN_DIR)/protoc-temp.zip $(TOOLCHAIN_DIR)/readme.txt
 
 build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION):
-	$(GO) get github.com/golang/protobuf/protoc-gen-go
-	mv $(GO_BIN)/protoc-gen-go$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION)
```
jeremyje: Missing the mkdir -p $(REPOSITORY_ROOT)/build/toolchain/bin/

`make clean build/toolchain/bin/protoc-gen-go` won't work otherwise.

yfei1:
The download behavior of this go build command is undocumented yet and I think it was shipped with go1.11+. Tested on my personal laptop with go version 1.10 installed, directly invoke 'go build' command without downloading first failed - error message: unable to find the source package under GOPATH.

PR #193, opened 2019-04-11
Update py3 MMF example to use Request/Response protos

ihrankouski:
This fixes py3 MMF example which was not updated after https://github.com/GoogleCloudPlatform/open-match/commit/8db062f7b9e09eb09bfdad3e32e7c6b23d285356


ihrankouski:
I didn't check that but it turns out that PHP example must be broken as well...

jeremyje:
Yeah there's a bug somewhere discussing this. PHP is probably going to stay broken for a while because it's very non-trivial to fix. Also PHP adds a lot of time to the build because it has no pre-built binaries.

PR #194, opened 2019-04-11
Ignore files for Mini Match

jeremyje:
https://github.com/GoogleCloudPlatform/open-match/issues/115


.dockerignore
```diff
@@ -1,4 +1,87 @@
...
+msbuild.wrn
+
+# Visual Studio 2015
+.vs/
+
+# Goland
+.idea/
+
+# Nodejs files placed when building Hugo, ok to allow if we actually start using Nodejs.
+package.json
```
yfei1: I have not used Hugo before, just wondering if the developers are still able to install the required npm modules without this package.json file

jeremyje: I create an isolated install of nodejs just for Hugo. Installing the node_modules/ does some funny things we'd like to avoid bleeding elsewhere. If we start requiring nodejs we can have a checked in node_modules/ directory and just use that but for now it doesn't make sense to leak buidl context.

PR #197, opened 2019-04-11
Add reduced permissions for mmforc service account.

jeremyje:
Somewhat related to https://github.com/GoogleCloudPlatform/open-match/issues/105. We are reducing the permissions that Open Match jobs have.


install/helm/open-match/templates/mmforc-service-account.yaml
```diff
@@ -0,0 +1,54 @@
...
+roleRef:
+  kind: ClusterRole
+  name: {{ .Values.openmatch.mmforc.serviceAccount }}
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ .Values.openmatch.mmforc.serviceAccount }}
+  namespace: {{ .Release.Namespace }}
```
ihrankouski: I think ClusterRole doesn't need a namespace in it's definition, it's cluster-wide: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole

install/yaml/install.yaml
```diff
@@ -997,6 +997,60 @@ metadata:
...
+    heritage: Tiller
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: mmforc-job-runner
+subjects:
+- kind: ServiceAccount
+  name: mmforc-job-runner
+  namespace: open-match
```
ihrankouski: This should be in the same namespace as mmforc deployment for ServiceAccount to work I think.
MMFOrc is in "default" namespace currently (in yaml files). So one of these two should be changed to match the second.

PR #198, opened 2019-04-11
Unify gRPC server initialization

jeremyje:


internal/app/backendapi/apisrv/apisrv.go
```diff
@@ -51,68 +49,33 @@ import (
...
-// Open starts the api grpc service listening on the configured port.
-func (s *BackendAPI) Open() error {
-	ln, err := net.Listen("tcp", ":"+s.cfg.GetString("api.backend.port"))
-	if err != nil {
-		beLog.WithFields(log.Fields{
-			"error": err.Error(),
-			"port":  s.cfg.GetInt("api.backend.port"),
-		}).Error("net.Listen() error")
-		return err
+func New(omSrv *serving.OpenMatchServer) *BackendAPI {
```
Laremere: Will we be using the return value from New ever?  It seems the pattern is that it registers the rpc server, and the return value isn't used?  

If not, New doesn't seem the best name, but naming is hard and I don't have any good suggestions.  Register?  Start (though it isn't actually serving until start is called on the serving.openMatchServer, right?)?

Also, if the return value is never used, could it be backendAPI instead of BackendAPI?  Is there any reason to leak the type?


Here and for the other services too.

internal/app/backendapi/backendapi.go
```diff
@@ -23,83 +23,37 @@ limitations under the License.
...
-	// Logrus structured logging setup
-	beLogFields = log.Fields{
-		"app":       "openmatch",
-		"component": "backend",
-	}
-	beLog = log.WithFields(beLogFields)
-)
-
-func initializeApplication() (config.View, error) {
+// InitializeServer adds the frontend api server to the OpenMatchServer object.
```
Laremere: backend, not frontend

internal/app/backendapi/backendapi.go
```diff
@@ -23,83 +23,37 @@ limitations under the License.
...
+func InitializeServer() *serving.ServerParams {
```
Laremere: Is this really initializing the server?  It seems the binding inside the params is doing that?
Maybe ServerParams()?

Here and for the other services too.

internal/app/mmlogicapi/mmlogicapi.go
```diff
@@ -22,84 +22,36 @@ limitations under the License.
...
-	mlLogFields = log.Fields{
-		"app":       "openmatch",
-		"component": "mmlogic",
-	}
-	mlLog = log.WithFields(mlLogFields)
-)
-
-func initializeApplication() (config.View, error) {
-	// Logrus structured logging initialization
+// InitializeServer adds the frontend api server to the OpenMatchServer object.
```
Laremere: Not frontend server.

internal/app/backendapi/apisrv/apisrv.go
```diff
@@ -51,68 +49,33 @@ import (
...
-func (s *BackendAPI) Open() error {
-	ln, err := net.Listen("tcp", ":"+s.cfg.GetString("api.backend.port"))
-	if err != nil {
-		beLog.WithFields(log.Fields{
-			"error": err.Error(),
-			"port":  s.cfg.GetInt("api.backend.port"),
-		}).Error("net.Listen() error")
-		return err
+// Bind binds the gRPC endpoint to OpenMatchServer
+func Bind(omSrv *serving.OpenMatchServer) error {
```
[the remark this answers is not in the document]
Laremere: sorry, misread.  This looks right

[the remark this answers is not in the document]
jeremyje: Right now it's not used but there will be times in the future like reading a config where bind can fail. This is when we also start adding worker go routines as well.

internal/app/backendapi/backendapi.go
```diff
@@ -23,83 +23,36 @@ limitations under the License.
...
-		}).Error("Unable to load config file")
-		return nil, err
+	return &serving.ServerParams{
+		BaseLogFields: log.Fields{
+			"app":       "openmatch",
+			"component": "backend",
+		},
+		PortConfigName:     "api.backend.port",
+		CustomMeasureViews: apisrv.DefaultBackendAPIViews,
+		Bindings: []serving.BindingFunc{
```
Laremere: Just do
`
Bindings: []server.BindingFunc{
  apisrv.Bind,
},
?

PR #200, opened 2019-04-12
Add temporary redirect site for Open Match

jeremyje:
Staging for https://github.com/GoogleCloudPlatform/open-match/issues/108


Makefile
```diff
@@ -618,6 +618,9 @@ browse-site: build/site/
 deploy-dev-site: build/site/
 	cd $(BUILD_DIR)/site && gcloud $(OM_SITE_GCP_PROJECT_FLAG) app deploy .app.yaml --promote --version=$(VERSION_SUFFIX) --quiet
 
+deploy-redirect-site:
```
Laremere: Should this be a phony?  I think you added phonys in a different uncommitted change?

PR #204, opened 2019-04-15
Make matchmaker_config.yaml a part of Helm chart config

ihrankouski:
Aa part of #130 file `/config/matchmaker_config.yaml` is removed along with all its symlinks. Instead, its content is stored inside `/install/helm/open-match/values.yaml`. At the time of chart installation (or just rendering) one may override various parts of matchmaker config using [`--set`](https://github.com/helm/helm/blob/master/docs/using_helm.md#the-format-and-limitations-of---set):
```sh
helm template open-match \
  --set openmatch.config.files."matchmaker_config\.yaml".logging.level=warn \
  --set openmatch.config.files."matchmaker_config\.yaml".metrics.reportingPeriod=15 \
  --set openmatch.config.files."matchmaker_config\.yaml".playerIndices="{some,thing,there}" \
```

Also moved config package under internal/.


ihrankouski:
Data race in Viper. And it turns out it always has been here...

jeremyje:
I'm also seeing a data race in Prometheus as well. This one was uncovered because of minimatch. https://github.com/GoogleCloudPlatform/open-match/issues/205 the Prometheus one isn't deterministic though so it will will commonly bypass the CI tests.

jeremyje:
This is unfortunate we might have to remove the -race flag until the situation improves overall. I'll create a bug for this to discuss.

jeremyje:
Has the situation improved? Also where's the split out for moving config to internal/config/?

jeremyje:
Can we move config/ first before the split? We are making serious changes in 0.6.0 and race conditions are a serious regression.

ihrankouski:
There's no split in this PR, only move of config/. Race condition has been there since the first adding of `cfg.WatchConfig()` call, but was only uncovered with tests recently.

PR #209, opened 2019-04-15
Move GOPROXY=off to CI only.

jeremyje:
Resolves https://github.com/GoogleCloudPlatform/open-match/issues/212
The root cause is that we disable downloading of go modules during the go build process in an effort to avoid races if 2 binaries are built at the same time. This causes other issues like this one so we'll enable it for now.


Laremere:
Please add to the description why this change is being made.  Thanks!

PR #215, opened 2019-04-16
Fix port allocation race condition during tests.

jeremyje:
This is done by now keeping the port open while calling next port library. This prevents races were 2 processes think the same port is available during the old model which was check open, close, blind reopen.

This change replaces nextport with using :0 as the random port number as well as providing a class to get the port only once to prevent multiple threads from using it.

Also increasing the test runs to 25x from 10x to reduce the risk of introducing more flakes.

Resolves https://github.com/GoogleCloudPlatform/open-match/issues/205.
Resolves https://github.com/GoogleCloudPlatform/open-match/issues/213


yfei1:
I found [this article](https://stackoverflow.com/questions/43424787/how-to-use-next-available-port-in-http-listenandserve) might help to resolve this issue.

jeremyje:
That's useful for obsolete nextport but the plumbing parts still are relevant.

internal/port/port.go
```diff
@@ -0,0 +1,65 @@
...
+	if p.listener != nil {
+		return p.listener.Close()
+	}
+	return nil
+}
+
+// PortFromNumber opens a TCP based on the port number provided.
+func PortFromNumber(port int) (*Port, error) {
+	addr := "0.0.0.0"
+	conn, err := net.Listen("tcp", addr)
```
yfei1: The port is not used?

jeremyje:
I've removed nextport since :0 does the same thing. It's now just reduced to single handoff port listener.

internal/util/netlistener/testing/randomport_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"fmt"
+	"testing"
+)
+
+const (
+	numIterations = 1000
+)
+
+func TestMustPort(t *testing.T) {
```
yfei1: MustListen

Issue #112, opened 2019-03-15
Add support for MMFs set up as a gRPC service.


sawagh:
Currently, Open Match can be configured with MMFs as containers that get run as K8s jobs. This mechanism does not scale well and will be deprecated. The recommended mechanism would be to host the MMF as a gRPC service and to pass along, the connection details of this service to Open Match. 

Open Match will prescribe the service proto and will create clients to this service and call into this service to execute MMFs.

Open Match will also provide a gRPC harness that handles the server setup bit, with instructions to stand up a gRPC server and samples of implementing the MMF that can plug into this harness.

PR #216, opened 2019-04-16
Serving GRPC Harness and example MMF scaffolding (#112) 

sawagh:
https://github.com/GoogleCloudPlatform/open-match/issues/112


examples/functions/golang/grpc-serving/doc.go
```diff
@@ -0,0 +1,16 @@
+/*
+FrontendAPI contains the unique files required to run the API endpoints for
```
jeremyje: This comment is all wrong.

examples/functions/golang/grpc-serving/doc.go
```diff
@@ -0,0 +1,16 @@
...
+
+Note that the main package for frontendapi does very little except read the
+config and set up logging and metrics, then start the server.  Almost all the
+work is being done by frontendapi/apisrv, which implements the gRPC server
+defined in the frontendapi/proto/frontend.pb.go file.
+
+<TODO - UPDATE THIS>
+*/
+
+package main
```
jeremyje: I cannot comment the binary but please don't check in binaries.
Add "examples/functions/golang/grpc-serving/grpc-serving" to .gitignore and friends.
Rebase to get rid of the binary from this change.

Makefile
```diff
@@ -207,6 +207,10 @@ push-mmf-go-mmlogic-simple-image: build-mmf-go-mmlogic-simple-image
 	docker push $(REGISTRY)/openmatch-mmf-go-mmlogic-simple:$(TAG)
 	docker push $(REGISTRY)/openmatch-mmf-go-mmlogic-simple:$(ALTERNATE_TAG)
 
+push-mmf-go-grpc-serving-simple-image: build-mmf-go-grpc-serving-simple-image
```
jeremyje: cloudbuild.yaml is missing this as an image artifact.

internal/harness/matchfunction/golang/apisrv/apisrv.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+// This is the function signature for the function to be implemented by the user.
+// The harness will pass the Rosters and PlayerPool for the match profile to this
+// function and it will return the Rosters to be populated in the proposal.
+type MatchFunction func(context.Context, string, []*pb.Roster, []*pb.PlayerPool) (string, []*pb.Roster, error)
+
+// MatchFunctionServer implements pb.MatchFunctionServer, the server generated
+// by compiling the protobuf, by fulfilling the pb.MatchFunctionServer interface.
+type MatchFunctionServer struct {
```
jeremyje: What's the difference between this an OpenMatchServer? Couldn't we just use the standard grpc binding stuff here and add a wrapper for MatchFunction?

internal/harness/matchfunction/golang/apisrv/apisrv.go
```diff
@@ -0,0 +1,53 @@
...
+	Config config.View
+	Logger *log.Entry
+	Func   MatchFunction
+}
+
+func (*MatchFunctionServer) Run(context.Context, *pb.RunRequest) (*pb.RunResponse, error) {
+	// TODO: Add Harness code here to perform the following:
```
jeremyje: Just to verify this is a TODO for this package not the examples/ that we need to provide to users.

Makefile
sawagh: Added now.

internal/harness/matchfunction/golang/apisrv/apisrv.go
sawagh: This does not need redis pool initialization - also this one needs to pass the MatchFunction to the server. I started with using the OpenMatchServer but mucking with its internals to add things specific to harness didnt seem right. This could possibly be refactored back into serving package as a HarnessServer separating out the construction of the params and sharing some code between the types of servers - may be for the future refactor in 0.6

internal/harness/matchfunction/golang/apisrv/apisrv.go
sawagh: Yes - to clarify all these steps will be code in this harness - the Step 3 is where the harness will run the Func that the example passes in to get proposal results.

examples/functions/golang/grpc-serving/doc.go
sawagh: I thought i deleted this file (currently examples currently dont have doc files). I have now deleted it.

examples/functions/golang/grpc-serving/doc.go
sawagh: Ah. Thanks for pointing out - accidental - added to gitignore too now.

sawagh:
> Missing .gcloudignore as well.

Added to .dockerignore - you meant that one right? .gcloudignore seems to just import .gitignore

PR #225, opened 2019-04-17
Prevent sudo on Makefile for commands that require auth.

jeremyje:
Fixes https://github.com/GoogleCloudPlatform/open-match/issues/164


yfei1:
It seems like GCB runs the build as root so this patch can't pass the cloud build.. Instead of preventing sudo invocations on Makefile, maybe we should just put it in a FAQ section under the installation guide. ;(

    echo "ERROR: Running Makefile as root (or sudo)"
    ERROR: Running Makefile as root (or sudo)
    echo "Please follow the instructions at https://docs.docker.com/install/linux/linux-postinstall/ if you are trying to sudo run the Makefile because of the 'Cannot connect to the Docker daemon' error."
    Please follow the instructions at https://docs.docker.com/install/linux/linux-postinstall/ if you are trying to sudo run the Makefile because of the 'Cannot connect to the Docker daemon' error.
    echo "NOTE: sudo/root do not have the authentication token to talk to any GCP service via gcloud."
    NOTE: sudo/root do not have the authentication token to talk to any GCP service via gcloud.
    exit 1
    Makefile:746: recipe for target 'no-sudo' failed
    make: *** [no-sudo] Error 1

jeremyje:
I'm working on a bypass for this. This is somewhat more complicated because it partially uses the openmatch user from Dockerfile.ci as well. So I'm removing that as part of this change most likely.

jeremyje:
> It seems like GCB runs the build as root so this patch can't pass the cloud build.. Instead of preventing sudo invocations on Makefile, maybe we should just put it in a FAQ section under the installation guide. ;(
> 
> ```
> echo "ERROR: Running Makefile as root (or sudo)"
> ERROR: Running Makefile as root (or sudo)
> echo "Please follow the instructions at https://docs.docker.com/install/linux/linux-postinstall/ if you are trying to sudo run the Makefile because of the 'Cannot connect to the Docker daemon' error."
> Please follow the instructions at https://docs.docker.com/install/linux/linux-postinstall/ if you are trying to sudo run the Makefile because of the 'Cannot connect to the Docker daemon' error.
> echo "NOTE: sudo/root do not have the authentication token to talk to any GCP service via gcloud."
> NOTE: sudo/root do not have the authentication token to talk to any GCP service via gcloud.
> exit 1
> Makefile:746: recipe for target 'no-sudo' failed
> make: *** [no-sudo] Error 1
> ```

This is fixed now. And the build is simpler now because there's no openmatch user.

Issue #167, opened 2019-04-09
Reduce log spam of Open Match


jeremyje:
A common complaint is Open Match produces a lot of logs that aren't really relevant to debugging or usage. We should audit the log output and see what can be removed or demoted in importance.

jeremyje:
@Laremere Can we close this issue?

PR #234, opened 2019-04-18
Reduce log spam in backendapi

Laremere:
For issue #167
The general strategy here is:
- Errors where things went wrong
- Info for request received, and then finished.
- Debug for things which might be useful.
- Trace for log statements which include lots of debug detail, or are just "I got to this spot" logs.


yfei1:
It seems like the changes only apply to the CreateMatch function?

internal/app/backendapi/apisrv/apisrv.go
```diff
@@ -87,16 +87,14 @@ func (s *backendAPI) CreateMatch(c context.Context, req *pb.CreateMatchRequest)
...
-		beLog.Info("contents exist?", gjson.Get(profile.Properties, s.cfg.GetString("jsonkeys.pools")).Exists())
-	*/
+	// Debugging logs
+	beLog.Trace("Pools nil? ", (profile.Pools == nil))
+	beLog.Trace("Pools empty? ", (len(profile.Pools) == 0))
+	beLog.Trace("Rosters nil? ", (profile.Rosters == nil))
+	beLog.Trace("Rosters empty? ", (len(profile.Rosters) == 0))
+	beLog.Trace("config set for json.pools?", s.cfg.IsSet("jsonkeys.pools"))
+	beLog.Trace("contents key?", s.cfg.GetString("jsonkeys.pools"))
+	beLog.Trace("contents exist?", gjson.Get(profile.Properties, s.cfg.GetString("jsonkeys.pools")).Exists())
```
yfei1: Would it make more sense to add the debugging logs for other functions if we wanna enable it here?

jeremyje: If the code was commented out I'd just delete it. No one is going to miss it.

Laremere: For this, I'm not trying to change the logging, just re-prioritize the current logging to reduce spam.  We're going to be changing so much for 0.6 that I don't think spending time improving the quality of the logs is worth it.

Laremere:
> It seems like the changes only apply to the CreateMatch function?

I reviewed the whole file, and only changes were in the CreateMatch.

Issue #233, opened 2019-04-17
Remove unnecessary binary from repo


yfei1:
There is a binary file under examples/functions/golang/grpc-serving directory. It is not a good practice to check-in binary files which will lead to an exploding commit history. We should probably rollback this change and unstage the binary file.

sawagh:
Fixed this in (#236)

PR #236, opened 2019-04-18
Example MMF demonstrating the use of the GRPC harness (#233) (#112)

sawagh:
https://github.com/GoogleCloudPlatform/open-match/issues/112


examples/functions/golang/grpc-serving/main.go
```diff
@@ -36,7 +48,88 @@ func main() {
...
+					rosters[ti].Players[si] = selectedPlayer
+					selectedPlayerCount++
+					break
+					/////////////////////////////////////////////////////////
+
+				} else {
+					// Weren't enough players left in the pool to fill all the
+					// slots.  If your game can handle non-full teams, feel
+					// free to customize this part as you see fit.
+					return "{\"error\": \"insufficient players\"}", rosters, errors.New("insufficient players")
```
jeremyje: Why are we sending an error as JSON here? If anything the error text should be extracted from the error variable. It's go convention to return "" for string or nil for structs.

sawagh: Good catch. The harness itself handles it like how you stated - if there is an error, pulls error text out to return on the match error (does not use this string) - so fixed this! I ve updated the comment to indicate clearly that if error is returned in the error variable - then this field is ignored and should be "".

PR #238, opened 2019-04-18
Evaluator Harness and Sample golang Evaluator

sawagh:
This PR implements a generic Evaluator harness that abstracts all the Open Match specific data concepts (Redis reads etc.) away from the user's evaluation function. In future releases, these will change completely and the user logic should still be minimally impacted. 

The PR also implements a sample barebones evaluator example that simply de-collides proposals.

### DIFF
--- a/config/matchmaker_config.yaml
+++ b/config/matchmaker_config.yaml
@@ -6,76 +6,77 @@ logging:
   format: text
   source: false
 
-api: 
-  backend: 
+api:
+  backend:
     hostname: om-backendapi
     port: 50505
     backoff: "[2 32] *2 ~0.33 <30"
     proxyport: 51505
-  frontend: 
+  frontend:
     hostname: om-frontendapi
     port: 50504
     backoff: "[2 32] *2 ~0.33 <300"
     proxyport: 51504
-  mmlogic: 
+  mmlogic:
     hostname: om-mmlogicapi
     port: 50503
     proxyport: 51503
   functions:
     port: 50502
     proxyport: 51502
-  
-evalutor:
+
+evaluator:
   # Evaluator intervals are in milliseconds
+  interval: 10
   pollIntervalMs: 1000
   maxWaitMs: 10000
 
-metrics: 
+metrics:
   port: 9555
   endpoint: /metrics
   reportingPeriod: 5
 
-queues: 
-  profiles: 
+queues:
+  profiles:
     name: profileq
     pullCount: 100
-  proposals: 
+  proposals:
     name: proposalq
-  
-ignoreLists: 
-  proposed: 
+
+ignoreLists:
+  proposed:
     name: proposed
     offset: 0
     duration: 800
-  deindexed: 
+  deindexed:
     name: deindexed
     offset: 0
     duration: 800
-  expired: 
+  expired:
     name: OM_METADATA.accessed
     offset: 800
-    duration: 0 
+    duration: 0
 
-defaultImages: 
-  evaluator: 
+defaultImages:
+  evaluator:
     name: gcr.io/open-match-public-images/openmatch-evaluator-simple
     tag: latest
-  mmf: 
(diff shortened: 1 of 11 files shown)


config/matchmaker_config.yaml
```diff
@@ -25,8 +25,10 @@ api:
     port: 50502
     proxyport: 51502
   
-evalutor: 
-  interval: 10
```
jeremyje: Will this break existing examples? Also "evaluator"

examples/evaluators/golang/serving/main.go
```diff
@@ -0,0 +1,125 @@
...
+// Output:
+//  - (proposals) : List of approved proposal IDs that can be returned as match results.
+func Evaluate(ctx context.Context, proposals []*pb.MatchObject) ([]string, error) {
+	// Map of approved and overloaded proposals. Using maps for easier lookup.
+	approvedProposals := make(map[string]bool)
+	overloadedProposals := make(map[string]bool)
+
+	// Map of all the players encountered in the proposals. Each entry maps a player id to
+	// the first match in which the player was encountered.
+	allPlayers := make(map[string]string)
```
jeremyje: allPlayers := map[string]string{}

examples/evaluators/golang/serving/main.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// Iterate over each proposal to either add to approved map or overloaded map.
+	for _, proposal := range proposals {
+		proposalID := proposal.Id
+		approved := true
+		players := getPlayersInProposal(proposal)
+		// Iterate over each player in the proposal to check if the player was encountered before.
+		for _, pID := range players {
```
jeremyje: Because there's player and proposal pID is ambiguous. playerID is preferred here.

internal/harness/evaluator/golang/apisrv/apisrv.go
```diff
@@ -0,0 +1,296 @@
...
+	logger := s.Logger
+	logger.Info("Starting evaluation")
+
+	// Get connection to redis
+	redisConn := s.Pool.Get()
+	defer redisConn.Close()
+
+	start := time.Now()
+	// Get the number of proposals currently queued for evaluation.
+	proposalq := s.Config.GetString("queues.proposals.name")
```
jeremyje: Not obvious that this is a name of the proposal queue. queueName?

config/matchmaker_config.yaml
sawagh: Nope. :) It was already misspelt and already kinda broken :')  I ve added it back and now I think the existing example evaluator will behave different :p thats gonna be deprecated though in the next couple PRs.

examples/evaluators/golang/serving/main.go
sawagh: Good idea. Done!

internal/harness/evaluator/golang/apisrv/apisrv.go
sawagh: Done!

examples/evaluators/golang/serving/main.go
sawagh: Done. Others too!

internal/harness/evaluator/golang/harness.go
```diff
@@ -0,0 +1,145 @@
...
+	if len(host) == 0 {
+		return nil, fmt.Errorf("Failed to get hostname for MMLogicAPI from the environment")
+	}
+
+	port := cfg.GetString("api.mmlogic.port")
+	if len(port) == 0 {
+		return nil, fmt.Errorf("Failed to get port for MMLogicAPI from the environment")
+	}
+
+	ip, err := net.LookupHost(host)
```
jeremyje: This shouldn't be necessary. Lookup by grpc.

PR #245, opened 2019-04-19
Add option for frontendclient to loop forever.

jeremyje:
https://github.com/GoogleCloudPlatform/open-match/issues/227


sawagh:
I think the existing frontend client as currently implemented is very flaky - it generates random players and in most cases where I ran it, if the player does not fit the hardcoded match request from the current backend client tool - the player would sit forever waiting for assignment. Given that it struggles to get single player assignments, it is not the best suited to cycle currently. It has been removed from install-examples helm for this reason. 

You can check in the change if you want to add the future looping functionality but as the tool as it currently stands itself is broken.

PR #249, opened 2019-04-19
Remove deprecated examples, evaluator and mmforc

sawagh:
This PR is still a work in progress - Creating it just so its easy to share the artificats that will be deprecated. In summary, the following:
- mmforc
- the simple evaluator
- all but the serving example function

A LOT of documentation needs to change along with this to avoid confusion. Will take that on once the code changes are through.


jeremyje:
I don't think this PR should be out yet let's wait until all the serving code is put up before we put this up for review.

sawagh:
I have not put this up for review yet (hence it is not assigned to anyone
and there is a comment in the PR clarifying that).

sawagh:
Apologize for the confusion earlier - Now the changes are ready to be reviewed (only deleting code in this PR)

PR #253, opened 2019-04-19
Changes to make the demo steps easier

sawagh:


examples/backendclient/Dockerfile
```diff
@@ -19,5 +19,6 @@ RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo .
 
 FROM gcr.io/distroless/static  
 COPY --from=builder /go/src/github.com/GoogleCloudPlatform/open-match/examples/backendclient/backendclient .
+COPY --from=builder /go/src/github.com/GoogleCloudPlatform/open-match/examples/backendclient/profiles profiles
```
jeremyje: Use something like this, https://stackoverflow.com/questions/52488316/kubectl-run-with-env-var-from-a-secret-config.

PR #260, opened 2019-04-20
Add proxy tests

yfei1:


internal/serving/grpcserver.go
```diff
@@ -40,13 +42,38 @@ func (gw *GrpcWrapper) AddService(handlerFunc func(*grpc.Server)) {
...
+			gw.logger.WithFields(log.Fields{
+				"error": err.Error(),
+				"ops":   stub.ops,
+				"pool":  stub.pool,
+			}).Error("runtime.NewPattern failed to register /ping")
+			return nil, err
+		}
+
+		mux.Handle(stub.method, pat, func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
+			io.WriteString(w, "pong")
```
golangcibot: Error return value of `io.WriteString` is not checked (from `errcheck`)

internal/serving/grpcserver.go
```diff
@@ -40,13 +42,42 @@ func (gw *GrpcWrapper) AddService(handlerFunc func(*grpc.Server)) {
...
+// AddProxy registers a reverse proxy from REST to gRPC when server is created.
 func (gw *GrpcWrapper) AddProxy(proxyHandlerFunc func(context.Context, *runtime.ServeMux, string, []grpc.DialOption) error) {
 	gw.proxyHandlerFunc = func(endpoint string) (*runtime.ServeMux, error) {
 		ctx := context.Background()
 		ctx, cancel := context.WithCancel(ctx)
 		defer cancel()
 
 		mux := runtime.NewServeMux()
+
+		type stubPattern struct {
```
jeremyje: Why are we faking a real grpc server with a semi-mocked http server? You should be able to use the real server in the e2e tests and be good.

internal/serving/testing/minimatch_test.go
```diff
@@ -39,15 +42,68 @@ func TestNewMiniMatch(t *goTesting.T) {
...
+			endpoint: "ping",
+			response: "pong",
+		},
 	}
-	if result == nil {
-		t.Errorf("insert player was not successful %v", result)
+
+	for _, stub := range testStubs {
+		t.Run(fmt.Sprintf("ProxyTest-%s-%s", stub.method, stub.endpoint), func(t *goTesting.T) {
+			apiAddr := fmt.Sprintf("http://:%d/%s", mm.Config.GetInt("api.frontend.proxyport"), stub.endpoint)
```
jeremyje: It's better to return an http.Client along with a base URL. The reason why is because in a later change we'll be providing TLS support so the http.Client will need to be authenticated first before we can use it and it'd be best not have the tests be aware of that.

internal/serving/testing/minimatch.go
```diff
@@ -53,31 +53,28 @@ func (mm *MiniMatchServer) Stop() {
...
 	if err != nil {
 		panic(err)
 	}
-	return mm, closer
+	return mm
 }
 
 // NewMiniMatch creates and starts an OpenMatchServer context for testing.
-func NewMiniMatch(params []*serving.ServerParams) (*MiniMatchServer, func(), error) {
+func NewMiniMatch(params []*serving.ServerParams) (*MiniMatchServer, error) {
```
jeremyje: Why is this being removed? This leaks a grpc server, redis, etc..

yfei1: Correct me if I am wrong. I thought calling `defer mm.Stop()` in https://github.com/GoogleCloudPlatform/open-match/blob/master/internal/serving/testing/minimatch_test.go#L33 is essentially the same as `defer closer()` https://github.com/GoogleCloudPlatform/open-match/blob/master/internal/serving/testing/minimatch_test.go#L41.

jeremyje: That's the only test that calls mm.Stop() directly the others use closer(). This test calls mm.Stop() to verify that the server is shutdown.

PR #261, opened 2019-04-22
make clean now deletes build/ (#262)

jeremyje:
https://github.com/GoogleCloudPlatform/open-match/issues/262

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -643,8 +643,14 @@ clean-binaries:
 	rm -rf test/cmd/clientloadgen/clientloadgen
 	rm -rf test/cmd/frontendclient/frontendclient
 
+clean-build: clean-toolchain clean-archives
+	rm -rf build/
+
 clean-toolchain:
 	rm -rf build/toolchain/
+
+clean-archives:
+	rm -rf build/archives/
 
 clean-nodejs:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -700,5 +706,5 @@ ifeq ($(shell whoami),root)
 endif
 endif
 
-.PHONY: docker gcloud deploy-redirect-site sync-deps sleep-10 proxy-dashboard proxy-prometheus proxy-grafana clean clean-toolchain clean-binaries clean-protos presubmit test test-in-ci vet
+.PHONY: docker gcloud deploy-redirect-site sync-deps sleep-10 proxy-dashboard proxy-prometheus proxy-grafana clean clean-build clean-toolchain clean-archives clean-binaries clean-protos presubmit test test-in-ci vet
```
sawagh: Looks like clean-archives is redundant here as clean-build already cleans the archive folder. Any case of only cleaning up archive without cleaning the build?

jeremyje: I added it just for consistency.

jeremyje: Oh wait this is a .PHONY target call out. It needs to be here to tell make it's a .PHONY.

PR #263, opened 2019-04-22
Automatically publish development.open-match.dev on Post Commit (#201)

jeremyje:
http://development.open-match.dev is the endpoint of the latest website. We should automatically push post commit CI builds to it.


api/protobuf-spec/backend.swagger.json
```diff
@@ -196,6 +196,36 @@
             "in": "query",
             "required": false,
             "type": "string"
+          },
+          {
+            "name": "mmfcfg.name",
```
sawagh: Are these supposed to be auto-generated ? How did i miss them when I made the proto change? (I remember seeing the change - and then after a make clean and proto rebuild, they did not regenrate)

jeremyje: Run `make presubmit` for every change. CI does this to at least verify that if you make an autogen change that might break something it'll do it but it doesn't verify that the old files have diffs. That turns out to be complex because /workspace loses it's git context.

Issue #270, opened 2019-04-23
Makefile for Windows


jeremyje:
There's already preliminary support for Windows/MINGW but it should be functional.

jeremyje:
There were a few changes that fixed this.

PR #272, opened 2019-04-23
Makefile for windows (#270)

jeremyje:
This change adds Windows support to the Makefile to build Open Match artifacts and use with Kubernetes.

Nodejs is still broken for now but that's for the website which isn't commonly used for development.


Makefile
```diff
@@ -451,7 +469,13 @@ build/archives/$(NODEJS_PACKAGE_NAME):
 
 build/toolchain/nodejs/: build/archives/$(NODEJS_PACKAGE_NAME)
 	mkdir -p build/toolchain/nodejs/
+ifeq ($(suffix $(NODEJS_PACKAGE_NAME)),.zip)
+	# TODO: This is broken, there's the node-v10.15.3-win-x64 directory also windows does not have the bin/ directory.
```
sawagh: So for windows, what does 'broken' mean from the perspective of build experience?

jeremyje: You can't build the rendered website. You can still use `make run-site` which is the Hugo preview mode. I verified that much.

Issue #269, opened 2019-04-23
issues with release process


Laremere:
Writing down all of the issues I encounter here:
-" Open a document for a draft release notes." link leads to 404.
-"The link resolves to, v{version}...master."  It resolves to v{PREVIOUS version}...master.  That whole sentence seems weird, just remove?  The instructions seemed good enough.
- The commits since last release links to commits, not the PRs.  Not all commits specify their PR.
- Search for merged pull request tagging should include " is:merged " to exclude pull requests which were abandoned.
- Unclear if all of the versions should be updated for release candidates.  Should the RC suffix be included in all of them?  I skipped the README.md for the RC.
- Skipped tagging, because the tags aren't specified.
- Tagging shouldn't be the job of the person doing the release.

- Why is it that one of the first steps is to open the release notes to start writing a draft, then later it says to create a draft release with the same template?
- cloudbuild.yaml needs to be included in places version number needs to be set.

- "git checkout master && git pull master && git rev-parse HEAD" is not the git command we want.  Instead: "git fetch upstream && git rev-parse upstream/master"
- "take note of the hash in git master" take note for what?
- I don't have permission to run post submit trigger.  However, it looks like it triggered automatically?
- Making the release notes is a lot of work, though you're working on this for 0.6 I think?

Post submit failed with: "Deploy: Deployment Configs": / [0 files][ 0.0 B/ 13.3 KiB] 
AccessDeniedException: 403 205790239254@cloudbuild.gserviceaccount.com does not have storage.objects.delete access to open-match-chart/install/yaml/master-latest/01-redis-chart.yaml.

.... I accidentally hit enter and made a release instead of it being a draft.

Laremere:
Ok, I managed to remove the accidentally submitted release by editing and pushing it back to draft, then deleting the tag.

jeremyje:
> -" Open a document for a draft release notes." link leads to 404.

Was this because it wasn't checked in? This seems to work against the repository now. Was there a specific way you were viewing it?

Laremere:
Try the link in the actual issue: https://github.com/GoogleCloudPlatform/open-match/issues/266
It's broken because the link is relative, and the issue is not in the same folder as the template.

Laremere:
> > The commits since last release links to commits, not the PRs. Not all commits specify their PR.
> 
> Going forward they should. There might be a few in v0.5.0 that do not but I think that's ok, just ignore. Is there in particular blocker here?
I think that will be fine.

[...]

PR #278, opened 2019-04-23
Include cloudbuild in places version needs to be set.  Specify rc for…

Laremere:
… all versions


Laremere:
This is the last item in #269

PR #290, opened 2019-04-25
Frontendapi load testing impl

yfei1:


Makefile
```diff
@@ -380,6 +380,11 @@ install-kubernetes-tools: build/toolchain/bin/kubectl$(EXE_EXTENSION) build/tool
 install-web-tools: build/toolchain/bin/hugo$(EXE_EXTENSION) build/toolchain/bin/htmltest$(EXE_EXTENSION) build/toolchain/nodejs/
 install-protoc-tools: build/toolchain/bin/protoc$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-grpc-gateway$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-swagger$(EXE_EXTENSION)
 
+install-stress-test-tools:
+	sudo apt-get install python3 python3-venv
```
jeremyje: Use the virtualenv install that we already create. Don't assume users have python. See the python protobuf gen targets.

jeremyje: Also just add the pip install to the python install you don't need a separate target.

yfei1: Our current Makefile doesnt have the python protobuf gen targets. I tried to install the virtualenv via `sudo apt-get install virtualenv python3-virtualenv` but it seems like nothing was installed into my bin directory. Virtualenv's official [installation guide](https://virtualenv.pypa.io/en/stable/installation/) seems like depends on either `pip` or `python`, that's why I explicitly list python3 as a requirement.

jeremyje: https://github.com/GoogleCloudPlatform/open-match/commit/4cbee9d8a7f5ccbbee15222e56fbde176fc3710c#diff-b67911656ef5d18c4ae36cb6741b7965

It was removed recently add it back in.

Makefile
```diff
@@ -567,6 +571,10 @@ ci-test:
 	$(GO) test ./... -race -test.count 25 -cover
 	$(GO) test ./... -run IgnoreRace$$ -cover
 
+stress-test-%: install-stress-test-tools
+	$(REPOSITORY_ROOT)/python/bin/locust -f $(REPOSITORY_ROOT)/test/stress/frontend.py --host=http://localhost:51504 \
```
jeremyje: Bring back the old style python virtualenv that lives in the build/python/ directory.

Makefile
```diff
@@ -381,6 +381,10 @@ install-web-tools: build/toolchain/bin/hugo$(EXE_EXTENSION) build/toolchain/bin/
 install-protoc-tools: build/toolchain/bin/protoc$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-grpc-gateway$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-swagger$(EXE_EXTENSION)
 install-openmatch-tools: build/toolchain/bin/certgen$(EXE_EXTENSION)
 
+install-stress-test-tools:
+	sudo apt-get install python3 virtualenv
```
jeremyje: This is problematic because any file dep will always rebuild.

Makefile
```diff
@@ -381,6 +381,18 @@ install-web-tools: build/toolchain/bin/hugo$(EXE_EXTENSION) build/toolchain/bin/
 install-protoc-tools: build/toolchain/bin/protoc$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-grpc-gateway$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-swagger$(EXE_EXTENSION)
 install-openmatch-tools: build/toolchain/bin/certgen$(EXE_EXTENSION)
 
+
+install-stress-test-tools:
```
jeremyje: Add the deleted code from https://github.com/GoogleCloudPlatform/open-match/commit/4cbee9d8a7f5ccbbee15222e56fbde176fc3710c#diff-b67911656ef5d18c4ae36cb6741b7965 back to replace this target. This way it's more cross platform. We purposely do not use apt-get in the makefile because that not going to work on Windows and Mac. It's ok to require users to download Python and I believe the instructions do already.

Using the virtualenv in the build directory allows us to have our own isolated environment of Python that does not depend on the user's environment and pollute their default configuration.

yfei1:
I'll add the build/toolchain/python requirement back to install-toolchain once we bring the python examples back.

Makefile
```diff
@@ -64,7 +64,7 @@ GOLANGCI_VERSION = 1.16.0
 PROTOC_RELEASE_BASE = https://github.com/protocolbuffers/protobuf/releases/download/v$(PROTOC_VERSION)/protoc-$(PROTOC_VERSION)
 GO = GO111MODULE=on go
 # Defines the absolute local directory of the open-match project
-REPOSITORY_ROOT := $(dir $(abspath $(MAKEFILE_LIST)))
+REPOSITORY_ROOT := $(realpath $(dir $(abspath $(MAKEFILE_LIST))))
```
jeremyje: What's the reason for this? It was causing issues in Windows.

Makefile
```diff
@@ -440,6 +440,11 @@ build/toolchain/bin/golangci-lint$(EXE_EXTENSION):
 	mv $(TOOLCHAIN_DIR)/temp-golangci/golangci-lint$(EXE_EXTENSION) $(TOOLCHAIN_BIN)/golangci-lint$(EXE_EXTENSION)
 	rm -rf $(TOOLCHAIN_DIR)/temp-golangci/
 
+build/toolchain/python/:
+	# mkdir -p $(TOOLCHAIN_DIR)/python/
```
jeremyje: Create the python directory.

Makefile
```diff
@@ -440,6 +440,11 @@ build/toolchain/bin/golangci-lint$(EXE_EXTENSION):
...
+	if ! [ -d "$(TOOLCHAIN_DIR)/python" ]; then virtualenv --python=python3 $(TOOLCHAIN_DIR)/python/; fi
```
jeremyje: There's lines missing from the old code that need to be re-added back.

Makefile
yfei1: using virtualenv will automatically create the python directory

Makefile
jeremyje: Why is there an if statement here then? If this rule is configured correctly it should only execute if the directory does not exist.

Makefile
```diff
@@ -440,6 +440,11 @@ build/toolchain/bin/golangci-lint$(EXE_EXTENSION):
...
+	$(TOOLCHAIN_DIR)/python/bin/pip install locustio
```
jeremyje: This is installing locustio globally not in the virtualenv.

Makefile
yfei1: without realpath the REPOSITORY_ROOT is something ending with a slash e.g. GoogleCloudPlatform/open-match/, which makes the local env variables defined in the Makefile becomes open-match//build - a path with two slashes. What is the issue in Windows?

jeremyje: It double adds the path in Windows so I had to remove it.

Makefile
yfei1: I use the pip inside the virtualenv so that we could install the python module directly without activating the virtualenv itself. It seems like it is doing the same thing as installing after activating the virtualenv.

yfei1: Done.

PR #291, opened 2019-04-25
Fix grammer in the open match demo script.

roberthbailey:
I created this PR against the release branch, but if it makes more sense to do it against master and cherry pick into the release branch let me know.


jeremyje:
Please don't submit this yet until https://github.com/GoogleCloudPlatform/open-match/pull/295 goes in. There's conflicts.

PR #299, opened 2019-04-25
Add ListenerHolder.AddrString() to avoid bad IPv6 assumptions. (#298) (#297)

### LINKED ISSUE #297: TestOpenClose is a flaky test
It's breaking in CI because we open up a lot of ports now and this test verifies that a just closed port is no longer in use but in reality it was just picked up by another process since it was free'd.

### LINKED ISSUE #298: Avoid Generating HTTP endpoint strings
`fmt.Sprintf(":d", portNumber)` is a bad pattern that breaks if IPv6 is used for localhost. Update `ListenerHolder` to return the address string instead of just port number.

jeremyje:
On machines that have IP v6 enabled localhost appears as `[::]:PORT` on IPv4 it's `:PORT`. When using certgen there's issues assuming IPv4 style when using server overrides to validating certificates.
Resolves #298
Resolves #297

### DIFF
--- a/internal/util/netlistener/listener_holder_test.go
+++ b/internal/util/netlistener/listener_holder_test.go
@@ -1,6 +1,8 @@
 package netlistener
 
 import (
+	"fmt"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"testing"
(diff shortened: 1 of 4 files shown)


internal/util/netlistener/listener_holder_test.go
```diff
@@ -10,6 +12,27 @@ const (
...
+	if err != nil {
+		t.Fatalf("NewFromPortNumber(0) had error, %s", err)
+	}
+
+	if !strings.HasSuffix(lh.AddrString(), fmt.Sprintf(":%d", lh.Number())) {
+		t.Errorf("%s does not have suffix ':%d'", lh.AddrString(), lh.Number())
+	}
+
+	port, err := lh.Obtain()
+	defer port.Close()
```
golangcibot: SA5001: should check returned error before deferring port.Close() (from `staticcheck`)

PR #300, opened 2019-04-25
Speed up CI

jeremyje:


cloudbuild-site.yaml
```diff
@@ -0,0 +1,79 @@
+# Copyright 2019 Google Inc. All Rights Reserved.
```
yfei1: Why do we have a new config file with its contents as a subset of the original cloudbuild file? Could you elaborate a bit in the commit description?

cloudbuild.yaml
```diff
@@ -96,14 +96,10 @@ steps:
   - name: 'go-vol'
     path: '/go'
   waitFor: ['Build: Protocol Buffers']
-- id: 'Build: Docker Images'
```
yfei1: Why are we skipping the build-images step?

Issue #231, opened 2019-04-17
Make CI Faster


jeremyje:
The continuous integration is taking a long time. It improve the build times there's a few things we can do.
Currently the build takes 15 minutes to complete.

* Add Kaniko cache. Currently: 2 min 36 sec
* Remove build steps that are unnecessary.
 * Binaries: Currently: 4 min 21 sec

Data points taken from, https://pantheon.corp.google.com/cloud-build/builds/01998d5b-ad3e-429b-8549-fc213529ca62?project=open-match-build

yfei1:
The current docker build steps run too slow (up to 6min 51 sec) since it attempts to send all context under the root directory (~50MB) for each build step. We should specify the context needed for each image build to avoid wasting time on transferring unused data.

yfei1:
Reopen this issue to track and benchmark CI time. Current CI time varies and takes around 15-19min to finish.

yfei1:
#701 Removes the Create Cluster step from cloudbuild. This could saves about 3-5min build time

yfei1:
Close this for now by a series of PRs.
Improved the CI build time from [19.5 min ](https://pantheon.corp.google.com/cloud-build/builds/8860a285-0761-4116-8f6b-3f933df0ce86?project=open-match-build) to [14.5 min](https://pantheon.corp.google.com/cloud-build/builds/3dbaf065-c8f4-4630-a4ae-b6cdffefc5d3?project=open-match-build).

PR #301, opened 2019-04-25
Use n1-highcpu-32 machine and cache base image for 2 days in CI (#231)

jeremyje:
Use a larger machine so we can build more artifacts in parallel. Increase the Kaniko cache for the base CI image since it barely changes and the tooling does not change day-to-day.


jeremyje:
> [Comments](https://github.com/GoogleCloudPlatform/open-match/blob/8dcdc460ece5300aa4ef0be60b39618c85cc94cf/cloudbuild.yaml#L202) in line 50 and line 202-207 are now resolved and should be deleted to increase file readability.

I've addressed these.

cloudbuild.yaml
```diff
@@ -47,10 +47,9 @@
 #    https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/go
 
 steps:
-# Blocked by https://github.com/GoogleContainerTools/kaniko/issues/477
 - id: 'Docker Image: open-match-build'
   name: gcr.io/kaniko-project/executor
-  args: ['--destination=gcr.io/$PROJECT_ID/open-match-build', '--cache=true', '--cache-ttl=6h', '--dockerfile=Dockerfile.ci', '.']
+  args: ['--destination=gcr.io/$PROJECT_ID/open-match-build', '--cache=true', '--cache-ttl=48h', '--dockerfile=Dockerfile.ci', '.']
```
sawagh: whats the motivation for this?

jeremyje: Added in the PR description. The cache retention was too short. The builds are still taking a while and caching (while the benefit is marginal) is short so there's more rebuilds than necessary.

Issue #306, opened 2019-04-26
Create directory for all new modules in internal/ that will eventually replace old ones.


jeremyje:
As part of the large change between 0.5.0. and 0.6.0 we'll be introducing a lot of new modules that will replace old ones. A few key ones are serving/, statestorage/, logging/ since they do not lend themselves to testability and have leaky abstractions.

As we build services on the 0.6.0 release we currently plan on writing the new modules along side the old ones so eng that wants to build a new replacement service can do so with the old code if the new modules are not available.

To help differentiate between the 2 we need a new directory in internal/ to make it clear that you're using a newer version of the service if it's in there. The current option is to use internal/future/. Ultimately this directory will get deleted and the contents will be moved to their permanent place (basically remove future/ from the path).

PR #307, opened 2019-04-26
Introduce internal/future/ directory with it's first file, fake_frontend.go. (#306)

jeremyje:


jeremyje:
> If it's for v0.6, name it as such? Given this isn't actually v2....

This was discussed in the meeting but I've changed it to internal/future/ now.

internal/future/testing/fake_frontend.go
```diff
@@ -0,0 +1,46 @@
+/*
```
golangcibot: package comment should be of the form "Package testing ..." (from `golint`)

internal/future/testing/fake_frontend.go
```diff
@@ -0,0 +1,46 @@
...
+Copyright 2019 Google LLC
```
sawagh: What does fake_frontend.go represent here? What do you anticipate goes under future/testing folder ? E2E tests - or - unit tests for each component (i would imaging UTs living within the component folder may be in a test subfolder)

jeremyje: It's used in the internal gRPC tests that are in the follow PRs. It's using the old FE def but should be updated to the new one once that's available.

PR #325, opened 2019-04-30
Auto-generate new protobufs and APIs.

jeremyje:


api/backend.proto
```diff
@@ -17,62 +17,76 @@ package api;
...
+// and make assignments for Tickets.
 service Backend {
-  // GetMatches triggers execution of the specfied MatchFunction for each of the specified MatchProfiles.
-  // Each MatchFunction execution returns a set of proposals which are then evaluated to generate results.
-  // GetMatches method streams these results back to the caller.
+  // GetMatches triggers execution of the specfied MatchFunction for each of the
+  // specified MatchProfiles. Each MatchFunction execution returns a set of
+  // proposals which are then evaluated to generate results. GetMatches method
+  // streams these results back to the caller.
+  // TODO: Should this be renamed to createProposal? It's not a "Get" if it's
```
sawagh: I can see why 'get' might not be a good idea (since matches get generated - not simply fetched)

This method does not cause any state change in Open Match (now that proposals, matches are not stored in state storage) - no Open Match stored resource is created here. Is 'create' still a write verb then? Or would a custom verb (say generatematches / makematches / ... be more appropriate?)

api/mmlogic.proto
```diff
@@ -17,22 +17,30 @@ package api;
...
+// as retreiving Tickets from state storage.
 service MmLogic {
-  // RetrievePool gets the list of Tickets that match every Filter in the specified Pool.
+  // RetrievePool gets the list of Tickets that match every Filter in the
+  // specified Pool.
+  // TODO: Consider renaming to "GetPool" to be consistent with HTTP REST CRUD
+  // conventions.
   rpc RetrievePool(RetrievePoolRequest) returns (stream RetrievePoolResponse) {
+    option (google.api.http) = {
+      post: "/v1/mmlogic/pools"
```
yfei1: I think it is a bit confusing to set up a POST endpoint for a Get operation, is it possible to use a GET endpoint instead?

api/backend.proto
jeremyje: I do not want to redesign the API in this change. Let's leave the TODO as is until reviewed. There were some things I had to workaround. In this case you cannot have a GET verb with a body so I have to use POST which cascades into requiring a verb name since a POST against a collection is an insert of that resource.

api/frontend.proto
```diff
@@ -48,28 +48,47 @@ message GetTicketUpdatesRequest {
...
   }
 
-  // GetTicketUpdates streams matchmaking results from Open Match for the provided Ticket id.
-  rpc GetTicketUpdates(GetTicketUpdatesRequest) returns (stream GetTicketUpdatesResponse) {
+  // GetTicketUpdates streams matchmaking results from Open Match for the
+  // provided Ticket id.
+  rpc GetTicketUpdates(GetTicketUpdatesRequest)
+      returns (stream GetTicketUpdatesResponse) {
+    option (google.api.http) = {
+      post: "/v1/frontend/tickets/assign"
```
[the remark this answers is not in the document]
jeremyje: This was a copy paste mistake.

PR #328, opened 2019-04-30
make test now does coverage

jeremyje:
`make test` does not produce coverage metrics. As we add more tests it'd be nice to know which packages have the most test gaps.


yfei1:
Just for curious, why do we have two go test commands right here? Maybe we should probably remove the test if we wanna enable/disable race detection accordingly?

PR #335, opened 2019-05-02
Add top level Swagger annotations for API.

jeremyje:
Adding top level swagger doc annotations for the Open Match API. These annotations help document the API so users that consume it via HTTPS get pointers to the documentation etc.


api/backend.swagger.json
```diff
@@ -312,13 +333,16 @@
...
...rshalAny(any, foo); err != nil {\n       ...\n     }\n\nThe pack methods provided by protobuf library will by default use\n'type.googleapis.com/full.type.name' as the type URL and the unpack\nmethods only use the fully qualified type name after the last '/'\nin the type URL, for example \"foo.bar.com/x/y.z\" will yield type\nname \"y.z\".\n\n\nJSON\n====\nThe JSON representation of an `Any` value uses the regular\nrepresentation of the deserialized, embedded message, with an\nadditional field `@type` which contains the type URL. Example:\n\n    package google.profile;\n    message Person {\n      string first_name = 1;\n      string last_name = 2;\n    }\n\n    {\n      \"@type\": \"type.googleapis.com/google.profile.Person\",\n      \"firstName\": \u003cstring\u003e,\n      \"lastName\": \u003cstring\u003e\n    }\n\nIf the embedded message type is well-known and has a custom JSON\nrepresentation, that representation will be embedded adding a field\n`value` which holds the custom JSON in addition to the `@type`\nfield. Example (for message [google.protobuf.Duration][]):\n\n    {\n      \"@type\": \"type.googleapis.com/google.protobuf.Duration\",\n      \"value\": \"1.212s\"\n    }"
```
yfei1: Does this commit includes changing the property bag from json to Any? We should keep the API description and message proto definition consistent.

api/frontend.proto
```diff
@@ -12,12 +12,49 @@
...
+	}
+	external_docs: {
+		url: "https://open-match.dev/site/docs/";
+		description: "Open Match Documentation";
+	}
+	schemes: HTTP;
+	schemes: HTTPS;
+	consumes: "application/json";
+	produces: "application/json";
+	responses: {
```
yfei1: Are we going to add the [security annotations](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/proto/examplepb/a_bit_of_everything.proto#L80) at here? I am not sure if users are still able to trigger api calls from swagger ui once the TLS support is enabled.

jeremyje: Yes but not right now. Will add a TODO.

PR #339, opened 2019-05-02
Move the future/fake_frontend to use the future/pb.

jeremyje:


internal/future/testing/fake_frontend.go
```diff
@@ -18,31 +18,39 @@ package testing
...
-// NewFakeFrontend creates a new FakeFrontend for tests.
-func NewFakeFrontend() *FakeFrontend {
-	return &FakeFrontend{}
+// CreateTicket will create a new ticket, assign a Ticket id to it and put the
+// Ticket in state storage. It will then look through the 'properties' field
+// for the attributes defined as indices the matchmakaking config. If the
+// attributes exist and are valid integers, they will be indexed. Creating a
+// ticket adds the Ticket to the pool of Tickets considered for matchmaking.
+func (s *FakeFrontend) CreateTicket(ctx context.Context, req *pb.CreateTicketRequest) (*pb.CreateTicketResponse, error) {
+	return &pb.CreateTicketResponse{}, nil
```
sawagh: return status.Error(codes.Unimplemented, "not implemented")

jeremyje: This is intentional. We use this in tests.

PR #343, opened 2019-05-03
Add go-bindata generation for Open Match new binaries

jeremyje:
The new target here is `make internal/future/assets/bindata.go`.
This file will hold all the serveable assets for open match servers.


yfei1:
Is it possible to put all swagger files under directories with different prefixes and rename them to swagger.json (e.g. backend/swagger.json, frontend/swagger.json...) ? This gives us the ability to render the swagger files with the same url path. This also gives us the ability to control what to render by specifying the prefix when using assetfs

internal/future/assets/bindata.go
```diff
@@ -0,0 +1,367 @@
+// Code generated by go-bindata. DO NOT EDIT.
+// sources:
+// backend.swagger.json (16.416kB)
+// bindata.go (67.281kB)
```
yfei1: It seems like we are having a nested bindata file here.

I suggest organizing the bindata with the following structure:

```
.
├── frontend
|   └── swagger.json
├── backend
|   └── swagger.json
├── matchfunction
|   └── swagger.json
├── mmlogic
|   └── swagger.json
```

This gives us the ability to serve all the swagger files under the same file path (e.g localhost:51504/swagger/swagger.json)

PR #352, opened 2019-05-07
Release v0.5.1

yfei1:
Bump open match version to 0.5.1 based on 0.5.0 release


jeremyje:
There shouldn't be a release-0.5.1 branch. You just work off of the release-0.5 branch.
Also there's no 0.5.0 release binaries yet so we need those to.

PR #365, opened 2019-05-07
Refactor protos to match CRUD operations

yfei1:


api/backend.proto
```diff
@@ -103,15 +103,13 @@ message AssignTicketsResponse {}
...
-  // proposals which are then evaluated to generate results. GetMatches method
+  // proposals which are then evaluated to generate results. CreateProposal method
   // streams these results back to the caller.
-  // TODO: Should this be renamed to createProposal? It's not a "Get" if it's
-  // executing a MatchFunction.
-  rpc GetMatches(GetMatchesRequest) returns (stream GetMatchesResponse) {
+  rpc CreateProposal(CreateProposalRequest) returns (stream CreateProposalResponse) {
     option (google.api.http) = {
-      post: "/v1/backend/matches/createProposal"
+      post: "/v1/backend/proposal/create"
```
jeremyje: /proposals/create

Proposal is the noun so it's supposed to be plural because it's represented as a collection.

api/mmlogic.proto
```diff
@@ -54,24 +54,23 @@ option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
...
 // The MMLogic API provides utility functions for common MMF functionality such
 // as retreiving Tickets from state storage.
 service MmLogic {
-  // RetrievePool gets the list of Tickets that match every Filter in the
+  // GetPoolTickets gets the list of Tickets that match every Filter in the
   // specified Pool.
-  // TODO: Consider renaming to "GetPool" to be consistent with HTTP REST CRUD
-  // conventions. Right now there's a GET and a POST for this verb.
-  rpc RetrievePool(RetrievePoolRequest) returns (stream RetrievePoolResponse) {
+  // Right now there's a GET and a POST for this verb.
```
jeremyje: Can we address this? This looks like a GET not a POST.

api/mmlogic.proto
```diff
@@ -54,24 +54,23 @@ option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
...
+message GetPoolTicketsResponse {
   // The Tickets that meet the Filter criteria requested by the Pool.
   repeated Ticket ticket = 1;
 }
 
 // The MMLogic API provides utility functions for common MMF functionality such
 // as retreiving Tickets from state storage.
 service MmLogic {
-  // RetrievePool gets the list of Tickets that match every Filter in the
+  // GetPoolTickets gets the list of Tickets that match every Filter in the
```
jeremyje: This is not a meaningful API name. I wonder if this query is too generic or so.

yfei1: how about GetFilteredTickets?

jeremyje: Can you run this by @sawagh? The names should be consistent and map to HTTP conventions.

api/mmlogic.proto
yfei1: The request contains a pool with repeated filters and it is not trivial and intuitive to include everything via a get url, so I leave the post binding for convenience purpose.

yfei1:
This update defines four data [collections](https://cloud.google.com/apis/design/resources#resources) for open-match: `Tickets`, `Assignments`, `Profile`, and `Proposals`, with `Assignments` as a sub-collection of `Tickets`.

The update also try to resolve #380 by replacing some of the custom verbs to standard ones and introduce custom verbs that follows the standard [HTTP mapping guidelines](https://cloud.google.com/apis/design/custom_methods#http_mapping)

Reference: [API Design Doc](https://cloud.google.com/apis/design/)

api/backend.proto
```diff
@@ -77,15 +77,15 @@ message MatchFunctionConfig {
...
   MatchFunctionConfig function = 1;
 
   // MatchProfiles for which this MatchFunction should be executed.
   repeated MatchProfile profile = 2;
 }
 
-message GetMatchesResponse {
+message CreateProposalResponse {
   // Result Match for the requested MatchProfile.
   Match match = 1;
```
jeremyje: Should we rename Match as Proposal?
Also @Laremere had an interesting proposal to split this definition out so there's an immutable and append-only object.

It's very strange to have a create verb that returns a resource of a type that's different.

api/backend.proto
```diff
@@ -120,7 +118,7 @@ service Backend {
   // IDs passed.
   rpc AssignTickets(AssignTicketsRequest) returns (AssignTicketsResponse) {
     option (google.api.http) = {
-      post: "/v1/backend/tickets/assign"
+      post: "/v1/backend/tickets:assign"
```
jeremyje: For my understanding, what is this?
I don't see it used in https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/proto/examplepb/a_bit_of_everything.proto.

PR #369, opened 2019-05-08
Implement the MMLogic Service

yfei1:


internal/future/app/mmlogic/mmlogic_service_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"open-match.dev/open-match/internal/config"
+	"open-match.dev/open-match/internal/future/pb"
+	"open-match.dev/open-match/internal/future/serving"
+	serveTesting "open-match.dev/open-match/internal/future/serving/testing"
+	"open-match.dev/open-match/internal/future/storage"
+	storageTesting "open-match.dev/open-match/internal/future/storage/testing"
+)
+
+var (
+	cfg     config.View
```
golangcibot: `cfg` is a global variable (from `gochecknoglobals`)

internal/future/app/mmlogic/mmlogic_service_test.go
```diff
@@ -0,0 +1,113 @@
...
+	mmlogic *mmlogicService
```
golangcibot: `mmlogic` is a global variable (from `gochecknoglobals`)

internal/future/app/mmlogic/mmlogic_service_test.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+func TestRetrievePool(t *testing.T) {
+	assert := assert.New(t)
+	mmlogicClient, closer := initialize(t)
+	defer closer()
+	cfg.Set("hi", 1)
```
golangcibot: cfg.Set undefined (type config.View has no field or method Set) (from `typecheck`)

internal/future/storage/testing/public.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"time"
+
+	"github.com/alicebob/miniredis"
+	"github.com/spf13/viper"
+	"open-match.dev/open-match/internal/future/storage"
+)
+
+func NewRedisForTesting() (storage.Service, func() error) {
```
golangcibot: exported function `NewRedisForTesting` should have comment or be unexported (from `golint`)

internal/future/app/mmlogic/mmlogic_service_test.go
```diff
@@ -0,0 +1,106 @@
...
+	assert.Nil(resp)
+}
+
+func initialize(t *testing.T) (pb.MmLogicClient, func()) {
+	assert := assert.New(t)
+	cfg := viper.New()
+	cfg.Set("storage.page.size", 1000)
+
+	rs, rsCloser := storageTesting.NewRedisForTesting()
+	mmlogic = &mmlogicService{cfg: cfg, storageService: rs}
```
golangcibot: undeclared name: `mmlogic` (from `typecheck`)

internal/future/app/mmlogic/mmlogic_service.go
```diff
@@ -15,31 +15,135 @@
...
 // specified Pool.
 // TODO: Consider renaming to "GetPool" to be consistent with HTTP REST CRUD
 // conventions. Right now there's a GET and a POST for this verb.
-func (s *mmlogicService) RetrievePool(req *pb.RetrievePoolRequest, stream pb.MmLogic_RetrievePoolServer) error {
-	ctx := stream.Context()
+func (s *mmlogicService) RetrievePool(req *pb.RetrievePoolRequest, responseServer pb.MmLogic_RetrievePoolServer) error {
+	poolName := req.Pool.Name
+	poolFilters := req.Pool.Filter
+	redisConn := storage.GetDeprecatedRedisPool(s.storageService).Get()
+	defer redisConn.Close()
```
golangcibot: Error return value of `redisConn.Close` is not checked (from `errcheck`)

internal/future/app/mmlogic/mmlogic_service.go
```diff
@@ -15,31 +15,135 @@
...
-
 		default:
-			err := stream.Send(&pb.RetrievePoolResponse{})
-			if err != nil {
+			// Initialize an empty page to send
+			page := []*pb.Ticket{}
+
+			for pOffset := 0; pOffset < pSize && setIdx < len(idSet); pOffset++ {
+				// A map from attribute names to attribute values per ticket
+				propertyPerTicket := make(map[string]int64, 0)
```
golangcibot: S1019: should use make(map[string]int64) instead (from `gosimple`)

internal/future/serving/server.go
```diff
@@ -49,8 +49,8 @@ type Params struct {
 	ServeMux             *http.ServeMux
 	handlersForGrpc      []GrpcHandler
 	handlersForGrpcProxy []GrpcProxyHandler
-	grpcListener         *netlistener.ListenerHolder
-	grpcProxyListener    *netlistener.ListenerHolder
+	GrpcListener         *netlistener.ListenerHolder
```
jeremyje: Please don't do this. But someone needs to solve the client story.
This is dangerous because the ListenerHolder is a single use object.

internal/future/app/mmlogic/mmlogic_service.go
```diff
@@ -15,31 +15,139 @@
...
+	poolName := req.Pool.Name
+	poolFilters := req.Pool.Filter
+	redisConn := storage.GetDeprecatedRedisPool(s.storageService).Get()
+	defer func() {
+		if err := redisConn.Close(); err != nil {
+			logger.WithError(err).Error("Failed to close Redis connection")
+		}
+	}()
+
+	logger.WithFields(logrus.Fields{
```
jeremyje: Unnecessary. We can add an interceptor to grpc to provide these generic log statements.

internal/future/app/mmlogic/mmlogic_service.go
```diff
@@ -15,31 +15,139 @@
...
+	"open-match.dev/open-match/internal/config"
 	"open-match.dev/open-match/internal/future/pb"
+	"open-match.dev/open-match/internal/future/storage"
+	"open-match.dev/open-match/internal/set"
+
+	"fmt"
+)
+
+var (
+	logger = logrus.WithFields(logrus.Fields{
```
jeremyje: There's a convention to use the filename as part of the logger name. That way each file has it's own logger. This might change in the future.

internal/future/app/mmlogic/mmlogic_service.go
```diff
@@ -15,31 +15,139 @@
...
 // RetrievePool gets the list of Tickets that match every Filter in the
 // specified Pool.
 // TODO: Consider renaming to "GetPool" to be consistent with HTTP REST CRUD
 // conventions. Right now there's a GET and a POST for this verb.
-func (s *mmlogicService) RetrievePool(req *pb.RetrievePoolRequest, stream pb.MmLogic_RetrievePoolServer) error {
-	ctx := stream.Context()
+func (s *mmlogicService) RetrievePool(req *pb.RetrievePoolRequest, responseServer pb.MmLogic_RetrievePoolServer) error {
+	poolName := req.Pool.Name
+	poolFilters := req.Pool.Filter
+	redisConn := storage.GetDeprecatedRedisPool(s.storageService).Get()
```
jeremyje: I'd say we should stuff the redis read data into the client now. Let's not use redis directly here. @sawagh Submitted the API shell for `statestore`.

internal/future/app/mmlogic/mmlogic_service.go
```diff
@@ -15,31 +15,139 @@
...
+	for i, filter := range poolFilters {
+		startTime := time.Now()
+		// Time Complexity O(logN + M), where N is the number of elements in the attribute set
+		// and M is the number of entries being returned.
+		idToValue, err := redis.Int64Map(redisConn.Do("ZRANGEBYSCORE", filter.Attribute, filter.Min, filter.Max, "WITHSCORES"))
+
+		if err != nil {
+			logger.WithFields(logrus.Fields{
+				"Command": fmt.Sprintf("ZRANGEBYSCORE %s %f %f WITHSCORES", filter.Attribute, filter.Min, filter.Max),
+			}).WithError(err)
```
jeremyje: If this query fails why are we continuing?
Also we should respond with a reasonable gRPC error code.

internal/future/testing/pb.go
```diff
@@ -0,0 +1,14 @@
+package testing
+
+import "open-match.dev/open-match/internal/future/pb"
+
+func NewPbFilter(attributeName string, min, max float64) *pb.Filter {
```
golangcibot: exported function `NewPbFilter` should have comment or be unexported (from `golint`)

internal/future/testing/pb.go
```diff
@@ -0,0 +1,14 @@
...
+	return &pb.Filter{Attribute: attributeName, Min: min, Max: max}
+}
+
+func NewPbPool(poolName string, filters ...*pb.Filter) *pb.Pool {
```
golangcibot: exported function `NewPbPool` should have comment or be unexported (from `golint`)

internal/future/testing/pb.go
```diff
@@ -0,0 +1,16 @@
+package testing
+
+import "open-match.dev/open-match/internal/future/pb"
+
+// A wrapper to create protobuf filter objects
```
golangcibot: comment on exported function `NewPbFilter` should be of the form `NewPbFilter ...` (from `golint`)

internal/future/testing/pb.go
```diff
@@ -0,0 +1,16 @@
...
+func NewPbFilter(attributeName string, min, max float64) *pb.Filter {
+	return &pb.Filter{Attribute: attributeName, Min: min, Max: max}
+}
+
+// A wrapper to create protobuf pool objects
```
golangcibot: comment on exported function `NewPbPool` should be of the form `NewPbPool ...` (from `golint`)

internal/future/app/mmlogic/mmlogic_service.go
```diff
@@ -43,21 +55,51 @@ func newMmlogic(cfg config.View) (*mmlogicService, error) {
...
+	// Send requests to the storage service
+	idsToProperties, err := s.store.FilterTickets(poolFilters)
+	if err != nil {
+		logger.WithError(err).Error("Failed to retrieve result from storage service.")
+		return status.Error(codes.Internal, err.Error())
+	}
 
-	for {
+	ctx := responseServer.Context()
+	idx := 0
```
sawagh: can you use better variable naming? idx seems confusing,

internal/future/app/mmlogic/mmlogic_service.go
```diff
@@ -43,21 +55,51 @@ func newMmlogic(cfg config.View) (*mmlogicService, error) {
...
 		select {
 		case <-ctx.Done():
 			return nil
-
 		default:
-			err := stream.Send(&pb.QueryTicketsResponse{})
+			propertyByte, err := json.Marshal(property)
 			if err != nil {
-				return err
+				logger.WithError(err).Error("Failed to convert go struct to JSON")
```
sawagh: nit: failed to convert property map to JSON

internal/future/app/mmlogic/mmlogic_service.go
```diff
@@ -43,21 +55,51 @@ func newMmlogic(cfg config.View) (*mmlogicService, error) {
...
+				return status.Errorf(codes.Internal, err.Error())
+			}
+			propertyJSON := string(propertyByte)
+			page = append(page, &pb.Ticket{Id: id, Properties: propertyJSON})
```
sawagh: nit: simply use string(propertyByte) here? no need for extra variable.

internal/future/app/mmlogic/mmlogic_service.go
```diff
@@ -43,21 +55,51 @@ func newMmlogic(cfg config.View) (*mmlogicService, error) {
...
+
+			idx++
+
+			endPage := idx%pSize == 0 || idx == len(idsToProperties)-1
+			if endPage {
+				// Reaches page limit; Send a stream response then reset the page
+				err := responseServer.Send(&pb.QueryTicketsResponse{Ticket: page})
+				if err != nil {
+					logger.WithError(err).Error("Failed to send Redis response to grpc server")
+					return status.Errorf(codes.Unknown, err.Error())
```
sawagh: Internal ? Or Unknown?

internal/future/app/mmlogic/mmlogic_service.go
```diff
@@ -43,21 +55,51 @@ func newMmlogic(cfg config.View) (*mmlogicService, error) {
...
+	}
 
-	for {
+	ctx := responseServer.Context()
+	idx := 0
+	// The number of tickets in a paging response
+	pSize := s.cfg.GetInt("storage.page.size")
+	page := []*pb.Ticket{}
+
+	for id, property := range idsToProperties {
```
sawagh: There is some paging logic in this that is interesting to test. May be break this helper into a separate method - so you can give it a populated map and validate it handles paging properly?

internal/future/statestore/public.go
```diff
@@ -48,7 +48,7 @@ type Service interface {
 	DeindexTicket(ctx context.Context, id string, indices []string) error
 
 	// FilterTickets returns the Ticket ids for the Tickets meeting the specified filtering criteria.
-	FilterTickets(ctx context.Context, filters []pb.Filter) ([]string, error)
```
sawagh: Once my public_test changes are checked in, please add a test to actually create tickets and filter to see if they return expected results.

internal/future/statestore/redis.go
```diff
@@ -119,7 +120,70 @@ func (rb *redisBackend) DeindexTicket(context.Context, string, []string) error {
...
+	attributeToTickets := make(map[string]map[string]int64)
+	// A set of playerIds that satisfies all filters
+	idSet := make([]string, 0)
+
+	// For each filter, do a range query to Redis on Filter.Attribute
+	for i, filter := range filters {
+		startTime := time.Now()
+		// Time Complexity O(logN + M), where N is the number of elements in the attribute set
+		// and M is the number of entries being returned.
+		idToValue, err := redis.Int64Map(redisConn.Do("ZRANGEBYSCORE", filter.Attribute, filter.Min, filter.Max, "WITHSCORES"))
```
sawagh: So if the responses exceed max here - we silently only process up to max right? how is that surfaced to the user? Should we log that we only processed so many? is there a way to get the total count so we can tell how many were missed? I ll create an issue to track the work to figure how this needs to be surfaces to the caller.

internal/future/statestore/public.go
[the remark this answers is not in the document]
yfei1: Done.

internal/future/app/mmlogic/mmlogic_service.go
yfei1: I will add the paging logic tests to `redis_test.go` file later. I didn't do this for now because I want to confirm the functions and signatures are well-established to avoid refactoring too many files

internal/future/app/mmlogic/mmlogic_service.go
yfei1: I think it might be worth to create self-explained variable names to improve code readability, but you are right!

internal/future/app/mmlogic/mmlogic_service.go
yfei1: I'll probably do Abort. It make sense to return a grpc abort error if something goes wrong when sending/receiving messages.

internal/future/statestore/redis.go
yfei1: Redis does not come with a default maximum number of entries being returned for each command. As long as we are executing the command a [`LIMIT`](https://redis.io/commands/zrangebyscore) option, Redis will return all results to the client.

internal/future/statestore/redis.go
```diff
@@ -119,7 +120,70 @@ func (rb *redisBackend) DeindexTicket(context.Context, string, []string) error {
...
+
+		if err != nil {
+			redisLogger.WithFields(logrus.Fields{
+				"Command": fmt.Sprintf("ZRANGEBYSCORE %s %f %f WITHSCORES", filter.Attribute, filter.Min, filter.Max),
+			}).WithError(err)
+			return nil, err
+		}
+
+		redisLogger.WithFields(logrus.Fields{
```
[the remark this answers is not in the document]
yfei1: Good point, does it make sens if I turn it into a trace log?

internal/future/app/mmlogic/mmlogic_service.go
sawagh: Ok. Create an issue to track the paging behavior / validation of the same. I think we need some thought to how excess results will be handled through the stack (in mmlogic / at storage layer)

internal/future/app/mmlogic/mmlogic_service.go
sawagh: From the description here, looks like Aborted is if the client must retry - https://github.com/grpc/grpc-go/blob/master/codes/codes.go - Do you think retrying will help here? Internal is when something unexpected happens - indicating some issue with the system itself. @jeremyje might have guidance here.

internal/future/statestore/public.go
```diff
@@ -59,10 +59,10 @@ func New(cfg config.View) (Service, error) {
 	return newRedis(cfg)
 }
 
-// GetDeprecatedRedisPool returns a Redigo Redis pool.
+// getDeprecatedRedisPool returns a Redigo Redis pool.
```
sawagh: Why are we making this lower case? I thought this was meant to be publicly exposed to legacy clients?

internal/future/statestore/redis.go
```diff
@@ -119,7 +120,72 @@ func (rb *redisBackend) DeindexTicket(context.Context, string, []string) error {
 	return status.Error(codes.Unimplemented, "not implemented")
 }
 
-// FilterTickets returns the Ticket ids for the Tickets meeting the specified filtering criteria.
-func (rb *redisBackend) FilterTickets(context.Context, []pb.Filter) ([]string, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+// FilterTickets returns the Ticket ids and required attribute key-value pairs for the Tickets meeting the specified filtering criteria.
+// map[ticket.Id]map[attributeName][attributeValue]
```
sawagh: Can we also add a commented example? That would make this super clear - something like:
{
    "testplayer1":  {"ranking" : 56, "loyalty_level": 4},
    "testplayer2":  {"ranking" : 50, "loyalty_level": 3}
}

internal/future/statestore/redis.go
sawagh: So are we using the LIMIT then? theres ZCOUNT too - the question is if do we really want to support this unbounded.

internal/future/statestore/redis.go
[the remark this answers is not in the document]
sawagh: Remove it for now? we can follow all the tracing changes in one go later?

internal/future/statestore/public.go
yfei1: Are you referring to examples/backendclient? That one does not talk to redis directly. The latest change that you made last week abstract the redis existence so we only need to expose this function package-wise

internal/future/statestore/redis.go
yfei1: Added a TODO to fix later.

internal/future/statestore/redis.go
```diff
@@ -119,7 +120,69 @@ func (rb *redisBackend) DeindexTicket(context.Context, string, []string) error {
...
+	}()
+
+	// A map[attribute]map[playerID]value
+	attributeToTickets := make(map[string]map[string]int64)
+	// A set of playerIds that satisfies all filters
+	idSet := make([]string, 0)
+
+	// For each filter, do a range query to Redis on Filter.Attribute
+	for i, filter := range filters {
+		startTime := time.Now()
```
golangcibot: ineffectual assignment to `startTime` (from `ineffassign`)

internal/future/statestore/public.go
sawagh: Is this used in code at all anymore? This was used - and hence it was in capital.. but i checked and does not seem so.. remove this method completely if not used,

PR #377, opened 2019-05-08
Initialize state storage in Frontend, Backend and MMLogic services

sawagh:


internal/future/serving/server.go
```diff
@@ -60,6 +60,8 @@ type Params struct {
 	publicCertificateFileData []byte
 	// Private key in PEM format.
 	privateKeyFileData []byte
+	// Store is an interface to the state storage used to access Open Match state.
+	Store statestore.Service
```
jeremyje: This should not be a part of the server.

sawagh: Just initialize in each service then?

sawagh:
Looks like BindService will need to be changed throughout! Will make the modification.

internal/future/app/backend/backend_service.go
```diff
@@ -17,12 +17,32 @@ package backend
...
 
+	"open-match.dev/open-match/internal/config"
 	"open-match.dev/open-match/internal/future/pb"
+	"open-match.dev/open-match/internal/future/statestore"
 )
 
 // The service implementing the Backent API that is called to generate matches
 // and make assignments for Tickets.
-type backendService struct {
+type BackendService struct {
```
golangcibot: type name will be used as backend.BackendService by other packages, and that stutters; consider calling this Service (from `golint`)

internal/future/app/frontend/frontend_service.go
```diff
@@ -17,38 +17,58 @@ package frontend
...
 
+	"open-match.dev/open-match/internal/config"
 	"open-match.dev/open-match/internal/future/pb"
+	"open-match.dev/open-match/internal/future/statestore"
 )
 
 // frontendAPI implements pb.ApiServer, the server generated by compiling
 // the protobuf, by fulfilling the pb.APIClient interface.
-type frontendService struct {
+type FrontendService struct {
```
golangcibot: type name will be used as frontend.FrontendService by other packages, and that stutters; consider calling this Service (from `golint`)

internal/future/app/mmlogic/mmlogic_service.go
```diff
@@ -15,19 +15,39 @@
...
 import (
+	"open-match.dev/open-match/internal/config"
 	"open-match.dev/open-match/internal/future/pb"
+	"open-match.dev/open-match/internal/future/statestore"
 )
 
 // The MMLogic API provides utility functions for common MMF functionality such
 // as retreiving Tickets from state storage.
-type mmlogicService struct {
+type MmlogicService struct {
```
golangcibot: type name will be used as mmlogic.MmlogicService by other packages, and that stutters; consider calling this Service (from `golint`)

internal/future/app/minimatch/minimatch.go
```diff
@@ -45,13 +45,29 @@ func RunApplication() {
 		}).Fatalf("cannot construct server.")
 	}
 
-	BindService(p)
+	BindService(p, cfg)
```
golangcibot: Error return value of `BindService` is not checked (from `errcheck`)

sawagh:
The tricky part here is - The Frontend, Backend, MMLogic services now need a config based initialization (to initialize the state storage). 

Option1:  Bind takes config - and to initialize services within. That works great except for the TestServerBinding which does not have a config - and hence fails the Bind

Option2: Bind takes an interface for a service - the initialization happens outside. Thus the actual logic, minimatch can use the right initialization while TestServerBinding can use an un-initialized service. This is what the current CL does - the drawback is - given that now the service is taken by bind, we need to expose the service outside the package. 

Option3: Add explicit initialization to each service - which the core service binaries, minimatch will call after binding. However, this is susceptible to errors (forgeting to initialize in a real service)

The current implementation of TestServerBinding is the one that causes issues - because it assumes the service getting initialized in the internals - and does not have a config to actually initialize it. I ve not considered options to tweak this service itself.

@jeremyje  - Please let me know what you think.

PR #384, opened 2019-05-10
Redis implementation for State Storage methods for Tickets

sawagh:


internal/future/statestore/public_test.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2019 Google LLC
```
jeremyje: Rename to redis_test.go? I don't see any public.go stuff in here.

internal/future/statestore/redis.go
```diff
@@ -94,29 +96,208 @@ func redisCloser(closer func() error) {
...
-func (rb *redisBackend) CreateTicket(context.Context, pb.Ticket, int) error {
-	return status.Error(codes.Unimplemented, "not implemented")
+// CreateTicket creates a new Ticket in the state storage. If the id already exists, it will be overwritten.
+func (rb *redisBackend) CreateTicket(ctx context.Context, ticket *pb.Ticket, ttl int) error {
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		redisLogger.WithFields(logrus.Fields{
+			"error": err.Error(),
+		}).Error("failed to connect to redis")
+		return err
```
jeremyje: Prefer returning gRPC error codes here because at this point we know if the error is retryable or not.

internal/future/statestore/redis.go
```diff
@@ -94,29 +96,208 @@ func redisCloser(closer func() error) {
...
+
+	return ticket, nil
 }
 
-// DeleteTicket removes the Ticket with the specified id from state storage. This method succeeds if the Ticket does not exist.
-func (rb *redisBackend) DeleteTicket(context.Context, string) error {
-	return status.Error(codes.Unimplemented, "not implemented")
+// DeleteTicket removes the Ticket with the specified id from state storage.
+func (rb *redisBackend) DeleteTicket(ctx context.Context, id string) error {
+	redisConn, err := rb.redisPool.GetContext(ctx)
```
jeremyje: The act of getting a redis connection and logging an error appears many times in the code already. Make that a method.

internal/future/statestore/redis.go
```diff
@@ -94,29 +96,208 @@ func redisCloser(closer func() error) {
...
+	defer redisConn.Close()
+
+	redisConn.Send("MULTI")
+	// Loop through all attributes we want to index.
+	for _, attribute := range indices {
+		// TODO: Currently, Open Match supports specifying attributes as JSON properties on Tickets.
+		// Alternatives for populating this information in proto fields are being considered.
+		// Also, we need to add Ticket creation time to either the ticket or index.
+		// Meta characters bein specified in JSON property keys is currently not supported.
+		v := gjson.Get(ticket.Properties, attribute)
```
jeremyje: Just use the `import "encoding/json"` package for this.

internal/future/statestore/redis.go
```diff
@@ -94,29 +96,208 @@ func redisCloser(closer func() error) {
...
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		redisLogger.WithFields(logrus.Fields{
+			"error": err.Error(),
+		}).Error("failed to connect to redis")
+		return err
+	}
+	defer redisConn.Close()
+
+	redisConn.Send("MULTI")
```
jeremyje: Handle the error?

internal/future/statestore/redis.go
```diff
@@ -94,29 +96,208 @@ func redisCloser(closer func() error) {
...
+	key := ticket.Id
```
jeremyje: `ticket.Id` is more meaningful than `key`. I'd just use that straight.

internal/future/statestore/redis.go
```diff
@@ -94,29 +96,208 @@ func redisCloser(closer func() error) {
...
+	if err != nil {
+		redisLogger.WithFields(logrus.Fields{
+			"key":   key,
+			"cmd":   "SET",
+			"error": err.Error(),
+		}).Error("failed to set the value for ticket in state storage")
+		return err
+	}
+
+	if ttl > 0 {
```
jeremyje: It's not clear why `ttl` is a separate parameter in this method. Also should that just be time.Duration instead of an int? It's not clear what the units are.

internal/future/statestore/redis.go
```diff
@@ -94,29 +96,208 @@ func redisCloser(closer func() error) {
...
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		redisLogger.WithFields(logrus.Fields{
+			"error": err.Error(),
+		}).Error("failed to connect to redis")
+		return err
+	}
+	defer redisConn.Close()
+
+	redisConn.Send("MULTI")
```
jeremyje: Error check?

internal/future/statestore/redis.go
```diff
@@ -94,29 +96,208 @@ func redisCloser(closer func() error) {
...
+		// TODO: Currently, Open Match supports specifying attributes as JSON properties on Tickets.
+		// Alternatives for populating this information in proto fields are being considered.
+		// Also, we need to add Ticket creation time to either the ticket or index.
+		// Meta characters bein specified in JSON property keys is currently not supported.
+		v := gjson.Get(ticket.Properties, attribute)
+
+		// If this attribute wasn't provided in the JSON, continue to the next attribute to index.
+		if !v.Exists() {
+			redisLogger.WithFields(logrus.Fields{
+				"attribute": attribute}).Warningf("Couldn't find index in Ticket Properties")
```
jeremyje: Using `Warningf` without vaargs. Just use `Warning`

internal/future/statestore/redis.go
```diff
@@ -94,29 +96,208 @@ func redisCloser(closer func() error) {
...
+			continue
+		}
+
+		// Value exists. Check if it is a supported value for indexed properties.
+		if v.Int() < -9223372036854775808 || v.Int() > 9223372036854775807 {
```
jeremyje: `math.MaxInt64` and `math.MinInt64`.

internal/future/statestore/redis.go
```diff
@@ -94,29 +96,208 @@ func redisCloser(closer func() error) {
...
+			redisLogger.WithFields(logrus.Fields{
+				"attribute": attribute}).Warningf("Invalid value for attribute, skip indexing.")
+			continue
+		}
+
+		// Index the attribute by value.
+		value := v.Int()
+		redisConn.Send("ZADD", attribute, value, ticket.Id)
```
jeremyje: Error handling?

internal/future/statestore/redis.go
```diff
@@ -94,29 +96,208 @@ func redisCloser(closer func() error) {
...
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		redisLogger.WithFields(logrus.Fields{
+			"error": err.Error(),
+		}).Error("failed to connect to redis")
+		return err
+	}
+	defer redisConn.Close()
+
+	redisConn.Send("MULTI")
```
jeremyje: Error handling.

internal/future/statestore/redis.go
```diff
@@ -94,29 +96,208 @@ func redisCloser(closer func() error) {
...
+		return err
+	}
+
+	return nil
 }
 
 // DeindexTicket removes the indexing for the specified Ticket. Only the indexes are removed but the Ticket continues to exist.
-func (rb *redisBackend) DeindexTicket(context.Context, string, []string) error {
-	return status.Error(codes.Unimplemented, "not implemented")
+func (rb *redisBackend) DeindexTicket(ctx context.Context, id string, indices []string) error {
```
yfei1: Do we need to check the existence of the passed-in ticket?

internal/future/statestore/redis.go
```diff
@@ -94,29 +96,208 @@ func redisCloser(closer func() error) {
...
+	}
+
+	return ticket, nil
 }
 
-// DeleteTicket removes the Ticket with the specified id from state storage. This method succeeds if the Ticket does not exist.
-func (rb *redisBackend) DeleteTicket(context.Context, string) error {
-	return status.Error(codes.Unimplemented, "not implemented")
+// DeleteTicket removes the Ticket with the specified id from state storage.
+func (rb *redisBackend) DeleteTicket(ctx context.Context, id string) error {
```
yfei1: What would be the behavior if receiving a request to delete a non-existing ticket?

internal/future/statestore/public_test.go
sawagh: This only tests the public interface methods - with passing in the cfg needed to create whatever is the backend of choice in createBackend. Except for createBackend, I expect the rest of the tests in this method to stay identical even if redis is swapped for some other backend.

redis_test IMO will have a check such as - CreateTicket validating that the backed redis db actually got the entry created (not via GetTicket - but via querying the db)

internal/future/statestore/redis.go
sawagh: Done.

internal/future/statestore/redis.go
sawagh: Done.

internal/future/statestore/redis.go
sawagh: Given that we don't know the structure of the Properties, gjson just fetches the value for the given attribute. Since we are considering moving away from JSON, i ll leave it as such. If we decide to stick to JSON, then we should at-least determine the structure and use encoding/json to unmarshal.

internal/future/statestore/redis.go
sawagh: Deindexing will be successful - there may possibly be indexes that were leaked for a ticket and ticket actually deleted. I think its ok to delete (so long as non existing are treated as success) than add an additional existence check.

internal/future/statestore/redis.go
sawagh: Same - delete succeeds if ticket does not exist. I ve added tests to actually confirm the same.

internal/future/statestore/public_test.go
sawagh: This only tests the public interface methods - with passing in the cfg needed to create whatever is the backend of choice in createBackend. Except for createBackend, I expect the rest of the tests in this method to stay identical even if redis is swapped for some other backend.

redis_test IMO will have a check such as - CreateTicket validating that the backed redis db actually got the entry created (not via GetTicket - but via querying the db)

internal/future/statestore/redis.go
```diff
@@ -17,17 +17,21 @@ package statestore
...
 	"github.com/sirupsen/logrus"
+	"github.com/tidwall/gjson"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"open-match.dev/open-match/internal/config"
 	"open-match.dev/open-match/internal/future/pb"
 )
 
 var (
+	redisTTL    = -1
```
jeremyje: This global variable, it's not necessary.

internal/future/statestore/redis.go
```diff
@@ -74,12 +78,17 @@ func newRedis(cfg config.View) (Service, error) {
...
-			"query": "SELECT 0"}).Error("state storage connection error")
+			"cmd":   "SELECT 0",
+			"error": err.Error()}).Error("state storage connection error")
 		return nil, fmt.Errorf("cannot connect to Redis at %s, %s", maskedURL, err)
 	}
 
 	redisLogger.Info("Connected to Redis")
+
+	if cfg.IsSet("redis.expiration") {
+		redisTTL = cfg.GetInt("redis.expiration")
```
jeremyje: It's easier to just save the cfg instance on the service and read the TTL per request. That way if there's a config push the new TTL takes.

internal/future/statestore/redis.go
```diff
@@ -94,29 +103,250 @@ func redisCloser(closer func() error) {
...
-// CreateTicket creates a new Ticket in the state storage. This method fails if the Ticket already exists.
-func (rb *redisBackend) CreateTicket(context.Context, pb.Ticket, int) error {
-	return status.Error(codes.Unimplemented, "not implemented")
+func (rb *redisBackend) connect(ctx context.Context) (redis.Conn, error) {
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		redisLogger.WithFields(logrus.Fields{
+			"error": err.Error(),
+		}).Error("failed to connect to redis")
+		return nil, status.Errorf(codes.Unavailable, "%v", err)
```
jeremyje: status.Error(codes.Unavailable, err.Error())

internal/future/statestore/redis.go
sawagh: Done. Cached cfg on the redis backend object.

internal/future/statestore/redis.go
yfei1: I did some tests on the gjson package and it turns out gjson.Get does not search for the attribute recursively. For example.

```
gjson.Get("{\"hey\": \"false\", \"hi\": {\"hello\": \"true\"}", "hello")          // returns nothing
gjson.Get("{\"hey\": \"false\", \"hi\": {\"hello\": \"true\"}", "hi.hello")       // returns true
gjson.Get("{\"hey\": \"false\", \"hi\": {\"hello\": \"true\"}", "hey")            // returns false
```

so `gjson.Get` here basically is equivalent to 
```
m := make(map[string]interface{})
err := json.Unmarshal(data, &m)
if err != nil {
    log.Fatal(err)
}
fmt.Println(m[attribute])
```

+1 for using Any/determine JSON structure and use encoding/json

PR #386, opened 2019-05-10
Add website autopush for 0.5

jeremyje:
This change enables autopush for 0.5 releases by backporting the website to 0.5 branch.

Since the release is 0.5 it'll be pushed to 0-5 GAE service target as well as default because it matches the _GCB_LATEST_VERSION string.


sawagh:
Will this go in next hotfix? As in .. This will take effect immediately - but does this warrant an official hotfix release?

PR #390, opened 2019-05-11
slim down binary size

yfei1:
We are able to slim down the image size and speed up CI a bit by disabling symbol table and DWARF generation on binaries. People who prefer to debug when running binaries can still choose to attach gdb on the minimatch binary.


Makefile
```diff
@@ -597,23 +597,23 @@ service-binaries: cmd/future/minimatch/minimatch$(EXE_EXTENSION) cmd/future/back
 tools-binaries: tools/certgen/certgen$(EXE_EXTENSION)
 
 cmd/future/backend/backend$(EXE_EXTENSION): internal/future/pb/backend.pb.go internal/future/pb/backend.pb.gw.go api/backend.swagger.json
-	cd cmd/future/backend; $(GO_BUILD_COMMAND)
+	cd cmd/future/backend; $(GO_BUILD_COMMAND) -ldflags='-s -w' .
```
jeremyje: This is not self documenting and involves copy-paste. I'd alias this.
`SLIM_GO_BUILD = -ldflags='-s -w'` or something

jeremyje:
What's the time savings for this change? Because the downside potentially significant.

yfei1:
> What's the time savings for this change? Because the downside potentially significant.

I see a 10-15 seconds time save in cloud builder's `make all` step. What are the downsides? The go build commands in our Dockerfiles are untouched, so the added flags only affect the binaries being generated in the CI step, which will be discarded.

jeremyje:
The issue is that this complicates debugging and logging. This will be more apparent once tracing comes online. I'd say we should hold off on this is needed for more practical reasons like it's adding more perf or something.

Shaving off half a minute does not really seem to be worth the gain here. There's more impactful changes that can be made to speed up the build process.

PR #393, opened 2019-05-13
Clientwrapper

yfei1:
Add a reusable client wrapper for testing

### DIFF
--- a/internal/future/serving/client/grpc.go
+++ b/internal/future/serving/client/grpc.go
@@ -23,7 +23,16 @@ import (
 	"open-match.dev/open-match/internal/future/serving"
 )
 
-// TODO: provides TLS client support
+// SecureGRPCFromConfig creates a gRPC client connection with TLS support from a configuration.
+func SecureGRPCFromConfig(cfg config.View, prefix string, publicKeyBytes []byte) (*grpc.ClientConn, error) {
+	hostname := cfg.GetString(prefix + ".hostname")
+	portNumber := cfg.GetInt(prefix + ".grpcport")
+	return GRPCFromParams(&Params{
+		Hostname:          hostname,
+		Port:              portNumber,
+		PublicCertificate: publicKeyBytes,
+	})
+}
 
 // InsecureGRPCFromConfig creates a gRPC client connection without TLS support from a configuration.
 func InsecureGRPCFromConfig(cfg config.View, prefix string) (*grpc.ClientConn, error) {
--- a/internal/future/serving/client/grpc.go
+++ b/internal/future/serving/client/grpc.go
@@ -1,0 +1,49 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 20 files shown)


internal/future/serving/client/grpc.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2019 Google LLC
```
jeremyje: Merge this into /rpc/ directory and remove all exported methods that are only needed because they are in separate packages.

internal/future/serving/client/grpc_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+func TestInsecureGRPCFromConfig(t *testing.T) {
+	runGrpcClientTests(t, false)
+}
+
+func runGrpcClientTests(t *testing.T, useTLS bool) {
+	assert := assert.New(t)
+
+	var (
+		configPrefix      = "grpcTest"
```
jeremyje: This is very strange.
use `:=` and define `grpcConn` in place.

internal/future/serving/client/grpc_test.go
```diff
@@ -0,0 +1,99 @@
...
+	cfg.Set(clientGRPCPortKey, grpcLh.Number())
+
+	grpcAddress := fmt.Sprintf("localhost:%d", grpcLh.Number())
+	proxyAddress := fmt.Sprintf("localhost:%d", httpLh.Number())
+	allHostnames := []string{grpcAddress, proxyAddress}
+	pub, priv, err := certgenTesting.CreateCertificateAndPrivateKeyForTesting(allHostnames)
+	assert.Nil(err)
+
+	// TLS setup
+	if useTLS {
```
jeremyje: This is usual because you just created the cert key pairs just above and that's only for the TLS codepaths.

internal/future/serving/client/http.go
```diff
@@ -0,0 +1,74 @@
...
+	"fmt"
+	"net/http"
+	"time"
+
+	"open-match.dev/open-match/internal/config"
+	"open-match.dev/open-match/internal/future/serving"
+)
+
+// SecureHTTPFromConfig creates an HTTPS client from a configuration.
+func SecureHTTPFromConfig(cfg config.View, prefix string, publicKeyBytes, privateKeyBytes []byte) (*http.Client, string, error) {
```
jeremyje: If there's a config file I'd expect paths to the pub/priv key set there and this method loads them in.

internal/future/serving/tls_util.go
```diff
@@ -0,0 +1,78 @@
+// Copyright 2019 Google LLC
```
jeremyje: I think these change can happen in multiple steps.
* Move files to /rpc/ directory. You can drop future since we are deleting the obsoleted code at this point.
* Break out the TLS util stuff
* Add client code.

yfei1: I'll drop future when the original serving package is not being referenced at all. We still have some example codes that are using it.

internal/future/serving/client/grpc.go
yfei1: Do I need if it make sense to merge the grpc client and http client into one file?

PR #394, opened 2019-05-13
Remove unused config methods

yfei1:
1. Remove unused command from Makefile
2. Consolidate config.go. Correct me if I am wrong :) I think we won't use the Sub function and envmappings in v.6 (Sub was used for ignorelist settings in v.5 and we have all the env bindings in the values.yaml file now)


internal/config/config.go
```diff
@@ -67,6 +43,19 @@ var (
 	}
 )
 
+// View is a read-only view of the Open Match configuration.
+// New accessors from Viper should be added here.
+type View interface {
```
jeremyje: Why are you removing view.go and the tests?

yfei1: we are not using `view.Sub` function anymore so basically view.go has only this structure defined. I remove view.go and the tests to consolidate the repo a bit.

internal/config/config.go
```diff
@@ -25,36 +27,10 @@ import (
...
 var (
 	// Logrus structured logging setup
-	logFields = logrus.Fields{
+	cfgLog = logrus.WithFields(logrus.Fields{
 		"app":       "openmatch",
 		"component": "config",
-	}
-	cfgLog = logrus.WithFields(logFields)
-
-	// Map of the config file keys to environment variable names populated by
```
jeremyje: Do we have a replacement connection config ready for Redis yet? If not I wouldn't move this. The reason we kept around the old code was there's a lot of scaffolding we didn't want to have to rewrite. This is part of that scaffolding.

internal/config/view_test.go
```diff
@@ -1,47 +0,0 @@
...
-	v.Set("a.b", "a.b")
-	av := Sub(v, "a")
-	if av == nil {
-		t.Fatalf("Sub(%v, 'a') => %v", v, av)
-	}
-
-	if av.GetString("a") != "a.a" {
-		t.Errorf("av.GetString('a') = %s, expected 'a.a'", av.GetString("a"))
-	}
-	if av.GetString("a.a") != "" {
```
jeremyje: Keep the test and just remove the sub call. Since av isn't there. You can update with `v.GetString()` calls that always return their inputs.

PR #395, opened 2019-05-13
Get started guide

yfei1:
Left some TODOs in L26-29 in docs/get_started.md file.


docs/get_started.md
```diff
@@ -0,0 +1,38 @@
+# Get Started
```
jeremyje: This document should be added to the website.
The rule is if you're modifying Open Match then /docs/ otherwise /site/...

PR #415, opened 2019-05-15
Implement clientwrapper for harness and testing

yfei1:


internal/rpc/clients_test.go
```diff
@@ -0,0 +1,186 @@
...
+	defer removeTempFile(assert, priFile.Name())
+
+	cfg, rpcParams := configureConfigAndKeysForTesting(assert, configPrefix, pubFile.Name(), priFile.Name())
+
+	runGrpcClientTests(assert, configPrefix, cfg, rpcParams)
+}
+
+func TestInsecureGRPCFromConfig(t *testing.T) {
+	assert := assert.New(t)
+	configPrefix := "httpTest"
```
golangcibot: string `httpTest` has 3 occurrences, make it a constant (from `goconst`)

internal/rpc/clients.go
```diff
@@ -0,0 +1,151 @@
...
+		PrivateKeyData:    privateKeyBytes,
+	})
+}
+
+// HTTPFromParams creates a HTTP client from the parameters.
+func HTTPClientFromParams(params *ClientParams) (*http.Client, string, error) {
+	address := fmt.Sprintf("%s:%d", params.Hostname, params.Port)
+	if params.usingTLS() {
+		baseURL := "https://" + address
+		tlsCert, err := certificateFromFileData(params.PublicCertificate, params.PrivateKeyData)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

internal/rpc/clients_test.go
```diff
@@ -0,0 +1,186 @@
...
+		pb.RegisterFrontendServer(s, ff)
+	}, pb.RegisterFrontendHandlerFromEndpoint)
+	s := &Server{}
+	defer s.Stop()
+	waitForStart, err := s.Start(rpcParams)
+	assert.Nil(err)
+	waitForStart()
+
+	// Acquire http client
+	httpClient, baseURL, err := HTTPClientFromConfig(cfg, configPrefix)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

internal/rpc/clients.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+var (
+	clientLogger = logrus.WithFields(logrus.Fields{
+		"app":       "openmatch",
+		"component": "client",
+	})
+)
+
+// Params contains the connection parameters to connect to an Open Match service.
```
golangcibot: comment on exported type ClientParams should be of the form "ClientParams ..." (with optional leading article) (from `golint`)

internal/rpc/clients.go
```diff
@@ -0,0 +1,151 @@
...
+	Port              int
+	PublicCertificate []byte
+	PrivateKeyData    []byte
+}
+
+func (p *ClientParams) usingTLS() bool {
+	return len(p.PublicCertificate) > 0
+}
+
+// SecureGRPCFromConfig creates a gRPC client connection with TLS support from a configuration.
```
golangcibot: comment on exported function `GRPCClientFromConfig` should be of the form `GRPCClientFromConfig ...` (from `golint`)

internal/rpc/clients.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	return GRPCClientFromParams(&ClientParams{
+		Hostname:          hostname,
+		Port:              portNumber,
+		PublicCertificate: publicKeyBytes,
+	})
+}
+
+// GRPCFromParams creates a gRPC client connection from the parameters.
```
golangcibot: comment on exported function `GRPCClientFromParams` should be of the form `GRPCClientFromParams ...` (from `golint`)

internal/rpc/clients.go
```diff
@@ -0,0 +1,151 @@
...
+		creds, err := clientCredentialsFromFileData(params.PublicCertificate, address)
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+		return grpc.Dial(address, grpc.WithTransportCredentials(creds))
+	}
+	return grpc.Dial(address, grpc.WithInsecure())
+}
+
+func HTTPClientFromConfig(cfg config.View, prefix string) (*http.Client, string, error) {
```
golangcibot: exported function `HTTPClientFromConfig` should have comment or be unexported (from `golint`)

internal/rpc/clients.go
```diff
@@ -0,0 +1,151 @@
...
+
+	return HTTPClientFromParams(&ClientParams{
+		Hostname:          hostname,
+		Port:              portNumber,
+		PublicCertificate: publicKeyBytes,
+		PrivateKeyData:    privateKeyBytes,
+	})
+}
+
+// HTTPFromParams creates a HTTP client from the parameters.
```
golangcibot: comment on exported function `HTTPClientFromParams` should be of the form `HTTPClientFromParams ...` (from `golint`)

internal/rpc/clients.go
```diff
@@ -0,0 +1,155 @@
...
+			return nil, errors.WithStack(err)
+		}
+		return grpc.Dial(address, grpc.WithTransportCredentials(creds))
+	}
+	return grpc.Dial(address, grpc.WithInsecure())
+}
+
+// HTTPClientFromConfig creates a HTTP client from from a configuration.
+func HTTPClientFromConfig(cfg config.View, prefix string) (*http.Client, string, error) {
+	hostname := cfg.GetString(prefix + ".hostname")
```
Laremere: This code would be cleaner if you make a ClientParams variable at the top, and fill it out throughout the function.  For the whole if statement about tls being enabled, you can invert it and only run it if it is enabled.  Within the if statement, add the fields tls parameters to the existing ClientParams variable.

internal/rpc/clients.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+		tlsTransport := &http.Transport{
+			TLSClientConfig: &tls.Config{
+				ServerName:   address,
+				RootCAs:      pool,
+				Certificates: []tls.Certificate{*tlsCert},
+			},
+		}
+		httpClient := &http.Client{
+			Timeout:   time.Second * 10,
```
Laremere: Why is the timeout different for tls?

internal/rpc/clients.go
```diff
@@ -0,0 +1,155 @@
...
+		PublicCertificate: publicKeyBytes,
+		PrivateKeyData:    privateKeyBytes,
+	})
+}
+
+// HTTPClientFromParams creates a HTTP client from the parameters.
+func HTTPClientFromParams(params *ClientParams) (*http.Client, string, error) {
+	address := fmt.Sprintf("%s:%d", params.Hostname, params.Port)
+	if params.usingTLS() {
+		baseURL := "https://" + address
```
Laremere: Why does baseurl include a final "/" not tls, but ommits it for tls?

internal/rpc/clients.go
```diff
@@ -0,0 +1,155 @@
...
+	return HTTPClientFromParams(&ClientParams{
+		Hostname:          hostname,
+		Port:              portNumber,
+		PublicCertificate: publicKeyBytes,
+		PrivateKeyData:    privateKeyBytes,
+	})
+}
+
+// HTTPClientFromParams creates a HTTP client from the parameters.
+func HTTPClientFromParams(params *ClientParams) (*http.Client, string, error) {
```
Laremere: Feels like the same pattern could be applied here, with a client created at the beginning of the function, and the tls transport optionally set.  There are some other differences, so idk.

Laremere: Should this be working with a https://golang.org/pkg/net/url/#URL instead of a raw string?

internal/rpc/clients.go
```diff
@@ -0,0 +1,155 @@
...
+
+	return HTTPClientFromParams(&ClientParams{
+		Hostname:          hostname,
+		Port:              portNumber,
+		PublicCertificate: publicKeyBytes,
+		PrivateKeyData:    privateKeyBytes,
+	})
+}
+
+// HTTPClientFromParams creates a HTTP client from the parameters.
```
Laremere: Why is a certificate required to use https as a client??

internal/rpc/clients.go
```diff
@@ -0,0 +1,155 @@
...
+	return len(p.PublicCertificate) > 0
+}
+
+// GRPCClientFromConfig creates a gRPC client connection from a configuration.
+func GRPCClientFromConfig(cfg config.View, prefix string) (*grpc.ClientConn, error) {
+	hostname := cfg.GetString(prefix + ".hostname")
+	portNumber := cfg.GetInt(prefix + ".grpcport")
+
+	// If TLS support is not enabled in the config, return an insecure grpc client.
+	if !cfg.GetBool("tls.enabled") {
```
Laremere: Same as comment for http.

internal/rpc/clients.go
```diff
@@ -0,0 +1,155 @@
...
+	publicKeyBytes, err := ioutil.ReadFile(publicKeyPath)
+	if err != nil {
+		clientLogger.WithError(err).Fatalf("cannot read public key from file %s", publicKeyPath)
+		return nil, err
+	}
+
+	return GRPCClientFromParams(&ClientParams{
+		Hostname:          hostname,
+		Port:              portNumber,
+		PublicCertificate: publicKeyBytes,
```
Laremere: Seems weird to me that public key would be required here.  What public key is this?

yfei1: > Seems weird to me that public key would be required here. What public key is this?

This is actually the public key from the one who signed the certificate. Client will need to use it to decrypt the signature of server certificate. I'll probably rename it to PublicKeyForCertificate

internal/rpc/clients.go
yfei1: See the comment above.

internal/rpc/clients.go
yfei1: I'll leave off the trailing slash. For tls the slash does not matter as we do rpc via the client. For http the baseurl is needed to construct a complete endpoint and the URL path typically includes the slash

internal/rpc/clients.go
yfei1: Will fix.

internal/rpc/clients.go
yfei1: Will fix

Laremere: Please only resolve once you have the fix in the PR.

internal/rpc/clients.go
```diff
@@ -0,0 +1,129 @@
...
+
+	return GRPCClientFromParams(clientParams)
+}
+
+// GRPCClientFromParams creates a gRPC client connection from the parameters.
+func GRPCClientFromParams(params *ClientParams) (*grpc.ClientConn, error) {
+	address := fmt.Sprintf("%s:%d", params.Hostname, params.Port)
+
+	if params.usingTLS() {
+		publicCertFileData, err := ioutil.ReadFile(params.TrustedKeyPath)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

internal/rpc/clients.go
```diff
@@ -0,0 +1,129 @@
...
+		pool, err := trustedCertificatesFromFile(params.TrustedKeyPath)
+		if err != nil {
+			return nil, "", err
+		}
+
+		tlsTransport := &http.Transport{
+			TLSClientConfig: &tls.Config{
+				ServerName:   address,
+				RootCAs:      pool,
+				Certificates: []tls.Certificate{*tlsCert},
```
golangcibot: invalid operation: cannot indirect tlsCert (variable of type credentials.TransportCredentials) (from `typecheck`)

internal/rpc/clients.go
```diff
@@ -0,0 +1,129 @@
...
+			},
+		}
+		httpClient.Transport = tlsTransport
+	}
+
+	baseURL := "http://" + address
```
golangcibot: no new variables on left side of := (from `typecheck`)

internal/rpc/clients.go
```diff
@@ -0,0 +1,125 @@
...
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	"open-match.dev/open-match/internal/config"
+)
+
+var (
+	clientLogger = logrus.WithFields(logrus.Fields{
```
golangcibot: `clientLogger` is unused (from `deadcode`)

internal/rpc/clients.go
```diff
@@ -0,0 +1,155 @@
...
+		Port:              portNumber,
+		PublicCertificate: publicKeyBytes,
+	})
+}
+
+// GRPCClientFromParams creates a gRPC client connection from the parameters.
+func GRPCClientFromParams(params *ClientParams) (*grpc.ClientConn, error) {
+	address := fmt.Sprintf("%s:%d", params.Hostname, params.Port)
+	if params.usingTLS() {
+		creds, err := clientCredentialsFromFileData(params.PublicCertificate, address)
```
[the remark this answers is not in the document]
yfei1: Will do in a separate PR

internal/rpc/clients.go
```diff
@@ -0,0 +1,129 @@
...
+	} else {
+		grpcOptions = append(grpcOptions, grpc.WithInsecure())
+	}
+
+	return grpc.Dial(address, grpcOptions...)
+}
+
+// HTTPClientFromConfig creates a HTTP client from from a configuration.
+func HTTPClientFromConfig(cfg config.View, prefix string) (*http.Client, string, error) {
+	hostname := cfg.GetString(prefix + ".hostname")
```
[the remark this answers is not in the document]
yfei1: These variables are used to initialize clients from different componenets. We are able to get a frondendclient by directly calling HTTPClientFromConfig(cfg, "frontend"). We have the same settings in server side wrappers as well.

internal/rpc/clients_test.go
```diff
@@ -0,0 +1,186 @@
...
+	shellTesting "open-match.dev/open-match/internal/testing"
+	netlistenerTesting "open-match.dev/open-match/internal/util/netlistener/testing"
+	certgenTesting "open-match.dev/open-match/tools/certgen/testing"
+)
+
+func TestSecureGRPCFromConfig(t *testing.T) {
+	assert := assert.New(t)
+
+	// Create temporary TLS key files for testing
+	pubFile, err := ioutil.TempFile("", "pub*")
```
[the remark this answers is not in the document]
yfei1: This file is used in the configureConfigAndKeysForTesting function. The helper function will generate some public/private key pairs and write the content into this temporary file. See internal/rpc/clients_test.go#L168 for details.

internal/rpc/clients.go
```diff
@@ -0,0 +1,123 @@
...
+		"app":       "openmatch",
+		"component": "client",
+	})
+)
+
+// ClientParams contains the connection parameters to connect to an Open Match service.
+type ClientParams struct {
+	Hostname       string
+	Port           int
+	TrustedKeyPath string
```
jeremyje: Shouldn't this be `TrustedCertificate []byte`
The reason why the file data (in PEM format) is specified is that there's 2 paths to load from, a file or the PEM data. They both converge on the PEM data so it make sense of the code flow join to happen there.

internal/rpc/clients.go
```diff
@@ -0,0 +1,123 @@
...
+// GRPCClientFromConfig creates a gRPC client connection from a configuration.
+func GRPCClientFromConfig(cfg config.View, prefix string) (*grpc.ClientConn, error) {
+	clientParams := &ClientParams{
+		Hostname: cfg.GetString(prefix + ".hostname"),
+		Port:     cfg.GetInt(prefix + ".grpcport"),
+	}
+
+	// If TLS support is enabled in the config, fill in the trusted key for decrpting server certificate.
+	if cfg.GetBool("tls.enabled") {
+		clientParams.TrustedKeyPath = cfg.GetString("tls.trustedKeyPath")
```
jeremyje: Read the configuration here.

internal/rpc/clients.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// GRPCClientFromParams creates a gRPC client connection from the parameters.
+func GRPCClientFromParams(params *ClientParams) (*grpc.ClientConn, error) {
+	address := fmt.Sprintf("%s:%d", params.Hostname, params.Port)
+
+	grpcOptions := []grpc.DialOption{}
+
+	if params.usingTLS() {
+		creds, err := clientCredentialsFromFile(params.TrustedKeyPath)
```
jeremyje: This does not support the code flow of the PEM data already loaded in memory which is used for tests.

internal/rpc/clients.go
```diff
@@ -0,0 +1,123 @@
...
+// HTTPClientFromConfig creates a HTTP client from from a configuration.
+func HTTPClientFromConfig(cfg config.View, prefix string) (*http.Client, string, error) {
+	clientParams := &ClientParams{
+		Hostname: cfg.GetString(prefix + ".hostname"),
+		Port:     cfg.GetInt(prefix + ".httpport"),
+	}
+
+	// If TLS support is enabled in the config, fill in the trustedKeyPath
+	if cfg.GetBool("tls.enabled") {
+		clientParams.TrustedKeyPath = cfg.GetString("tls.trustedKeyPath")
```
jeremyje: You'll want to validate that the file exists and can be read here and return a reasonable error otherwise.

internal/rpc/clients.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return HTTPClientFromParams(clientParams)
+}
+
+// HTTPClientFromParams creates a HTTP client from the parameters.
+func HTTPClientFromParams(params *ClientParams) (*http.Client, string, error) {
+	address := fmt.Sprintf("%s:%d", params.Hostname, params.Port)
+	httpClient := &http.Client{Timeout: time.Second * 3}
```
jeremyje: Should this be a config value? At least add a TODO here.

internal/rpc/clients.go
```diff
@@ -0,0 +1,146 @@
...
+// HTTPClientFromConfig creates a HTTP client from from a configuration.
+func HTTPClientFromConfig(cfg config.View, prefix string) (*http.Client, string, error) {
+	clientParams := &ClientParams{
+		Hostname: cfg.GetString(prefix + ".hostname"),
+		Port:     cfg.GetInt(prefix + ".httpport"),
+	}
+
+	// If TLS support is enabled in the config, fill in the trusted certificates for decrpting server certificate.
+	if cfg.GetBool("tls.enabled") {
+		_, err := os.Stat(cfg.GetString("tls.trustedCertificatesPath"))
```
jeremyje: Not sure about how this is cased.
Also trustedCertificatePath (no s)

internal/rpc/clients.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+	}
+
+	return HTTPClientFromParams(clientParams)
+}
+
+// HTTPClientFromParams creates a HTTP client from the parameters.
+func HTTPClientFromParams(params *ClientParams) (*http.Client, string, error) {
+	address := fmt.Sprintf("%s:%d", params.Hostname, params.Port)
+	// Make client Timeout configurable
```
jeremyje: `// TODO: `

PR #416, opened 2019-05-15
Enable most of the golangci checks and fix internal/set tests

yfei1:
1. Enable most of the golangci checks except the errcheck plugins for tests
2. Rewrite internal/set tests to avoid using global variables
3. Fix some other golangci issues


internal/config/config.go
```diff
@@ -30,32 +30,6 @@ var (
...
-	// REDIS_SENTINEL_PORT_6379_TCP_ADDR=10.55.253.195
-	// REDIS_SENTINEL_SERVICE_PORT=6379
-	// REDIS_SENTINEL_PORT_6379_TCP_PORT=6379
-	// REDIS_SENTINEL_PORT_6379_TCP_PROTO=tcp
-	// REDIS_SENTINEL_SERVICE_HOST=10.55.253.195
-	//
-	// MMFs are expected to get their configuation from env vars instead
-	// of reading the config file.  So, config parameters that are required
-	// by MMFs should be populated to env vars.
-	envMappings = map[string]string{
```
jeremyje: Isn't this necessary because of how we use Redis?

PR #417, opened 2019-05-15
Implement Frontend Service Create / Get / Delete Ticket

sawagh:


internal/app/frontend/frontend_service.go
```diff
@@ -45,32 +56,87 @@ func newFrontend(cfg config.View) (*frontendService, error) {
...
-// for the attributes defined as indices the matchmakaking config. If the
-// attributes exist and are valid integers, they will be indexed. Creating a
-// ticket adds the Ticket to the pool of Tickets considered for matchmaking.
+// CreateTicket will create a new ticket, assign an id to it and put it in state
+// storage. It will index the Ticket attributes based on the indexing configuration.
+// Indexing a Ticket adds the it to the pool of Tickets considered for matchmaking.
 func (s *frontendService) CreateTicket(ctx context.Context, req *pb.CreateTicketRequest) (*pb.CreateTicketResponse, error) {
-	return &pb.CreateTicketResponse{}, nil
+	// Generate a ticket id and create a Ticket in state storage
+	ticket := proto.Clone(req.Ticket).(*pb.Ticket)
```
golangcibot: Error return value is not checked (from `errcheck`)

sawagh: added nil check

yfei1: you might missed the type assertion check at here
ticket, ok := proto.Clone(req.Ticket).(*pb.Ticket)
if !ok {
// type assertion failed
}

internal/app/frontend/frontend_service.go
```diff
@@ -45,32 +57,92 @@ func newFrontend(cfg config.View) (*frontendService, error) {
...
+	if ticket == nil {
```
yfei1: ticket wont be nil if the succeeded

internal/app/frontend/frontend_service.go
```diff
@@ -45,32 +57,92 @@ func newFrontend(cfg config.View) (*frontendService, error) {
...
+// deleteTicket is a 'lazy' ticket delete that should be called after a ticket
+// has been deindexed.
+func (s *frontendService) deleteTicket(id string) {
+	err := s.store.DeleteTicket(context.Background(), id)
+	if err != nil {
+		logger.WithFields(logrus.Fields{
+			"error": err.Error(),
+			"id":    id,
+		}).Error("failed to delete the ticket")
+		return
```
yfei1: Do we need a retry logic at here if the call to `store.DeleteTicket` failed?

internal/app/frontend/frontend_test.go
```diff
@@ -32,3 +40,90 @@ func TestServerBinding(t *testing.T) {
...
+		}
+
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	assert.Failf(t, "ticket %v not deleted after 5 seconds", id)
+}
+
+// TestFrontendService tests creating, getting and deleting a ticket using Frontend service.
+func TestFrontendService(t *testing.T) {
```
yfei1: This looks more like a statestore test. Is it possible to use a frontend client for testing?

internal/app/frontend/frontend_service.go
```diff
@@ -45,32 +57,93 @@ func newFrontend(cfg config.View) (*frontendService, error) {
...
 func (s *frontendService) CreateTicket(ctx context.Context, req *pb.CreateTicketRequest) (*pb.CreateTicketResponse, error) {
-	return &pb.CreateTicketResponse{}, nil
+	// Perform input validation.
+	if req.Ticket == nil {
+		logger.Error("invalid argument - ticket cannot be nil")
+		return nil, status.Errorf(codes.InvalidArgument, "ticket cannot be nil")
+	}
+
+	// Generate a ticket id and create a Ticket in state storage
+	ticket := proto.Clone(req.Ticket).(*pb.Ticket)
```
golangcibot: Error return value is not checked (from `errcheck`)

internal/app/frontend/frontend_service.go
sawagh: Not here - given that there are no known failure paths we are trying to retry on. At some point, we should explore if there needs to be some garbage collection to catch any data in the database that is stray.

internal/app/frontend/frontend_test.go
sawagh: What do you mean? This test fetches a frontend client from newFrontend and then tests all methods only from the frontend client. The frontend client will always need a state store - so yes, we are creating one here - but this does test FE Client methods. We should have E2E tests that talk via APIs to the frontend service - but this is just testing the client. (I am hoping our integration tests account for those)

internal/app/frontend/frontend_service.go
sawagh: The input can be nil right? I ve moved the check directly on the input now. Added the casting error check here so gocli should be happy.

internal/app/frontend/frontend_service.go
```diff
@@ -45,32 +57,98 @@ func newFrontend(cfg config.View) (*frontendService, error) {
...
-// for the attributes defined as indices the matchmakaking config. If the
-// attributes exist and are valid integers, they will be indexed. Creating a
-// ticket adds the Ticket to the pool of Tickets considered for matchmaking.
+// CreateTicket will create a new ticket, assign an id to it and put it in state
+// storage. It will index the Ticket attributes based on the indexing configuration.
+// Indexing a Ticket adds the it to the pool of Tickets considered for matchmaking.
 func (s *frontendService) CreateTicket(ctx context.Context, req *pb.CreateTicketRequest) (*pb.CreateTicketResponse, error) {
-	return &pb.CreateTicketResponse{}, nil
+	// Perform input validation.
+	if req.Ticket == nil {
```
yfei1: The `req` itself could be nil, but grpc protocol does not allow any clients to send nil request so I think this check is not needed.

sawagh: Yes - but this check is not checking if request is nil - it is checking if ticket inside of the request is nil. Wont &pb.CreateTicketRequest{} be a valid request (not nil - but empty) - and still have ticket as a nil in there?

yfei1: Ah you are right, but it seems like we are missing this nil check everywhere.. maybe worth to send a separate pr to fix them.

PR #420, opened 2019-05-16
Replace manual miniredis setup with statestore testing utility code

yfei1:
Expose a statestore util function for testing purpose and replace repetitive miniredis manual setup this function.

### DIFF
--- a/internal/app/frontend/frontend_test.go
+++ b/internal/app/frontend/frontend_test.go
@@ -19,16 +19,15 @@ import (
 	"testing"
 	"time"
 
-	"github.com/alicebob/miniredis"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	"open-match.dev/open-match/internal/config"
 	"open-match.dev/open-match/internal/pb"
 	"open-match.dev/open-match/internal/rpc"
 	rpcTesting "open-match.dev/open-match/internal/rpc/testing"
+	statestoreTesting "open-match.dev/open-match/internal/statestore/testing"
 )
 
 func TestServerBinding(t *testing.T) {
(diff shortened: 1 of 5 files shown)


internal/app/mmlogic/mmlogic_test.go
```diff
@@ -17,12 +17,30 @@ package mmlogic
...
 	"google.golang.org/grpc"
 	"open-match.dev/open-match/internal/pb"
 	"open-match.dev/open-match/internal/rpc"
 	rpcTesting "open-match.dev/open-match/internal/rpc/testing"
+	statestoreTesting "open-match.dev/open-match/internal/statestore/testing"
 )
 
+func TestQueryTicketsBasic(t *testing.T) {
+	assert := assert.New(t)
+	vp := viper.New()
```
Laremere: why is this "vp" instead of "cfg"?

internal/app/frontend/frontend_test.go
```diff
@@ -76,7 +75,12 @@ func validateDelete(t *testing.T, fe *frontendService, id string) {
 // TestFrontendService tests creating, getting and deleting a ticket using Frontend service.
 func TestFrontendService(t *testing.T) {
 	assert := assert.New(t)
-	cfg := createStore(t)
+	vp := viper.New()
+	vp.Set("playerIndices", []string{"testindex1", "testindex2"})
```
Laremere: This is also set by NewRedisForTesting.

internal/app/mmlogic/mmlogic_test.go
yfei1: because cfg is used at Line 34.

internal/statestore/testing/public.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"time"
+
+	"github.com/alicebob/miniredis"
+	"github.com/spf13/viper"
+	"open-match.dev/open-match/internal/config"
+)
+
+// NewRedisForTesting create a new in-memory miniredis for testing
+func NewRedisForTesting(cfg *viper.Viper) (config.View, func(), error) {
```
sawagh: NewTestStateStorage would be more apt. Might be best to abstract the fact that there is redis under the covers - from any outside components using this.

sawagh: This should not expose the closer - any component using this cfg to build a StateStore object should be calling .Close() on the public state store method. 

Do consider not reusing this helper for redis_test.go.

yfei1:
Closed as #462 resolved most of the issues in this PR.

PR #421, opened 2019-05-16
Proto changes for implementing Backend Service

sawagh:
Breaking down backend service changes - submitting proto changes separately.


internal/harness/golang/matchfunction_service.go
```diff
@@ -93,22 +91,11 @@ func (s *matchFunctionService) Run(req *pb.RunRequest, mfServer pb.MatchFunction
...
-		default:
-			err := mfServer.Send(&pb.RunResponse{Proposal: match})
-			if err != nil {
-				matchfunctionLogger.WithError(err).Error("Failed to send Run response to grpc server.")
-				return status.Error(codes.Aborted, err.Error())
-			}
-		}
-	}
-	return nil
+	proposals := s.function(mfView)
```
jeremyje: Was deleting this code intentional?

sawagh: Yes  - previous one fetched and streamed each match back - now we return the set of matches.

PR #424, opened 2019-05-17
Add a new Evaluator service to Open Match core components to perform match evaluation

sawagh:


internal/app/evaluator/evaluator_service.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	return es, nil
+}
+
+// Evaluate accepts a list of matches, triggers the user configured evaluation
+// function with these and other matches in the evaluation window and returns
+// matches that are accepted by the Evaluator as valid results.
+func (s *evaluatorService) Evaluate(ctx context.Context, req *pb.EvaluateRequest) (*pb.EvaluateResponse, error) {
+	return &pb.EvaluateResponse{}, nil
```
Laremere: I think a better "dumb" evaluator would be one which returns all of the matches from the request.  That way it doesn't block end to end working.

.golangci.yaml
```diff
@@ -196,10 +196,12 @@ issues:
   exclude-rules:
     - path: internal[/\\]app[/\\]backendapi[/\\]
       linters: [goimports, gochecknoglobals, errcheck, misspell, govet, unparam, golint, gofmt]
-    - path: internal[/\\]app[/\\]backendapi[/\\]
+    - path: internal[/\\]app[/\\]frontendapi[/\\]
```
yfei1: These rules were set up to exclude lint checking the legacy codes from v.5. #416 will disable all these rules from now on.

Makefile
```diff
@@ -806,6 +824,12 @@ proxy-mmlogic: build/toolchain/bin/kubectl$(EXE_EXTENSION)
 	@echo "Trace: http://localhost:$(MMLOGIC_PORT)/debug/tracez"
 	$(KUBECTL) port-forward --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) $(shell $(KUBECTL) get pod --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) --selector="app=open-match,component=mmlogic,release=$(OPEN_MATCH_CHART_NAME)" --output jsonpath='{.items[0].metadata.name}') $(MMLOGIC_PORT):51503 $(PORT_FORWARD_ADDRESS_FLAG)
 
+proxy-evaluator: build/toolchain/bin/kubectl$(EXE_EXTENSION)
+	@echo "Health: http://localhost:$(EVALUATOR_PORT)/healthz"
+	@echo "RPC: http://localhost:$(EVALUATOR_PORT)/debug/rpcz"
+	@echo "Trace: http://localhost:$(EVALUATOR_PORT)/debug/tracez"
+	$(KUBECTL) port-forward --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) $(shell $(KUBECTL) get pod --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) --selector="app=open-match,component=evaluator,release=$(OPEN_MATCH_CHART_NAME)" --output jsonpath='{.items[0].metadata.name}') $(EVALUATOR_PORT):51505 $(PORT_FORWARD_ADDRESS_FLAG)
```
yfei1: $(EVALUATOR_PORT):51506

internal/app/evaluator/evaluator_service.go
sawagh: This change is just scaffolding for the evaluator. Current state is that the backend service does not wire any evaluator at all. My planned next steps:
- Try to get an E2E working without evaluator
- Wire up a simple evaluator (time window based evaluation - but passthrough - no real evaluation), make sure E2E works.
- Wire up the real evaluation logic and validate that E2E continues to work.

I ve updated the commit message - thanks for feedback.

PR #429, opened 2019-05-17
 Have redis filter query return full tickets

Laremere:
This changes the statestore interface, and the redis implementation to pass the full tickets to a callback.  This aligns with recommendation from unity that the vast majority of the time the MMF will want the full ticket.  This also makes the next change, switching from incorrect int usage to using floats easier.  (which is itself required to make the proto change to use struct.)

I'm not happy with the final state here, but plan on coming back to it after I make the changes required to the proto files.

This change also deletes redis_test.go because it doesn't properly set up the state of the database.  Once the changes to the protos are in, a more normal test can be written.  It's also lacking unit tests on helper methods within the filter endpoint, as they are also going to change heavily based on proto changes.


internal/statestore/redis.go
```diff
@@ -384,42 +382,75 @@ func (rb *redisBackend) FilterTickets(ctx context.Context, filters []*pb.Filter)
...
-		}
-
-		attributeToTickets[filter.Attribute] = idToValue
-
-		idsPerFilter := make([]string, 0)
-		for id := range idToValue {
-			idsPerFilter = append(idsPerFilter, id)
+				"Command": fmt.Sprintf("ZRANGEBYSCORE %s %f %f", filter.Attribute, filter.Min, filter.Max),
+			}).WithError(err).Error("Failed to lookup index.")
+			return err
```
jeremyje: We should be returning grpc errors with status codes.

PR #430, opened 2019-05-18
Use proto3's struct for properties, not string.

Laremere:
This makes the API more usuable for both json clients (who no longer
have to encode json into a string and put that into their json) and also
for proto clients (who no longer have to use json...).  When converted
to json, struct will be encoded directly into an object, which is much
more convenient.

The majority of the rest of this change is fixing tickets in tests.

This removes the dependancy on the json parser that was used for reading
from properties.

I also added a test on indexing and filtering tickets from the state
store, because I can't help myself and I have a problem.


internal/statestore/public_test.go
```diff
@@ -91,8 +97,73 @@ func TestTicketLifecycle(t *testing.T) {
...
-	// Testing Ticket indexing will be added along with this implementation.
+	// Create State Store
+	assert := assert.New(t)
+	cfg := createBackend(t)
+	service, err := New(cfg)
+	assert.Nil(err)
+	assert.NotNil(service)
+	defer service.Close()
+
+	for i := 0; i < 10; i += 1 {
```
golangcibot: should replace `i += 1` with `i++` (from `golint`)

internal/statestore/public_test.go
```diff
@@ -91,8 +97,73 @@ func TestTicketLifecycle(t *testing.T) {
...
+			Max:       8.5,
+		},
+		{
+			Attribute: "testindex2",
+			Min:       0.49,
+			Max:       0.51,
+		},
+	}
+
+	err = service.FilterTickets(context.Background(), filters, 2, func(tickets []*pb.Ticket) error {
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

internal/statestore/public_test.go
golangcibot: should replace `i += 1` with `i++` (from `golint`)

PR #434, opened 2019-05-21
Fail FetchMatches if match configuration is missing or unsupported

sawagh:
Add input validation to FetchMatches. The call now fails with InvalidArgument if match configuration is missing or unsupported.

### DIFF
--- a/internal/app/backend/backend_service.go
+++ b/internal/app/backend/backend_service.go
@@ -65,6 +65,12 @@ func newBackend(cfg config.View) (*backendService, error) {
 // proposals which are then evaluated to generate results. FetchMatches method
 // streams these results back to the caller.
 func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Backend_FetchMatchesServer) error {
+	// Validate that the function configuration and match profiles are provided.
+	if req.Config == nil {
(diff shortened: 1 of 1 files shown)


internal/app/backend/backend_service.go
```diff
@@ -65,6 +65,17 @@ func newBackend(cfg config.View) (*backendService, error) {
 // proposals which are then evaluated to generate results. FetchMatches method
 // streams these results back to the caller.
 func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Backend_FetchMatchesServer) error {
+	// Validate that the function configuration and match profiles are provided.
+	if req.Config == nil {
+		logger.Error("invalid argument - match function config cannot be nil")
+		return status.Errorf(codes.InvalidArgument, "match function config cannot be nil")
```
Laremere: This wording makes sense for Go callers, but not from callers in other languages.  Please word it in a way which is more natural to protobufs and rest callers.  Eg, "FetchMatchesRequest requires config to be set."

sawagh: Agreed that nil was a golang specific reference. Removed it. User visible errors need to be user friendly though - so i dont think protobuf specific references should be in them either. Replaced with a generic user friendly string.

PR #437, opened 2019-05-23
Fix minor issues that were encountered in an E2E test case for generating matches.

sawagh:
Here are the issues this change fixes:

1. Add error check to avoid accessing failed connection in tcp net listener
2. Divide by zero error in the redis state storage library if page size is 0.
3. Ability to support null hostname when connecting to mmlogic client in match function harness.
4. Set min / max page size limits to initialize correctly if page size is set to unsupported values.

### DIFF
--- a/internal/app/mmlogic/mmlogic_service.go
+++ b/internal/app/mmlogic/mmlogic_service.go
@@ -30,6 +30,15 @@ var (
 	})
 )
 
+const (
+	// Minimum number of tickets to be returned in a streamed response for QueryTickets. This value
+	// will be used if page size if not configured or is configured lower than the minimum value.
+	minPageSize int = 10
+	// Maximum number of tickets to be returned in a streamed response for QueryTickets. This value
+	// will be used if page size is configured higher than the minimum value.
+	maxPageSize int = 1000
+)
+
 // The MMLogic API provides utility functions for common MMF functionality such
 // as retreiving Tickets from state storage.
 type mmlogicService struct {
@@ -69,6 +78,15 @@ func (s *mmlogicService) QueryTickets(req *pb.QueryTicketsRequest, responseServe
 	}
 
 	pSize := s.cfg.GetInt("storage.page.size")
+	if pSize <= 0 {
+		logger.Warningf("page size %v is lower than minimum limit of %v, setting page size as %v", pSize, minPageSize, minPageSize)
+		pSize = minPageSize
+	}
+
+	if pSize > 1000 {
+		logger.Warningf("page size %v is higher than maximum limit of %v, setting page size as %v", pSize, maxPageSize, maxPageSize)
+		pSize = maxPageSize
+	}
 
 	// Send requests to the storage service
 	err := s.store.FilterTickets(ctx, poolFilters, pSize, callback)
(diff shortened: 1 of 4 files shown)


internal/app/mmlogic/mmlogic_service.go
```diff
@@ -69,6 +78,15 @@ func (s *mmlogicService) QueryTickets(req *pb.QueryTicketsRequest, responseServe
 	}
 
 	pSize := s.cfg.GetInt("storage.page.size")
+	if pSize <= 0 {
```
Laremere: nit: pSize <  minPageSize

pSize > maxPageSize

internal/statestore/redis.go
```diff
@@ -440,7 +440,7 @@ func (rb *redisBackend) FilterTickets(ctx context.Context, filters []*pb.Filter,
 }
 
 func idsToPages(ids []string, pageSize int) [][]interface{} {
-	result := make([][]interface{}, 0, len(ids)/pageSize+1)
+	result := make([][]interface{}, 0, len(ids)/(pageSize+1))
```
Laremere: The previous logic was correct:
Eg, if page size is 10, and there are 15 ids:
(15/10) + 1 = 2

While 15/(10+1) = 1

We are expecting to return two pages, so 2 is the correct answer.  This logic maybe over allocates by 1, but the maximum capacity is roughly known.  Logic works either way, as append will create a new backing array if going over capacity.

Laremere: It's an error with the calling code if it tries to do page size = 0.  Maybe there should be a panic with a better error message, but failing on pagesize=0 seems right to me.

internal/util/netlistener/listener_holder.go
```diff
@@ -72,6 +72,9 @@ func (lh *ListenerHolder) Close() error {
 func NewFromPortNumber(portNumber int) (*ListenerHolder, error) {
 	addr := fmt.Sprintf(":%d", portNumber)
 	conn, err := net.Listen("tcp", addr)
+	if err != nil {
```
Laremere: This check collides with the if err == nil check on line 84.

internal/harness/golang/matchfunction_service.go
```diff
@@ -146,11 +146,9 @@ func (s *matchFunctionService) getMatchManifest(ctx context.Context, req *pb.Run
 
 // TODO: replace this method once the client side wrapper is done.
 func getMMLogicClient(cfg config.View) (pb.MmLogicClient, error) {
+	// Retrieve host name and port. It is ok to not specify host name as there are scenarios
```
Laremere: Should it maybe be set to localhost in that scenario?

sawagh: I think it is ok to not force a null check here - if someone mistakenly does not specify, it fails with connection not being established.

internal/util/netlistener/listener_holder.go
sawagh: I ve removed that check - This one is required since there is a null dereference panic otherwise on 79.

internal/app/mmlogic/mmlogic_service.go
sawagh: Good catch - fixed.

internal/statestore/redis.go
sawagh: Ah - This was originally my hack to work past 0 page size. You are right - this is not needed - and I think failing due to page size 0 here is a code bug - so it is OK to panic. Reverted!

PR #438, opened 2019-05-23
Have sample MMF create simple 1v1 matches

Laremere:
The previous MMF required a weird behavior where you needed to set tickets in rosters 
to be overridden.  It would also break if there weren't enough tickets to fill the rosters.  This
is a simpler example which takes tickets from the pool and assigns them into 1v1 matches.


examples/functions/golang/simple/main.go
```diff
@@ -41,34 +41,39 @@ func main() {
...
-	// 3. Create one single match candidate and return it.
+	// 1. Deduplicates the tickets from the pools into a single list.
+	// 2. Groups players into 1v1 matches.
 
-	allTickets := make([]*pb.Ticket, 0)
-	ticketOffset := 0
-	for _, tickets := range view.PoolNameToTickets {
-		allTickets = append(allTickets, tickets...)
-	}
+	tickets := make(map[string]*pb.Ticket)
```
jeremyje: tickets := map[string]*pb.Ticket{}

examples/functions/golang/simple/main.go
```diff
@@ -41,34 +41,39 @@ func main() {
 // makeMatches is where your custom matchmaking logic lives.
 func makeMatches(view *goHarness.MatchFunctionParams) []*pb.Match {
```
jeremyje: `view` is a strange variable name for this.

PR #448, opened 2019-05-29
Modify FunctionConfig to indicate mmf server type

### LINKED ISSUE #446: pb.FunctionConfig cannot describe mmf secureness
Current `pb.FunctionConfig` definition only contains host and port information but lack of the ability to describe if we are using an insecure/secure mmf. We may need to add a boolean field to describe the type of mmf server that we are currently using.

yfei1:
Resolves #446


api/backend.proto
```diff
@@ -58,12 +58,14 @@ option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
...
   string host = 1;
   int32 port = 2;
+  bool useSecure = 3;
 }
 
 // Configuration for a REST Match Function.
 message RestFunctionConfig {
   string host = 1;
   int32 port = 2;
+  bool useSecure = 3;
```
jeremyje: Would it make sense to just hinge this off of `http://` vs `https://`

api/backend.proto
```diff
@@ -58,12 +58,14 @@ option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
 message GrpcFunctionConfig {
   string host = 1;
   int32 port = 2;
+  bool useSecure = 3;
```
jeremyje: Should this be the CA cert that the client should trust rather than say I'm turning on TLS.

internal/app/backend/backend_service.go
```diff
@@ -142,6 +179,15 @@ func (s *backendService) getGRPCClient(config *pb.FunctionConfig_Grpc) (pb.Match
 	return client.(pb.MatchFunctionClient), nil
 }
 
+func (s *backendService) getHTTPClient(config *pb.FunctionConfig_Rest) (*http.Client, string, error) {
```
golangcibot: U1000: func `(*backendService).getHTTPClient` is unused (from `unused`)

internal/app/backend/backend_service.go
```diff
@@ -50,6 +61,32 @@ func newBackend(cfg config.View) (*backendService, error) {
 		cfg: cfg,
 	}
 
+	if cfg.GetBool("tls.enabled") {
```
jeremyje: Move this code to rpc/ with signature like. `func CreateHTTPClient(cfg config.View, address string) (http.Client, error)` and `func CreateGRPCClient(cfg config.View, address string) (grpc.ClientConn, error)`.

That will prevent the need from exposing the TLS utils.

internal/rpc/clients.go
```diff
@@ -33,6 +35,8 @@ var (
 		"app":       "openmatch",
 		"component": "client",
 	})
+	pool *x509.CertPool
```
golangcibot: `pool` is a global variable (from `gochecknoglobals`)

internal/rpc/clients.go
```diff
@@ -33,6 +35,8 @@ var (
...
+	tc   credentials.TransportCredentials
```
golangcibot: `tc` is a global variable (from `gochecknoglobals`)

internal/rpc/clients.go
jeremyje: This is a legit lint finding. Having a global pool is bad for a few reasons.
Unit tests are now going to read global state so this will be a source of flakes and race conditions.

I'd much prefer reading the certs each time and caching the objects. We can improve that later. If we have a global that's much harder to change.

internal/rpc/clients.go
```diff
@@ -71,6 +72,40 @@ func GRPCClientFromConfig(cfg config.View, prefix string) (*grpc.ClientConn, err
 	return GRPCClientFromParams(clientParams)
 }
 
+// GRPCClientFromEndpoint creates a gRPC client connection from endpoint.
+func GRPCClientFromEndpoint(cfg config.View, hostname string, port int) (*grpc.ClientConn, error) {
```
jeremyje: I'm going to approve but I think this code is doing too much.
Add a TODO to investigate keep a cache of the `*tls.Config`.

PR #452, opened 2019-05-29
Reuse existing helper functions to grab GRPC clients

yfei1:


test/e2e/minimatch/minimatch_test.go
```diff
@@ -39,7 +40,7 @@ const (
 )
 
 func TestMinimatchStartup(t *testing.T) {
-	assert := require.New(t)
+	assert := assert.New(t)
```
jeremyje: This code now relies on the fact that all the endpoints are served from the same server. This may not be true if we decide to point to a real Open Match cluster for our e2e tests.

What do you think? Is this abstraction worth removing?

yfei1: I think e2e tests are used to test whether the flow of an application from the start to finish is behaving as expected. Is it an overkill to switch to a real Open Match cluster if the current tests already proved it works?

sawagh: Why are we changing back from require to assert here? This changes the assert behavior through the test - and the reason for using require was that there are points you want the test to fail if some asserts fail - than proceeding and panic'ing. 

We can possibly use both assert, require and use them as needed through the test - but switching back to assert from require will cause test panics in certain failure cases.

yfei1: > Why are we changing back from require to assert here? This changes the assert behavior through the test - and the reason for using require was that there are points you want the test to fail if some asserts fail - than proceeding and panic'ing.
> 
> We can possibly use both assert, require and use them as needed through the test - but switching back to assert from require will cause test panics in certain failure cases.

I'm not intended to change back from require to assert. Just wanna make sure we are using the same assert library such that our tests have the same behavior whenever it failed. I this assert and require doesnt have that much difference as long as we handle the test resources properly.

sawagh: 'assert' and 'require' are different packages from the same project: github.com/stretchr/testify. They expose similar methods with very different behavior - require fails execution of the test if the condition fails, assert continues. They both have their intended usage - and we should be using them both as appropriate. I do not see a reason to pick or standardize on one of them. We can use either of them (or both) as needed by the test.

PR #453, opened 2019-05-29
Move to Helm 3-alpha-1.

jeremyje:
Helm removes the need to install Tiller into kube-system namespace which simplifies the use of Helm. It's still pretty early for Helm 3 but the improvement to our workflow is worth it.

Helm 2 issues:
 * Helm can't be deleted cleanly
 * Helm push occasionally doesn't wait long enough for tiller to be ready.

Helm 3 issues:
 * `helm upgrade --wait` doesn't work, https://github.com/helm/helm/issues/5814


bacongobbler:
I'd highly recommend against moving to pre-production software, especially while things are in flux. Releases created by Helm 2 and Helm 3 are incompatible with each other for example, and there's no migration path available at this time.

jeremyje:
> I'd highly recommend against moving to pre-production software for use in production, especially while things are in flux. Releases created by Helm 2 and Helm 3 are incompatible with each other for example, things are expected to break over time and there's no migration path available at this time.

Open Match is a pre-production product that's also going through a major rewrite at this time. We mainly use helm for creating the Kubernetes YAML files and deployments for development. Given this and at first look it seems Helm 3 has less issues than Helm 2 it makes sense for us to move over.

Issue #442, opened 2019-05-25
Add support to call an MMF implemented as a REST endpoint. 


sawagh:
Currently we only support calling a GRPC MMF. We need to support calling a REST MMF endpoint. This would involve implementing the logic here to support calling the function type:
FunctionConfig_Rest

https://github.com/GoogleCloudPlatform/open-match/blob/master/internal/app/backend/backend_service.go#L92

Given that our harness today already exposes an HTTP proxy endpoint, we may be able to leverage the existing sample mmf in the E2E test to trigger the same MMF as a HTTP implemented method. However, you may need to validate this / come up with alternate test strategy.

PR #458, opened 2019-05-30
Reorganize backend service and e2e test for incoming rest config logic

### LINKED ISSUE #450: e2e test silently ignores all mmf failures
#436 returns an EOF error even if we have a bad mmf client. In the worst case, our [e2e test](https://github.com/GoogleCloudPlatform/open-match/blob/master/test/e2e/minimatch/minimatch_test.go#L194-L211) will mark itself as passed without testing the behavior of `FetchMatches`.

yfei1:
This commit provides the basic code structure to implement REST mmf support #442 
Resolves #450


internal/app/backend/backend_service.go
```diff
@@ -71,38 +71,65 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
 	case *pb.FunctionConfig_Rest:
+		// httpClient, baseURL, err := s.getHTTPClient((req.Config.Type).(*pb.FunctionConfig_Rest))
 		return status.Error(codes.Unimplemented, "not implemented")
-
 	default:
 		logger.Error("unsupported function type provided")
 		return status.Error(codes.InvalidArgument, "provided match function type is not supported")
 	}
 
+	ctx := stream.Context()
```
jeremyje: ctx := stream.Context()

internal/app/backend/backend_service.go
```diff
@@ -71,38 +71,65 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
+					// Get the channel over which the generated match results will be sent.
+					s.matchesFromGrpcMMF(ctx, profile, grpcClient, matchChan, errChan)
+				case *pb.FunctionConfig_Rest:
+					// TODO: implement matchesFromHttpMMF function
+					// s.matchesFromHttpMMF(ctx, profile, httpClient, baseURL, matchChan, errChan)
+				}
+			}(profile)
+		}
+		wg.Wait()
+		close(matchChan)
```
jeremyje: defer func() {
  wg.Wait()
  close(matchChan)
  close(errChan)
}()
At the top of this func. Otherwise there's a decent chance of leaking these resources.

Also keep the wg.Wait() here as well.

internal/app/backend/backend_service.go
```diff
@@ -71,38 +71,65 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
+			go func(profile *pb.MatchProfile) {
+				defer wg.Done()
+
+				switch (req.Config.Type).(type) {
+				case *pb.FunctionConfig_Grpc:
+					// Get the channel over which the generated match results will be sent.
+					s.matchesFromGrpcMMF(ctx, profile, grpcClient, matchChan, errChan)
+				case *pb.FunctionConfig_Rest:
+					// TODO: implement matchesFromHttpMMF function
+					// s.matchesFromHttpMMF(ctx, profile, httpClient, baseURL, matchChan, errChan)
```
jeremyje: Emit unimplemented error here.

internal/app/backend/backend_service.go
```diff
@@ -145,53 +172,32 @@ func (s *backendService) getGRPCClient(config *pb.FunctionConfig_Grpc) (pb.Match
...
-					c <- match
-				}
-			}(profile)
-		}
-
-		// Wait for all the goroutines handling each profile to return and then close the channel.
-		wg.Wait()
-		close(c)
-	}(c)
+func (s *backendService) matchesFromGrpcMMF(ctx context.Context, profile *pb.MatchProfile, client pb.MatchFunctionClient, matchChan chan<- *pb.Match, errChan chan<- error) {
```
jeremyje: nit: Seems like this method is standalone and doesn't need to be a method of this class.

PR #459, opened 2019-05-30
Implemented backend rest config logic

### LINKED ISSUE #442: Add support to call an MMF implemented as a REST endpoint.

yfei1:
This commit along with #458 resolves #442


internal/app/backend/backend_service.go
```diff
@@ -142,6 +147,57 @@ func (s *backendService) getGRPCClient(config *pb.FunctionConfig_Grpc) (pb.Match
 	return client.(pb.MatchFunctionClient), nil
 }
 
+func (s *backendService) matchesFromHttpMMF(ctx context.Context, profile *pb.MatchProfile, client *http.Client, baseURL string, matchChan chan<- *pb.Match, errChan chan<- error) {
```
golangcibot: `(*backendService).matchesFromHttpMMF` - `ctx` is unused (from `unparam`)

internal/app/backend/backend_service.go
```diff
@@ -142,6 +147,57 @@ func (s *backendService) getGRPCClient(config *pb.FunctionConfig_Grpc) (pb.Match
...
+		return
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		errChan <- err
+		logger.WithError(err).Error("failed to get response from mmf run")
+		return
+	}
+	defer resp.Body.Close()
```
golangcibot: Error return value of `resp.Body.Close` is not checked (from `errcheck`)

internal/app/backend/backend_service.go
```diff
@@ -142,6 +147,62 @@ func (s *backendService) getGRPCClient(config *pb.FunctionConfig_Grpc) (pb.Match
 	return client.(pb.MatchFunctionClient), nil
 }
 
+func (s *backendService) matchesFromHttpMMF(profile *pb.MatchProfile, client *http.Client, baseURL string, matchChan chan<- *pb.Match, errChan chan<- error) {
```
golangcibot: method matchesFromHttpMMF should be matchesFromHTTPMMF (from `golint`)

internal/app/backend/backend_service.go
```diff
@@ -142,10 +147,66 @@ func (s *backendService) getGRPCClient(config *pb.FunctionConfig_Grpc) (pb.Match
 	return client.(pb.MatchFunctionClient), nil
 }
 
+func (s *backendService) matchesFromHTTPMMF(profile *pb.MatchProfile, client *http.Client, baseURL string, matchChan chan<- *pb.Match, errChan chan<- error) {
```
golangcibot: `profile` can be `open-match.dev/open-match/vendor/github.com/gogo/protobuf/proto.Message` (from `interfacer`)

internal/app/backend/backend_service.go
```diff
@@ -142,10 +147,67 @@ func (s *backendService) getGRPCClient(config *pb.FunctionConfig_Grpc) (pb.Match
...
+		return
+	}
+
+	req, err := http.NewRequest("post", baseURL+"/v1/matchfunction:run", bytes.NewBuffer(reqBody))
+	if err != nil {
+		errChan <- err
+		logger.WithError(err).Error("failed to create mmf http request")
+		return
+	}
+
```
jeremyje: req.WithContext(ctx)

This should ensure that tracing works.

internal/app/backend/backend_service.go
```diff
@@ -142,10 +147,67 @@ func (s *backendService) getGRPCClient(config *pb.FunctionConfig_Grpc) (pb.Match
...
+	resp, err := client.Do(req)
+	if err != nil {
+		errChan <- err
```
jeremyje: We should generate grpc errors here instead since we have enough context here of what probably happened.
For example in this case the error is retryable since it's probably because TCP connection broke.

internal/app/backend/backend_service.go
```diff
@@ -142,10 +147,67 @@ func (s *backendService) getGRPCClient(config *pb.FunctionConfig_Grpc) (pb.Match
...
+		logger.WithError(err).Error("failed to get response from mmf run")
```
jeremyje: There's 2 issues. I'm not sure if these log statements are useful here because they get logged on the RPC responses. Also the amount of information from these log entries does not provide enough information.
Who did I call?
Is there any payload information that would be relevant to the error as well?

internal/app/backend/backend_service.go
```diff
@@ -142,10 +147,67 @@ func (s *backendService) getGRPCClient(config *pb.FunctionConfig_Grpc) (pb.Match
...
+		errChan <- err
+		logger.WithError(err).Error("failed to read from response body")
+		return
+	}
+
+	pbResp := &pb.RunResponse{}
+	err = proto.Unmarshal(body, pbResp)
+	if err != nil {
+		errChan <- err
+		logger.WithError(err).Error("failed to unmarshal response body to response pb")
```
jeremyje: This would be `codes.FailedPrecondition`.

internal/app/backend/backend_service.go
```diff
@@ -142,10 +147,66 @@ func (s *backendService) getGRPCClient(config *pb.FunctionConfig_Grpc) (pb.Match
 	return client.(pb.MatchFunctionClient), nil
 }
 
+// nolint: unused
+func matchesFromHTTPMMF(ctx context.Context, profile *pb.MatchProfile, client *http.Client, baseURL string, matchChan chan<- *pb.Match, errChan chan<- error) {
```
golangcibot: `matchesFromHTTPMMF` is unused (from `deadcode`)

PR #460, opened 2019-05-30
Create third party folder with grpc-gateway *.proto dependencies

### LINKED ISSUE #454: Create third_party folder and move build/toolchain/include/* artifacts there.
We currently download master of https://github.com/googleapis/googleapis and https://github.com/grpc-ecosystem/grpc-gateway via `make build/toolchain/include/protoc-gen-swagger/` and `make build/toolchain/include/google/api/`.

It's more appropriate to create a third_party/ directory and store the files there. This way the builds are hermetic. We can consider using git submodules but that adds some complexity to our workflows.

@sawagh @Laremere @yfei1 I think this needs a discussion on how to do this. I think the 2 candidates are:
* Use git submodules
* Create some automation to download and apply the source code for each repository to our project.

Some things to consider
* https://github.com/grpc-ecosystem/grpc-gateway only requires a subdirectory, `protoc-gen-swagger/options` and we only need [*.proto files](https://github.com/GoogleCloudPlatform/open-match/blob/master/Makefile#L503). The other files may complicate the Makefile because it could break `go test ./...` and friends because it'll be inclined to test another repository that's not in it's go path. They have opted to manually copy over github.com/googleapis/googleapis, https://github.com/grpc-ecosystem/grpc-gateway/tree/master/third_party/googleapis.
* grpc (non-go) repository uses submodules, https://github.com/grpc/grpc/tree/master/third_party. It looks like it's influenced by bazel which I'd say we probably don't want a too weird of a structure to break users that wish to use Open Match via, https://github.com/bazelbuild/rules_go. We don't need a WORKSPACE file to support these users.

yfei1:
Resolves #454


jeremyje:
The bug needs some discussion before we decide which route to go. If any case this change needs automation so that we don't have to think about how to add the code from external repositories here.

Makefile
```diff
@@ -892,6 +872,27 @@ update-deps:
...
+	mkdir -p $(TOOLCHAIN_DIR)/googleapis-temp/
+	curl -o $(TOOLCHAIN_DIR)/googleapis-temp/googleapis.zip -L https://github.com/googleapis/googleapis/archive/master.zip
+	(cd $(TOOLCHAIN_DIR)/googleapis-temp/; unzip -q -o googleapis.zip)
+	cp -f $(TOOLCHAIN_DIR)/googleapis-temp/googleapis-master/google/api/annotations.proto \
+		$(TOOLCHAIN_DIR)/googleapis-temp/googleapis-master/google/api/http.proto \
+		$(TOOLCHAIN_DIR)/googleapis-temp/googleapis-master/google/api/httpbody.proto \
+		$(REPOSITORY_ROOT)/third_party/google/api/
+	rm -rf $(TOOLCHAIN_DIR)/googleapis-temp
+
+update-protoc-gen-swagger-options:
```
jeremyje: `third_party/protoc-gen-swagger/options:`

Fix the others as well.

PR #463, opened 2019-05-30
Implements frontend and backend Assignments method with tests

### LINKED ISSUE #441: Set / Get Assignment support to Open Match v0.6
The v0.6 Open Match implementation currently does not have the Assignment Support implemented. At a high level, the following needs to be added to support this flow: 

On the Backend, implement: func (s *backendService) AssignTickets
This would enable setting assignments for the backend API

On the Frontend, Implement: func (s *frontendService) GetAssignments
This would enable fetching assignments from the frontend API

This would need the ability to set assignment on the State Storage. too. 

To validate this, adding tests for state storage, frontend API (component test) and an E2E test that validates assignment being returned through the streaming frontend API when it gets set by the Backend API - might be a good idea.

yfei1:
Resolves #441


internal/app/backend/backend_service.go
```diff
@@ -197,5 +197,11 @@ func (s *backendService) matchesFromGrpcMMF(ctx context.Context, profiles []*pb.
 // AssignTickets sets the specified Assignment on the Tickets for the Ticket
 // ids passed.
 func (s *backendService) AssignTickets(ctx context.Context, req *pb.AssignTicketsRequest) (*pb.AssignTicketsResponse, error) {
+	err := s.store.UpdateAssignments(ctx, req.TicketId, req.Assignment)
+	if err != nil {
+		logger.WithError(err).Error("failed to update assignments for requested tickets")
```
jeremyje: We are better off creating an interceptor to log errors in grpc than do this.

internal/statestore/redis.go
```diff
@@ -439,6 +439,63 @@ func (rb *redisBackend) FilterTickets(ctx context.Context, filters []*pb.Filter,
...
+			err = redisConn.Send("MULTI")
+			if err != nil {
+				redisLogger.WithError(err).Error("failed to open a multi transaction block in redis")
+				return status.Errorf(codes.Internal, "%v", err)
+			}
+
+			// Store assignment data by fields
+			err = redisConn.Send("HSET", connectionTable, id, assignment.Connection)
+			if err != nil {
+				redisLogger.WithError(err).Errorf("failed to hset assignment connection for ticket %#v", id)
```
jeremyje: Error should only be emitted like this if it's a serious internal error. Like something is misconfigured. A bad query shouldn't write an error to the logs.

Issue #169, opened 2019-04-09
Add Deployment Health Check for CI


jeremyje:
CI for Open Match is currently limited to building and unit testing. It does not verify the correctness of the Helm charts are install/yaml files. We should add a build step that deploys Open Match to a Kubernetes cluster and verifies the healthiness of the cluster.

yfei1:
I can take this. Any specific tools or resources I should look into?

sawagh:
I am

jeremyje:
This was pretty much done by 
https://github.com/googleforgames/open-match/pull/537
https://github.com/googleforgames/open-match/pull/536

PR #469, opened 2019-05-31
Add Kubernetes health checks. (#169)

jeremyje:


internal/rpc/tls_server.go
```diff
@@ -119,11 +119,8 @@ func (s *tlsServer) start(params *ServerParams) (func(), error) {
 	}
 
 	// Bind HTTPS handlers
-	s.httpMux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
-		fmt.Fprintf(w, "ok")
-	})
+	s.httpMux.HandleFunc("/healthz", monitoring.NewHealthProbe(params.handlersForHealthCheck))
 	s.httpMux.Handle("/", s.proxyMux)
-	zpages.Handle(s.httpMux, "/debug")
```
yfei1: why are we removing zpage?

internal/rpc/insecure.go
```diff
@@ -87,11 +86,8 @@ func (s *insecureServer) start(params *ServerParams) (func(), error) {
 		}
 	}
 
-	s.httpMux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
-		fmt.Fprintf(w, "ok")
-	})
+	s.httpMux.HandleFunc("/healthz", monitoring.NewHealthProbe(params.handlersForHealthCheck))
 	s.httpMux.Handle("/", s.proxyMux)
-	zpages.Handle(s.httpMux, "/debug")
```
yfei1: why are we removing zpage?

internal/statestore/redis.go
```diff
@@ -82,22 +82,26 @@ func NewRedis(cfg config.View, redisURL string, maskedURL string) (Service, erro
...
-
-	redisLogger.Info("Connected to Redis")
 	return &redisBackend{
 		redisPool: pool,
 		cfg:       cfg,
 	}, nil
 }
 
+// HealthCheck indicates if the database is reachable.
+func (rb *redisBackend) HealthCheck(ctx context.Context) error {
```
yfei1: Can you add the timeout settings to https://github.com/GoogleCloudPlatform/open-match/blob/master/internal/statestore/redis.go#L76? Otherwise, the healthcheck function will hang if redis is not ready yet.

```
		Dial: func() (redis.Conn, error) {
			return redis.DialURL(
				redisURL,
				redis.DialConnectTimeout(5*time.Second),
				redis.DialReadTimeout(3*time.Second),
			)
		},
```

internal/rpc/tls_server.go
jeremyje: `monitoring.bindZpages` already calls this.

internal/rpc/insecure.go
jeremyje: `monitoring.bindZpages` already calls this.

PR #475, opened 2019-06-01
Enable golangci in presubmit

yfei1:
https://github.com/golangci/golangci-lint/issues/500 is fixed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,6 +40,5 @@ require (
 	github.com/stretchr/testify v1.3.0
 	github.com/yuin/gopher-lua v0.0.0-20190514113301-1cd887cd7036 // indirect
 	go.opencensus.io v0.21.0
-	google.golang.org/genproto v0.0.0-20190404172233-64821d5d2107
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -40,6 +40,5 @@ require (
 	github.com/stretchr/testify v1.3.0
```
Laremere: this seems like an unrelated change?

internal/app/backend/backend_service.go
```diff
@@ -42,8 +42,7 @@ type backendService struct {
 }
 
 type mmfClients struct {
-	addrMap    sync.Map
```
Laremere: this seems like an unrelated change?

go.mod
yfei1: This is done by the `go mod tidy` command in `make presubmit` automatically.

internal/app/backend/backend_service.go
Laremere: Per discussion: golangci complains that this is unused.  Therefor removing.

PR #476, opened 2019-06-03
Make fake state store only returns the closer

yfei1:


internal/app/backend/backend_test.go
```diff
@@ -17,17 +17,14 @@ package backend
...
 func TestServerBinding(t *testing.T) {
-	bs := func(p *rpc.ServerParams) {
-		p.AddHandleFunc(func(s *grpc.Server) {
-			pb.RegisterBackendServer(s, &backendService{})
-		}, pb.RegisterBackendHandlerFromEndpoint)
+	bs := func(p *rpc.ServerParams, cfg config.Mutable) {
+		BindService(p, cfg)
 	}
 
 	rpcTesting.TestServerBinding(t, bs)
```
Laremere: It seems you should be able to do rpcTesting.TestServerBinding(t, BindService)
Though BindService returns err, so it won't fit.  But that err isn't being testing when it should be?

internal/app/mmlogic/mmlogic_service_test.go
```diff
@@ -77,19 +76,12 @@ func queryTicketsLoop(t *testing.T, tc *rpcTesting.TestContext, req *pb.QueryTic
...
 
-		mmlogic := &mmlogicService{
-			cfg:   cfg,
-			store: store,
-		}
-		p.AddHandleFunc(func(s *grpc.Server) {
-			pb.RegisterMmLogicServer(s, mmlogic)
-		}, pb.RegisterMmLogicHandlerFromEndpoint)
+		cfg.Set("storage.page.size", 10)
+		BindService(p, cfg)
```
Laremere: not checking err.

internal/app/mmlogic/mmlogic_service_test.go
```diff
@@ -77,19 +76,12 @@ func queryTicketsLoop(t *testing.T, tc *rpcTesting.TestContext, req *pb.QueryTic
 
 func createMmlogicForTest(t *testing.T) *rpcTesting.TestContext {
 	var closerFunc func()
-	tc := rpcTesting.MustServe(t, func(p *rpc.ServerParams) {
-		cfg := viper.New()
-		store, closer := statestoreTesting.New(t, cfg)
+	tc := rpcTesting.MustServe(t, func(p *rpc.ServerParams, cfg config.Mutable) {
+		closer := statestoreTesting.New(t, cfg)
```
Laremere: just assign closerFunc directly?

internal/rpc/testing/helpers.go
```diff
@@ -24,31 +24,35 @@ import (
...
+	"open-match.dev/open-match/internal/config"
 	"open-match.dev/open-match/internal/rpc"
 	netlistenerTesting "open-match.dev/open-match/internal/util/netlistener/testing"
 	certgenTesting "open-match.dev/open-match/tools/certgen/testing"
 )
 
 // MustServe creates a test server and returns TestContext that can be used to create clients.
 // This method pseudorandomly selects insecure and TLS mode to ensure both paths work.
-func MustServe(t *testing.T, binder func(*rpc.ServerParams)) *TestContext {
+func MustServe(t *testing.T, bindAndConfig func(*rpc.ServerParams, config.Mutable)) *TestContext {
```
Laremere: It seems the bindAndConfig signature is used enough to define an interface.

internal/statestore/testing/fake_redis.go
```diff
@@ -35,13 +33,5 @@ func New(t *testing.T, cfg config.Mutable) (statestore.Service, func()) {
...
-	redisURL := fmt.Sprintf("redis://%s:%s", mredis.Host(), mredis.Port())
-
-	redis, err := statestore.NewRedis(cfg, redisURL, redisURL)
-	if err != nil {
-		t.Fatalf("error connecting to fake redis, %s, %v", redisURL, err)
-	}
-	return redis, func() {
-		mredis.Close()
-	}
+	return func() { mredis.Close() }
```
Laremere: return mredis.Close() ?

Laremere: If this returns an error, it should be logged?

internal/app/backend/backend_test.go
Laremere: Ah, I see, the config here is mutable, instead of just a view.
It seems that everything should be done in two steps, not one:
- Do all configuration
- Then do all binding.

Otherwise will result in possible bugs from ordering problems.  I see that miniredis doesn't like this as is, but looking at the documentation, it is possible to tell mini redis the host and port to serve on: https://godoc.org/github.com/alicebob/miniredis#Miniredis.StartAddr

internal/app/evaluator/evaluator_test.go
```diff
@@ -17,17 +17,15 @@ package evaluator
...
 	rpcTesting "open-match.dev/open-match/internal/rpc/testing"
 )
 
 func TestServerBinding(t *testing.T) {
-	bs := func(p *rpc.ServerParams) {
-		p.AddHandleFunc(func(s *grpc.Server) {
-			pb.RegisterEvaluatorServer(s, &evaluatorService{})
-		}, pb.RegisterEvaluatorHandlerFromEndpoint)
+	bs := func(p *rpc.ServerParams, cfg config.Mutable) {
+		BindService(p, cfg, &FunctionSettings{Func: func([]*pb.Match) []*pb.Match { return nil }})
```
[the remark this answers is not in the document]
Laremere: Also, it's unfortunate that the evaluator doesn't have the same BindService signature as the other services.  I wonder if there's a way to fix that?

PR #489, opened 2019-06-04
Add Swagger UI to enable interactive calls.

jeremyje:
This change adds a small server that hosts Swagger UI and directs (via HTTP reverse proxy) to the HTTP reverse proxy on the appropriate server which talks to gRPC.

Only works in insecure mode but should be easy to add TLS support later.


Makefile
```diff
@@ -852,27 +873,27 @@ clean-stress-test-tools:
 clean: clean-images clean-binaries clean-site clean-release clean-build clean-protos clean-swagger-docs clean-nodejs clean-install-yaml clean-stress-test-tools clean-secrets
```
yfei1: Also need to modify clean-binaries and clean-images

jeremyje: This was there.

PR #494, opened 2019-06-05
Remove field FunctionName from harness.FunctionSetting

yfei1:
This commit aims to remove the field `FunctionName`from `harness.FunctionSetting` struct. The idea is every match function owns its name and should use that name to fill the RunResponse. There is no need to pass the name via the harness.


internal/harness/golang/harness.go
```diff
@@ -33,8 +33,7 @@ var (
 
 // FunctionSettings is a collection of parameters used to customize matchfunction views.
 type FunctionSettings struct {
-	FunctionName string
-	Func         matchFunction
```
sawagh: I see you removing this because it is not used. Do you think it may be useful to have the function name be added as a field to the harness logger / match function service logger? that way, any logs they emit can be tracked to which function they were running in context of?

yfei1: My two cents: if anything goes wrong inside the harness, then it would be a harness scope error and there is no need to do the log with the function name.

yfei1: My two cents: if anything goes wrong inside the harness, then it would be a harness scope error and there is no need to do the log with the function name.

PR #473, opened 2019-05-31
e2e test with random port

yfei1:
Resolves #456 
Temporarily disable TLS mode testing since open match backend service create a mmf client using `GRPCClientFromEndpoint`, which requires reading a trusted certificate from disk to enable secure mode.

PR #495, opened 2019-06-05
 Adds backend service tests

yfei1:
Part of the tests will move to e2e minimatch once #473 comes in


internal/app/backend/backend_service_test.go
```diff
@@ -0,0 +1,141 @@
...
+			break
+		}
+		handleResponse(resp, err)
+		if err != nil {
+			return
+		}
+	}
+}
+
+func assignTicketsLoop(t *testing.T, tc *rpcTesting.TestContext, req *pb.AssignTicketsRequest, handleResponse func(*pb.AssignTicketsResponse, error)) {
```
golangcibot: `assignTicketsLoop` - `t` is unused (from `unparam`)

internal/app/backend/backend_service_test.go
```diff
@@ -0,0 +1,141 @@
...
+			return
+		}
+	}
+}
+
+func assignTicketsLoop(tc *rpcTesting.TestContext, req *pb.AssignTicketsRequest, handleResponse func(*pb.AssignTicketsResponse, error)) {
+	c := pb.NewBackendClient(tc.MustGRPC())
+	resp, err := c.AssignTickets(tc.Context(), req)
+	handleResponse(resp, err)
+	if err != nil {
```
sawagh: redundant return?

internal/app/backend/backend_service_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+	fetchMatchesLoop(t, tc, &pb.FetchMatchesRequest{}, func(_ *pb.FetchMatchesResponse, err error) {
+		assert.Equal(codes.InvalidArgument, status.Convert(err).Code())
+	})
+}
+
+// TODO: Add FetchMatchesNormalTest when Hostname getter used to initialize mmf service is in
+// https://github.com/GoogleCloudPlatform/open-match/pull/473
+func fetchMatchesLoop(t *testing.T, tc *rpcTesting.TestContext, req *pb.FetchMatchesRequest, handleResponse func(*pb.FetchMatchesResponse, error)) {
+	c := pb.NewBackendClient(tc.MustGRPC())
```
sawagh: need to create this for every time this method is called? Can this be a client you create once and reuse through test?

internal/app/backend/backend_service_test.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+		handleResponse(resp, err)
+		if err != nil {
+			return
+		}
+	}
+}
+
+func assignTicketsLoop(tc *rpcTesting.TestContext, req *pb.AssignTicketsRequest, handleResponse func(*pb.AssignTicketsResponse, error)) {
+	c := pb.NewBackendClient(tc.MustGRPC())
```
sawagh: If you can reuse the client, then assignTicketsLoop seems overkill? about same lines of code calling this with the handleResponse than calling AssignTickets and handling the response.

internal/app/backend/backend_service_test.go
```diff
@@ -0,0 +1,141 @@
...
+	rpcTesting "open-match.dev/open-match/internal/rpc/testing"
+	statestoreTesting "open-match.dev/open-match/internal/statestore/testing"
+)
+
+func TestAssignTicketsEmptyRequest(t *testing.T) {
+	assert := assert.New(t)
+	tc := createBackendForTest(t)
+	defer tc.Close()
+
+	assignTicketsLoop(tc, &pb.AssignTicketsRequest{}, func(_ *pb.AssignTicketsResponse, err error) {
```
sawagh: This is a good example of where you can create a struct that has a input / response - and iterate over it and call the same assign as different test cases.

internal/app/backend/backend_service_test.go
```diff
@@ -0,0 +1,141 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"open-match.dev/open-match/internal/app/mmlogic"
+	"open-match.dev/open-match/internal/pb"
+	"open-match.dev/open-match/internal/rpc"
+
+	rpcTesting "open-match.dev/open-match/internal/rpc/testing"
+	statestoreTesting "open-match.dev/open-match/internal/statestore/testing"
+)
+
```
sawagh: You could have a single TestAssignTickets and have the normal, empty etc., all as different test cases within it.

internal/app/backend/backend_service_test.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	assignTicketsLoop(tc, req, func(resp *pb.AssignTicketsResponse, err error) {
+		assert.Equal(&pb.AssignTicketsResponse{}, resp)
+	})
+
+	gtResp, err := fe.GetTicket(tc.Context(), &pb.GetTicketRequest{TicketId: ctResp.Ticket.Id})
+	assert.Nil(err)
+	assert.Equal("localhost", gtResp.Assignment.Connection)
+}
+
```
sawagh: Same as above - TestFetchMatches as a test - and all test cases as table based tests within it.

internal/app/backend/backend_service_test.go
```diff
@@ -0,0 +1,141 @@
...
+func assignTicketsLoop(tc *rpcTesting.TestContext, req *pb.AssignTicketsRequest, handleResponse func(*pb.AssignTicketsResponse, error)) {
+	c := pb.NewBackendClient(tc.MustGRPC())
+	resp, err := c.AssignTickets(tc.Context(), req)
+	handleResponse(resp, err)
+	if err != nil {
+		return
+	}
+}
+
```
sawagh: This test coverage is great - Lets discuss offline where it must live though. If this test relies on frontend and mmlogic to be up, then should it be a part of E2E?

internal/statestore/redis.go
```diff
@@ -456,6 +456,10 @@ func (rb *redisBackend) FilterTickets(ctx context.Context, filters []*pb.Filter,
 
 // UpdateAssignments update the match assignments for the input ticket ids
 func (rb *redisBackend) UpdateAssignments(ctx context.Context, ids []string, assignment *pb.Assignment) error {
+	if assignment == nil || (assignment.Connection == "" && assignment.Error == "" && assignment.Properties == "") {
```
sawagh: Should we care about this?

yfei1: I think we do need the nil check, otherwise calling this method with nil assignment will panic. What do you think for the content checks?

internal/app/backend/backend_service_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"google.golang.org/grpc/status"
+	"open-match.dev/open-match/internal/app/mmlogic"
+	"open-match.dev/open-match/internal/pb"
+	"open-match.dev/open-match/internal/rpc"
+
+	rpcTesting "open-match.dev/open-match/internal/rpc/testing"
+	statestoreTesting "open-match.dev/open-match/internal/statestore/testing"
+)
+
+func TestAssignTicketsRequest(t *testing.T) {
```
sawagh: nit: rename to TestAssignTickets, TestFetchMatches short n simple

PR #496, opened 2019-06-05
Adds bazel protobuf rules for API

cgrinker:
This PR adds `proto_library` rules for the API. These are useful for generating frontend/backend clients as well as implementing custom Match Function/Evaluator endpoints.

### GRPC Gateway
The `proto_library` rules are dependent on the [grpc-ecosystem/grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway). As discussed here [Here](https://github.com/grpc-ecosystem/grpc-gateway/pull/941) many projects are vendoring [googleapis/googleapis](https://github.com/googleapis/googleapis)'s annotation definition file. This dependency could be edited to point at the googleapis repo once it becomes stable enough.

### Gazelle 
I opted to avoid including any [Gazelle](https://github.com/bazelbuild/bazel-gazelle) as there are [outstanding issues](https://github.com/bazelbuild/bazel-gazelle/issues/416) regarding usage on Windows.


Laremere:
Hey, thanks for showing interest in open match!

Could you explain why this change is necessary?  Are you trying to integrate open match with a personal project that uses Bazel?

While we want to support developers using open match, there are a couple of concerns:

- We want to prevent proliferation of build system files.  Ideally there is one blessed way to build things within open match.
- How do we guarantee that this code is maintained?  eg, version upgrades happen, builds work properly.  There are no tests here to ensure things don't break.

We want to keep the project simple, and prevents changes becoming bogged down with keeping multiple build systems happy. We don't want to take on the maintenance cost of more build systems unless it's really justified.

Thanks!

WORKSPACE
```diff
@@ -0,0 +1,28 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
jeremyje: Seems like a lot of go_repository entries are missing. Is that no longer required by rules_go?

.gitignore
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+bazel*
```
jeremyje: This would ignore all the BUILD.bazel files.

api/BUILD
```diff
@@ -0,0 +1,63 @@
+proto_library(
+    visibility = ["//visibility:public"],
```
jeremyje: Use default_visibility. The api/ directory is meant to be public.

jeremyje:
I hate to do this but I don't think we can accept this PR right now. My comment above should hopefully get you unblocked for using Open Match in bazel. If there's any changes that make this code difficult to use within bazel via third_party BUILD rules. Let us know we hopefully should be able to accommodate for that.

PR #498, opened 2019-06-06
Update the behavior of backend.FetchMatches

yfei1:
Per discussion, `backend.FetchMatches` returns safely unless the context is canceled. FetchMatches moves to the next response candidate if it encounters any internal issues

### DIFF
--- a/internal/app/backend/backend_service.go
+++ b/internal/app/backend/backend_service.go
@@ -133,32 +133,38 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
 		}
 	}(resultChan)
 
-	var done bool
+	proposals := make(*pb.Match)
 
+	// Query if mmf encounters any errors in any of the profiles
 	for result := range resultChan {
 		select {
 		case <-ctx.Done():
-			done = true
 			continue
 		default:
 			// Log and continue with the results from next profile if the current ones fail
-			if result.err != nil {
-				logger.Error(result.err.Error())
+			if err != nil {
+				continue
 			}
-			for _, match := range result.matches {
-				err = stream.Send(&pb.FetchMatchesResponse{Match: match})
-				if err != nil {
-					logger.WithError(err).Error("failed to stream back the response")
-				}
-			}
+
+			proposals = append(proposals, result.matches...)
+			err = result.err
 		}
 	}
 
-	if done {
-		return ctx.Err()
+	if err != nil {
+		logger.WithError(err).Error("failed to execute fetchmatches.")
+		return err
 	}
 
-	return nil
+	for _, proposal := range proposals {
+		err = stream.Send(&pb.FetchMatchesResponse{Match: match})
+		if err != nil {
+			logger.WithError(err).Error("failed to stream back the response")
+			return err
+		}
+	}
+
+	return ctx.Err()
 }
 
 func (s *backendService) getHTTPClient(config *pb.FunctionConfig_Rest) (*http.Client, string, error) {
--- a/internal/app/backend/backend_service.go
+++ b/internal/app/backend/backend_service.go
@@ -135,19 +135,15 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
 
 	proposals := []*pb.Match{}
 
-	// Query if mmf encounters any errors in any of the profiles
+	// Always drain out the channel to make sure we dont have any channel leaks
 	for result := range resultChan {
-		select {
-		case <-ctx.Done():
-			continue
-		default:
-			// Log and continue with the results from next profile if the current ones fail
-			if err != nil {
-				continue
-			}
-
-			proposals = append(proposals, result.matches...)
+		// Set the return error on the first erroneous result
+		if result.err != nil && err == nil {
 			err = result.err
+		}
+		// Only append if we have not encountered any errors yet
+		if err == nil {
+			proposals = append(proposals, result.matches...)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


internal/app/backend/backend_service.go
```diff
@@ -95,59 +101,64 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
 				}
+
+				resultChan <- mmfResult{matches, err}
 			}(profile)
 		}
-	}(matchChan, errChan)
+	}(resultChan)
 
-	for {
+	var done bool
```
Laremere: Remove this done logic, and just return ctx.Err()
It only returns non-nil if it was canceled, so it gets the same behavior without the bookkeeping.

internal/app/backend/backend_service.go
```diff
@@ -95,59 +101,64 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
-			return err
-		case match, ok := <-matchChan:
-			if !ok {
-				// Channel closed indicating that we are done processing matches.
-				return nil
+			done = true
+			continue
+		default:
+			// Log and continue with the results from next profile if the current ones fail
+			if result.err != nil {
```
Laremere: because of this logic, it could fail to run the MMF, but return no error, only logging it.  We should be returning an error in this case, right?  Note, we do want to avoid logging errors more than once per system.

internal/app/backend/backend_service.go
```diff
@@ -95,59 +101,70 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
 				}
+
+				resultChan <- mmfResult{matches, err}
 			}(profile)
 		}
-	}(matchChan, errChan)
+	}(resultChan)
 
-	for {
+	proposals := make(*pb.Match)
```
golangcibot: invalid argument: cannot make *pb.Match; type must be slice, map, or channel (from `typecheck`)

internal/app/backend/backend_service.go
```diff
@@ -95,59 +101,70 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
-			if err != nil {
-				return err
-			}
+	if err != nil {
+		logger.WithError(err).Error("failed to execute fetchmatches.")
+		return err
+	}
+
+	for _, proposal := range proposals {
+		err = stream.Send(&pb.FetchMatchesResponse{Match: match})
```
golangcibot: undeclared name: `match` (from `typecheck`)

internal/app/backend/backend_service.go
```diff
@@ -95,59 +101,70 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
-			err := stream.Send(&pb.FetchMatchesResponse{Match: m})
-			if err != nil {
-				return err
-			}
+	if err != nil {
+		logger.WithError(err).Error("failed to execute fetchmatches.")
+		return err
+	}
+
+	for _, proposal := range proposals {
```
golangcibot: ineffectual assignment to `proposal` (from `ineffassign`)

sawagh:
Had an offline conversation - the behavior we will settle for in v0.6 is - if any of the MMFs fail, the FetchMatches fails. FetchMatches will wait for all Profile MMF results before triggering the evaluator. The expectation from the user is - if there are a bunch of profiles that need to succeed or fail together (and be evaluated together) - club them in the same FetchMatches call. If they can fail, succeed or evaluate independently - split them across multiple FetchMatches call. We can take customer feedback on this simplified behavior to inform future changes.

internal/app/backend/backend_service.go
```diff
@@ -213,53 +223,35 @@ func matchesFromHTTPMMF(ctx context.Context, profile *pb.MatchProfile, client *h
...
-	// FetchMatches should fail or atleast hint that a part of MMF executions failed.
 	resp, err := client.Run(ctx, &pb.RunRequest{Profile: profile})
 	if err != nil {
-		logger.WithFields(logrus.Fields{
-			"error":   err.Error(),
-			"profile": profile,
-		}).Error("failed to run match function for profile")
-		errChan <- err
-		return
+		logger.WithError(err).Error("failed to run match function for profile")
```
Laremere: Why does this one, unlike the others, log the error with context,  then return error without context?

internal/app/backend/backend_service.go
```diff
@@ -95,59 +101,70 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
 		select {
 		case <-ctx.Done():
-			return ctx.Err()
-		case err := <-errChan:
-			return err
-		case match, ok := <-matchChan:
-			if !ok {
-				// Channel closed indicating that we are done processing matches.
-				return nil
+			continue
```
Laremere: This context done check does nothing.

internal/app/backend/backend_service.go
yfei1: This one makes a call to the service that we own, and we should expect the error being returned contains the status code with context.

internal/app/backend/backend_service.go
```diff
@@ -95,59 +101,53 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
-				}
-			}(profile)
-		}
-	}(matchChan, errChan)
+	proposals := []*pb.Match{}
 
-	for {
+	for i := 0; i < len(req.Profile); i++ {
 		select {
 		case <-ctx.Done():
```
Laremere: returning early here will deadlock goroutines which are trying to send results on resultChan.

PR #502, opened 2019-06-07
Move generated pb files from internal/pb to pkg/pb

### LINKED ISSUE #122: Move public API protos to pkg/pb/ directory.
As our APIs stabilize and we want to start exposing them to be consumed publically we'll need a directory that indicates "safe to depend on". Typically this is the pkg/ directory for golang applications as stated by [golang-standards/project-layout/pkg](https://github.com/golang-standards/project-layout/tree/master/pkg). The official golang website does not have an opinion on this but using a well understood pattern I think is a reasonable fallback here.

At this point it's unclear when we'll reach that state but we should at least discuss and maybe put a doc.go file there to reserve the package for those changes. At the very latest it should be done populated once we start pushing v1.0.0 release candidates.

Some examples of stuff that should go into pkg/.
 1. Publicly consumable protocol buffer definitions like frontend.proto.
 1. Useful libraries that can be used to extend Open Match.
 1. The Go implementation of the gRPC harness.

yfei1:
Resolves #122 

This commit moves the generated files of open-match frontend, backend, mmf, mmlogic, and evaluator to /pkg/pb directory. Frontend, backend, mmf are safe to expose them to be consumed publicly among open match's components. Mmlogic may support open-match's stat APIs in the near future so it should also be moved to /pkg/pb directory. The only thing I am unsure about is the evaluator. @sawagh what you do think?


sawagh:
Evaluator is definitely an internal only service - Not to be exposed publicly. It will likely be renamed to synchronizer so we an reserve 'Evaluator' for the user function.

yfei1:
The tricky thing is, if we want to move the generated files other than the evaluator ones to `pkg/pb` directory, we need a way to instruct `internal/pb/evaluator.pb.go` to use the proto definitions in `pkg/pb/messages.pb.go`, which requires changing [this line](https://github.com/googleforgames/open-match/blob/master/api/messages.proto) to `option go_package = "open-match.dev/open-match/pkg/pb";`. This will force us to use [absolute path](https://github.com/googleforgames/open-match/blob/8b07b5ecba0a92841bde7d944f427df603f8b575/Makefile#L645) in the Makefile when generating protos. Otherwise, the protoc compiler will generate an `open-match.dev/open-match` subdirectory inside the current workspace.

The con of using absolute path is, it requires users who want to build open-match locally having open-match setup in `path/to/open-match.dev/open-match` directory. Cloud build, for instance, will fail since it does not follow this setup. 

@sawagh @Laremere @jeremyje I made the change to cloudbuild and Dockerfile.ci accordingly in the latest commit, please let me know your thoughts on this change.

jeremyje:
Sync this PR.

Issue #88, opened 2019-02-26
Add end-to-end and scale testing support.


jeremyje:
Proposal:
https://docs.google.com/document/d/13ligVygwQ__1liFZdHpMLwGCXb35T-5QzQ6daf6r_JM/edit

jeremyje:
The document discusses the approach to get scale testing and the refactoring changes necessary to make it happen.

yfei1:
Closing this issue since most of the goals listed in the doc are resolved.

PR #503, opened 2019-06-07
Create cluster reaper for Open Match e2e tests. (#88)

jeremyje:
This is a command line tool to purge leaked clusters that will occur when there's CI test failures during cluster based end to end tests.

This command line tool may not exist as is when we need to actually start doing e2e testing. It might change to a Cloud Scheduler job.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -144,6 +144,11 @@ steps:
   - name: 'go-vol'
     path: '/go'
 
+- id: 'Test: Delete Old Clusters'
+  name: 'gcr.io/$PROJECT_ID/open-match-build'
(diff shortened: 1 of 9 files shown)


cloudbuild.yaml
```diff
@@ -144,6 +144,16 @@ steps:
   - name: 'go-vol'
     path: '/go'
 
+- id: 'View Makefile'
+  name: 'gcr.io/$PROJECT_ID/open-match-build'
+  args: ['cat', 'Makefile']
```
yfei1: Why do we need to this cat in the cloudbuild?
```
|\---/|
| o_o |
 \_^_/
```

Issue #509, opened 2019-06-10
Demo for Open Match v0.6


sawagh:
Issue tracks any changes for making an E2E demo for Open Match. This includes implementation and documentation of the demo steps.

jeremyje:
@Laremere Is this done?

Laremere:
Documentation isn't quite done.

jeremyje:
Since that's in a different repository there should be an issue in open-match-docs and this issue should be closed.

PR #510, opened 2019-06-10
Add bytesub for broadcasting demo state

Laremere:
This will be used by the demo.  The demo will have a central state, which will be updated, serialized to json, and then announced on an instance of ByteSub.  Demo webpage clients will subscribe using a websocket to receive the latest state.


examples/demo/bytesub/bytesub_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}()
+	go func() {
+		s.Subscribe(ctx, chanWriter{slow})
+		close(slow)
+	}()
+
+	for _, i := range []string{"0", "1", "2", "3"} {
+		s.AnnounceLatest([]byte(i))
+		if v := <-fast; v != i {
+			t.Errorf("Espected \"%s\", got \"%s\"", i, v)
```
jeremyje: Expected

examples/demo/bytesub/bytesub.go
```diff
@@ -0,0 +1,81 @@
...
+
+var partialWriteError = errors.New("ByteSub subscriber didn't consume the whole message.")
+
+// Subscribe writes the latest value in a single call to w.Write.  It does not
+// guarantee that all values published to AnnounceLatest will be written.
+// However once things catch up, it will write the latest value.  If no values
+// have been announced, waits for a value before writing.
+func (s *ByteSub) Subscribe(ctx context.Context, w io.Writer) error {
+	nextReady := make(chan struct{})
+	close(nextReady)
```
jeremyje: Should this be `defer close(nextReady)`?

examples/demo/bytesub/bytesub.go
```diff
@@ -0,0 +1,81 @@
...
+	s.b = b
+}
+
+func (s *ByteSub) get() (b []byte, nextReady chan struct{}) {
+	s.r.RLock()
+	defer s.r.RUnlock()
+	return s.b, s.nextReady
+}
+
+var partialWriteError = errors.New("ByteSub subscriber didn't consume the whole message.")
```
jeremyje: Move to the top.

examples/demo/bytesub/bytesub.go
```diff
@@ -0,0 +1,81 @@
...
+
+type ByteSub struct {
+	nextReady chan struct{}
+	b         []byte
+	r         sync.RWMutex
+}
+
+func New() *ByteSub {
+	return &ByteSub{
+		nextReady: make(chan struct{}),
```
jeremyje: I don't think you need this. You can just check if nextReady is nil. That way you can create this struct via `&ByteSub{}`.

Laremere: When a subscriber subscribes before any values have been announced, they need to have a new channel to listen to.  They can't set the value because they only have a read lock.

examples/demo/bytesub/bytesub.go
Laremere: No the purpose of this channel is to trigger a read in the select immediately without code duplication.

jeremyje:

PR #511, opened 2019-06-11
Add image for demo

Laremere:
This is barebones work to get an image for the demo working.  This image will eventually contain the demo go-routines that emulate the clients, and presenting a dashboard for the state of the demo.  Will followup with actual logic in the demo itself.

TESTED=Manually started om and the demo, ran proxy-demo and got the expected 404.


install/helm/open-match-demo/templates/demo.yaml
```diff
@@ -0,0 +1,86 @@
...
+          requests:
+            memory: 100Mi
+            cpu: 100m
+        ports:
+        - name: http
+          containerPort: {{ .Values.openmatch.demo.http.port }}
+        - name: metrics
+          containerPort: {{ .Values.openmatch.metrics.port }}
+        env:
+        - name: REDIS_SERVICE_HOST
```
jeremyje: Delete these lines of code if redis is not necessary.

jeremyje:

PR #513, opened 2019-06-11
Add filter package

Laremere:
This package will be useful for a simple definition of how filters work, as well as a way to process tickets without relying on indexes.  This will allow other filter types (eg, strings) to be added without a redis implementation.  See package documentation for some more details.

I plan on replacing the current redis indexing with a "all tickets" index to remove the edge cases it gets wrong that we don't want to spend time fixing for v0.6.  Instead this package will be responsible for filtering which tickets to return.  This also removes the index configuration problem from v0.6. Then for v0.7, once the indexing and database solutions are chosen, we can go back to implementing the correct way to be indexing tickets.

Also included are test cases, separated in their own definition.  These test cases should be used in the future for end to end tests, and for tests on indexes.  This will help ensure that the system as a whole maintains the behavior specified here.

### DIFF
--- a/internal/filter/filter.go
+++ b/internal/filter/filter.go
@@ -1,0 +1,66 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package filter defines which tickets pass which filters.  Other implementations which help
+// filter tickets (eg, a range index lookup) must conform to the same set of tickets being within
+// the filter as here.
+package filter
+
+import (
+	structpb "github.com/golang/protobuf/ptypes/struct"
+	"open-match.dev/open-match/internal/pb"
+)
+
+// Fitler returns all of the passed tickets which are included by all the passed filters.
+func Filter(tickets []*pb.Ticket, filters []*pb.Filter) []*pb.Ticket {
+	var result []*pb.Ticket
+
+	for _, t := range tickets {
+		if InFilters(t, filters) {
+			result = append(result, t)
+		}
+	}
+
+	return result
+}
+
+// InFilters returns if the given ticket is included in all of the passed filters.
+func InFilters(ticket *pb.Ticket, filters []*pb.Filter) bool {
+	for _, f := range filters {
+		if !InFilter(ticket, f) {
+			return false
+		}
+	}
+	return true
+}
+
+// InFilter returns if the given ticket is included in the given filter.
+func InFilter(ticket *pb.Ticket, filter *pb.Filter) bool {
+	if ticket.Properties == nil || ticket.Properties.Fields == nil {
+		return false
+	}
+
+	v, ok := ticket.Properties.Fields[filter.Attribute]
+	if !ok {
+		return false
+	}
+
+	switch v.Kind.(type) {
+	case *structpb.Value_NumberValue:
+		nv := v.GetNumberValue()
+		return filter.Min <= nv && nv <= filter.Max
+	default:
+		return false
+	}
+}
--- a/internal/filter/filter.go
+++ b/internal/filter/filter.go
@@ -22,7 +22,7 @@ import (
 	"open-match.dev/open-match/internal/pb"
 )
 
-// Fitler returns all of the passed tickets which are included by all the passed filters.
+// Filter returns all of the passed tickets which are included by all the passed filters.
 func Filter(tickets []*pb.Ticket, filters []*pb.Filter) []*pb.Ticket {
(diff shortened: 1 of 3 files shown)


internal/filter/filter_test.go
```diff
@@ -0,0 +1,204 @@
...
+		})
+	}
+}
+
+type testCase struct {
+	ticket *pb.Ticket
+	filter *pb.Filter
+}
+
+var includedTestCases = []testCase{
```
golangcibot: `includedTestCases` is a global variable (from `gochecknoglobals`)

internal/filter/filter_test.go
```diff
@@ -0,0 +1,204 @@
...
+	simpleRangeTest("simpleInRange", 5, 0, 10),
+	simpleRangeTest("exactMatch", 5, 5, 5),
+	simpleRangeTest("infinityMax", math.Inf(1), 0, math.Inf(1)),
+	simpleRangeTest("infinityMin", math.Inf(-1), math.Inf(-1), 0),
+}
+
+var excludedTestCases = []testCase{
```
golangcibot: `excludedTestCases` is a global variable (from `gochecknoglobals`)

internal/filter/filter.go
```diff
@@ -0,0 +1,66 @@
...
+// filter tickets (eg, a range index lookup) must conform to the same set of tickets being within
+// the filter as here.
+package filter
+
+import (
+	structpb "github.com/golang/protobuf/ptypes/struct"
+	"open-match.dev/open-match/internal/pb"
+)
+
+// Fitler returns all of the passed tickets which are included by all the passed filters.
```
golangcibot: comment on exported function `Filter` should be of the form `Filter ...` (from `golint`)

internal/filter/filter_test.go
Laremere: This is recommended style: https://github.com/golang/go/wiki/TableDrivenTests
I will figure out how to make golangcibot happy.

internal/filter/filter_test.go
Laremere: This is recommended style: https://github.com/golang/go/wiki/TableDrivenTests
I will figure out how to make golangcibot happy.

internal/filter/filter_test.go
Laremere: Thinking about it on the way home, I want to move this to a separate package, so these test cases can be shared across all the use cases (indexing, end to end, etc.)

internal/filter/testcases/testcases.go
```diff
@@ -0,0 +1,110 @@
...
+package testcases
+
+import (
+	"math"
+
+	structpb "github.com/golang/protobuf/ptypes/struct"
+	"open-match.dev/open-match/internal/pb"
+)
+
+type TestCase struct {
```
golangcibot: exported type `TestCase` should have comment or be unexported (from `golint`)

internal/filter/filter.go
```diff
@@ -0,0 +1,66 @@
...
+		if !InFilter(ticket, f) {
+			return false
+		}
+	}
+	return true
+}
+
+// InFilter returns if the given ticket is included in the given filter.
+func InFilter(ticket *pb.Ticket, filter *pb.Filter) bool {
+	if ticket.Properties == nil || ticket.Properties.Fields == nil {
```
yfei1: I think we can move the sanity checks to our `mmlogic_service.go` file. It is the caller's responsibility to make sure the inputs are valid.

internal/filter/testcases/testcases.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"math"
+
+	structpb "github.com/golang/protobuf/ptypes/struct"
+	"open-match.dev/open-match/internal/pb"
+)
+
+// TestCase defines a single filtering test case to run.
+type TestCase struct {
```
yfei1: I would prefer to move this file to `internal/filter` directory and declare it as a private type. What do you think?

internal/filter/filter.go
Laremere: I want this here for a few reasons:

- Not having this check increases the complexity of cross package contracts.
- This package is the definition of how tickets are filtered.  This statement specifies how improper tickets are handled.
- It's actually valid to not have any properties.  If you ask for a tickets with no filters, tickets with no properties are returned.
- The call code will mostly be calling Filter, which contains a list of tickets.  This means callers don't have to do pre-filtering.

internal/filter/testcases/testcases.go
Laremere: That's how I first made it, but I've intentionally moved it out.  Updated commit comment for more info.

internal/filter/filter.go
yfei1: 1. The reason that I wan to move this check is similar sanity checks are done at service level (e.g. [1](https://github.com/googleforgames/open-match/blob/master/internal/app/backend/backend_service.go#L61), [2](https://github.com/googleforgames/open-match/blob/master/internal/app/frontend/frontend_service.go#L49), [3](https://github.com/googleforgames/open-match/blob/master/internal/app/mmlogic/mmlogic_service.go#L53)). They guarantee even if a user pass in an invalid request our service won't panic.
2. I think it is fair to define this package as "how valid tickets are filtered".
3. It is valid to have an empty property bag as long as ticket.Properties != nil (which is checked in frontend_service.go), therefore any call codes with tickets from the storage service already satisfied this sanity check.

PR #514, opened 2019-06-11
Update golangci version and enable it in presubmit check

yfei1:
Golangci 1.17.1 resolves its original segfault issue. It includes gofmt and govet so I use to replace these two presubmit checks.

This commit disable the `bodyclose` plugin when running golangci against the test files as it only accepts closing the http body with specific pattern.

### DIFF
--- a/internal/rpc/server_test.go
+++ b/internal/rpc/server_test.go
@@ -95,8 +95,11 @@ func runGrpcWithProxyTests(assert *assert.Assertions, s grpcServerWithProxy, con
 	httpResp, err := httpClient.Do(httpReq)
 	assert.Nil(err)
 	assert.NotNil(httpResp)
-	defer httpResp.Body.Close()
-
+	defer func() {
+		if httpResp != nil {
+			httpResp.Body.Close()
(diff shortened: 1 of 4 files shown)


internal/rpc/server_test.go
```diff
@@ -95,11 +95,7 @@ func runGrpcWithProxyTests(assert *assert.Assertions, s grpcServerWithProxy, con
 	httpResp, err := httpClient.Do(httpReq)
 	assert.Nil(err)
 	assert.NotNil(httpResp)
-	defer func() {
-		if httpResp != nil {
-			httpResp.Body.Close()
-		}
-	}()
+	defer httpResp.Body.Close()
```
Laremere: assert.Nil does not stop the test case.  It will have an error, then panic at this defer.
Is there a better way to do this that makes golangci happy?

yfei1: this [golangci check](https://github.com/timakin/bodyclose) only accept this body close pattern. I think we have to disable it when running the checks in our test files.

PR #523, opened 2019-06-12
Change GoogleCloudPlatform to googleforgames in v0.5 branch.

jeremyje:
Backporting the same approach as https://github.com/googleforgames/open-match/pull/483 but for `release-0.5` branch.

There's more changes in this PR because in 0.5 we don't use the vanity URL.


jeremyje:
I don't think a new release is needed. The pointers to images and artifacts are not changing.

/gcbrun

jeremyje:
I'm going to close this change. The website will be ported over to the next repository. Instead I'll bring up a follow up change to delete the website from the release-0.5 directory so we don't have 2 sources of truth.

PR #526, opened 2019-06-13
Use absolute paths in the makefile. Fix macos sed bug.

jeremyje:
* `sed -i` commands work on macOS.
* Use absolute paths for `Makefile` targets.
* Merge website build targets in prep for move to open-match-docs.


yfei1:
It seems like the Makefile in our `open-match-docs` repo is not using the abs path. Do you think we need to keep both Makefile consistent?

jeremyje:
There's a pending change to do that.

PR #528, opened 2019-06-13
Disable website autopush in v0.5. Moved to open-match-docs.

jeremyje:


jeremyje:
Force merging, the Golangci is broken due to other issues.

PR #530, opened 2019-06-14
Rename Evaluator component to Synchronizer

sawagh:
The Synchronizer exposes APIs to give a synchronization window context
and to add matches to evaluate for that synchronization window. The
actual evaluator will be re-introduced as the component authored by the
customer that the Synchronizer triggers whenever the window expires.


Makefile
```diff
@@ -810,11 +810,11 @@ proxy-mmlogic: build/toolchain/bin/kubectl$(EXE_EXTENSION)
 	@echo "MmLogic Trace: http://localhost:$(MMLOGIC_PORT)/debug/tracez"
 	$(KUBECTL) port-forward --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) $(shell $(KUBECTL) get pod --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) --selector="app=open-match,component=mmlogic,release=$(OPEN_MATCH_CHART_NAME)" --output jsonpath='{.items[0].metadata.name}') $(MMLOGIC_PORT):51503 $(PORT_FORWARD_ADDRESS_FLAG)
 
-proxy-evaluator: build/toolchain/bin/kubectl$(EXE_EXTENSION)
```
yfei1: I think we can actually get rid of the proxy for synchronizer as it is an internal service

Makefile
```diff
@@ -609,11 +609,11 @@ gcp-apply-binauthz-policy: build/policies/binauthz.yaml
 	gcloud beta $(GCP_PROJECT_FLAG) container binauthz policy import build/policies/binauthz.yaml
 
 all-protos: golang-protos http-proxy-golang-protos swagger-json-docs
-golang-protos: internal/pb/backend.pb.go internal/pb/frontend.pb.go internal/pb/matchfunction.pb.go internal/pb/messages.pb.go internal/pb/mmlogic.pb.go internal/pb/evaluator.pb.go
```
yfei1: `make all-protos` will fail since `api/synchronizer.proto` does not exist

Makefile
```diff
@@ -810,11 +810,11 @@ proxy-mmlogic: build/toolchain/bin/kubectl$(EXE_EXTENSION)
 	@echo "MmLogic Trace: http://localhost:$(MMLOGIC_PORT)/debug/tracez"
 	$(KUBECTL) port-forward --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) $(shell $(KUBECTL) get pod --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) --selector="app=open-match,component=mmlogic,release=$(OPEN_MATCH_CHART_NAME)" --output jsonpath='{.items[0].metadata.name}') $(MMLOGIC_PORT):51503 $(PORT_FORWARD_ADDRESS_FLAG)
 
-proxy-evaluator: build/toolchain/bin/kubectl$(EXE_EXTENSION)
-	@echo "Evaluator Health: http://localhost:$(EVALUATOR_PORT)/healthz"
-	@echo "Evaluator RPC: http://localhost:$(EVALUATOR_PORT)/debug/rpcz"
-	@echo "Evaluator Trace: http://localhost:$(EVALUATOR_PORT)/debug/tracez"
-	$(KUBECTL) port-forward --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) $(shell $(KUBECTL) get pod --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) --selector="app=open-match,component=evaluator,release=$(OPEN_MATCH_CHART_NAME)" --output jsonpath='{.items[0].metadata.name}') $(EVALUATOR_PORT):51506 $(PORT_FORWARD_ADDRESS_FLAG)
+proxy-synchronizer: build/toolchain/bin/kubectl$(EXE_EXTENSION)
+	@echo "synchronizer Health: http://localhost:$(SYNCHRONIZER_PORT)/healthz"
```
jeremyje: Synchronizer

Makefile
sawagh: Looks like the -a did not add all the new / renamed files. Thanks for catching - uploaded now.

Makefile
sawagh: Yes, as per offline discussion we had, this component may have some differences such as not needing http proxy (and the proto, pb being under internal etc.) Let me keep this PR to just renaming and handle all those in a separate issue: https://github.com/googleforgames/open-match/issues/534

Makefile
sawagh: Done!

sawagh:
> I don't see the code for the synchronizer binary.

Good catch! :) This was missing some files from the commit (for some reason -a did not add moved files) You should see them now. The implementation of this PR is simply a placeholder.

PR #531, opened 2019-06-14
Add test for frontend.GetAssignment method

yfei1:


internal/app/frontend/frontend_service_test.go
```diff
@@ -79,6 +81,87 @@ func TestDoCreateTickets(t *testing.T) {
...
+	for i := 0; i < totalAssignments; i++ {
+		testAssignments = append(testAssignments, &pb.Assignment{Connection: string(i)})
+	}
+
+	tests := []struct {
+		description      string
+		count            int
+		preAction        func(*testing.T, statestore.Service)
+		senderGenerator  func(chan *pb.Assignment, *int) func(*pb.Assignment) error
+		shouldCode       codes.Code
```
sawagh: This is a little confusing. Here is a suggestion to make it more readable: use got / want variables (gotCode, wantCode .. gotAssignment, wantAssignment..) to track stuff you expect as result (in want) and you actually get (in got variables) 

In this case, the testAssignments should be wantAssignment and set in each case. shouldCode is wantCode, shouldCount may not be necessary (can just be len of wantAssignment). shouldAssignment is then gotAssignment only stuff you are getting back - can initialize in the calling test case - no need to be a part of the test state.

yfei1: You are right, count is not needed. Comments resolved.

internal/app/frontend/frontend_service_test.go
```diff
@@ -79,6 +81,83 @@ func TestDoCreateTickets(t *testing.T) {
...
+	for i := 0; i < totalAssignments; i++ {
+		wantAssignments = append(wantAssignments, &pb.Assignment{Connection: string(i)})
+	}
+
+	tests := []struct {
+		description        string
+		preAction          func(*testing.T, statestore.Service)
+		senderGenerator    func(chan *pb.Assignment, *int) func(*pb.Assignment) error
+		wantCode           codes.Code
+		wantAssignmentChan chan *pb.Assignment
```
sawagh: You may not have this at all here. you are simply initializing it here but then really using it to get assignments - so right before executing the test case (before the call to doGetAssignments) you can just initialize a gotAssignmentChannel channel variable. (set it to length of wantAssignments described in other comment)

internal/app/frontend/frontend_service_test.go
```diff
@@ -79,6 +81,83 @@ func TestDoCreateTickets(t *testing.T) {
...
+func TestDoGetAssignments(t *testing.T) {
+	testTicket := &pb.Ticket{
+		Id: "test-id",
+	}
+
+	totalAssignments := 2
+	var wantAssignments []*pb.Assignment
+
+	for i := 0; i < totalAssignments; i++ {
+		wantAssignments = append(wantAssignments, &pb.Assignment{Connection: string(i)})
```
sawagh: This should be a part of the test cases. So wantAssignments []*pb.Assignment in the struct. 

In the first test case, since you expect error - it will be empty / uninitialized. 
In the second one, you can initialize it. Also, since these are static expected values, you could simply then initialize as []*pb.Assignment{ <actually just have the assignment objects>}

Thus everythign specific to the test case is within the struct.

PR #533, opened 2019-06-14
minimatch client load gen CLI

yfei1:


internal/app/minimatch/minimatch.go
```diff
@@ -51,9 +56,39 @@ func RunApplication() {
...
 
+	if err := newMinimatchCmd(cfg).Execute(); err != nil {
+		minimatchLogger.Fatal(err)
+		return
+	}
+
 	rpc.MustServeForever(p)
 }
 
+func newMinimatchCmd(cfg config.View) *cobra.Command {
```
jeremyje: I thought we decided to make this a separate tool. It's possible to shell exec and communicate that way.

sawagh: +1 - lets discuss this in person so we can build a more reusable client load gen if required.

PR #536, opened 2019-06-14
Prepare CI for Cluster E2E Testing

jeremyje:
* Added a serving mode for Cluster Reaper
* Add a docker image for the cluster reaper.
* Updated Makefile to use `$(GCLOUD)` instead of `gcloud` for control over verbosity.
* Post submit builds will generate images with the `:canary` label instead of `:dev`.
* Re-enable markdown tests.

For the cluster reaper the plan will be to create a Cloud Run deployment that serves it and a Cloud Scheduler job that tickles it every minute to delete orphaned clusters.

### DIFF
--- a/tools/reaper/internal/internal.go
+++ b/tools/reaper/internal/internal.go
@@ -16,14 +16,15 @@
 package internal
 
 import (
+	container "cloud.google.com/go/container/apiv1"
 	"context"
 	"fmt"
+	containerpb "google.golang.org/genproto/googleapis/container/v1"
 	"log"
+	"net"
+	"net/http"
 	"strings"
 	"time"
-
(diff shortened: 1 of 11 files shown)


tools/reaper/internal/internal.go
```diff
@@ -34,29 +35,79 @@ type Params struct {
...
+func Serve(conn net.Listener, params *Params) error {
+	port := portFromListener(conn)
+	addr := fmt.Sprintf(":%d", port)
+
+	server := &http.Server{
+		Addr: addr,
+	}
+
+	mux := &http.ServeMux{}
+	mux.HandleFunc("/livenessz", func(w http.ResponseWriter, req *http.Request) {
```
yfei1: "healthz"

PR #548, opened 2019-06-18
Reduce the size of the default chart for Open Match.

jeremyje:
This change will remove Jaeger and Prometheus and their dependencies from the default Open Match chart install since we don't use them for development. There's a new target `make install-large-chart` that now holds the old behavior.


yfei1:
I see we still have stackdriver enabled in the slim chart. Do we need stackdriver for development?

jeremyje:
Stackdriver is deployed as part of Kubernetes in kube-system. It's basically free to add the bindings for Stackdriver logging.

PR #549, opened 2019-06-18
Add replicas to Open Match deployments to ensure statelessness.

jeremyje:


install/helm/open-match/values.yaml
```diff
@@ -33,35 +33,41 @@ openmatch:
...
   mmlogic:
     install: true
+    replicas: 3
     grpc:
       port: 50503
     http:
       port: 51503
   synchronizer:
     install: true
+    replicas: 1
```
sawagh: What might be helpful is to add a comment here that indicates why this is 1 (and should be 1 for now) to prevent someone in future accidentally assuming this was a typo (since all others are 3) and setting it to 3.

PR #554, opened 2019-06-19
Use pb getters to avoid program panics when required fields are missing

yfei1:
There are a bunch of places in open match that will panic if some of the required fields are missing and for some functions it is hard to do nil checks for each fields (especially nested fields) of the pb request. This commit replace almost all directly field references by pb's built-in getter function to ensure programs are able to exit gracefully if these fields are missing.

Background: pb's built-in getter functions return typed nil interface if the required field is nil, which makes acquiring some nested fields gracefully returns nil instead of suffering from NPE


jeremyje:
Is there a proto dos and don'ts to provide guidance here? I think we should use protos the way they are intended or at least evaluate those recommendations.

PR #556, opened 2019-06-19
Fix Swagger UI errors.

jeremyje:
Add slash in URL to prevent double appendage in Chrome.

### DIFF
--- a/cmd/swaggerui/Dockerfile
+++ b/cmd/swaggerui/Dockerfile
@@ -19,7 +19,7 @@ RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo .
 COPY api/*.json /go/src/open-match.dev/open-match/third_party/swaggerui/api/
 
 # We need to change the location of the frontend.swagger.json so that we point to this server.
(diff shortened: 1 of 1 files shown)


cmd/swaggerui/Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo .
 COPY api/*.json /go/src/open-match.dev/open-match/third_party/swaggerui/api/
 
 # We need to change the location of the frontend.swagger.json so that we point to this server.
-RUN sed -i 's/url:.*/url: \"api\/frontend.swagger.json\",/g' /go/src/open-match.dev/open-match/third_party/swaggerui/index.html
+RUN sed -i 's/url:.*/url: \"\/api\/frontend.swagger.json\",/g' /go/src/open-match.dev/open-match/third_party/swaggerui/index.html
```
yfei1: This change makes the sed command less readable. Does docker support sed with customized separator? If so, we can rewrite this command to `sed -i |url:.*|url: \"/api/frontend.swagger.json\",|g`

https://stackoverflow.com/a/40715028

jeremyje: Thanks for the suggestion.

PR #557, opened 2019-06-19
 Add e2e tests to Assignment logic and refine redis.UpdateAssignment workflow

yfei1:
This commit includes:
1. Some refinements to redis.UpdateAssignment methods, which does a sanity check before actual assignment updates to make sure all input ids are valid.
2. More test coverage to the corner cases in our assign/create ticket workflow.


test/e2e/minimatch/match_test.go
```diff
@@ -34,20 +34,24 @@ func TestFetchMatches(t *testing.T) {
 	be := pb.NewBackendClient(tc.MustGRPC())
 
 	var tt = []struct {
-		req  *pb.FetchMatchesRequest
-		resp *pb.FetchMatchesResponse
-		code codes.Code
+		description string
+		req         *pb.FetchMatchesRequest
```
sawagh: I would not have req / resp here - I understand this is not specific to this change - so you can refactor this in future. Preferably, you want the test structure to have actual test data - so in this case, I would say - gotProfile, gotMMF, wantMatches - is your test data. Then you create the req based on the data in the execution loop and compare matches, code in response to the wantCode. (Also, please stick to got / want naming - becomes much easier to read the test struct to see whats test data and whats expected response)

test/e2e/minimatch/match_test.go
```diff
@@ -34,20 +34,24 @@ func TestFetchMatches(t *testing.T) {
...
 			&pb.FetchMatchesRequest{},
 			nil,
 			codes.InvalidArgument,
 		},
 	}
 
 	for _, test := range tt {
-		fetchMatchesLoop(t, tc, be, test.req, func(_ *pb.FetchMatchesResponse, err error) {
-			assert.Equal(test.code, status.Convert(err).Code())
+		t.Run(test.description, func(t *testing.T) {
```
sawagh: consider running these different cases in Paralell (see t.Paralell)

test/e2e/minimatch/ticket_test.go
```diff
@@ -28,73 +28,91 @@ import (
...
+	assert.Nil(t, err)
+	fakeAssignment := &pb.Assignment{Connection: "localhost"}
 
 	var tt = []struct {
-		// description string
-		req  *pb.AssignTicketsRequest
-		resp *pb.AssignTicketsResponse
-		code codes.Code
+		description    string
+		req            *pb.AssignTicketsRequest
```
sawagh: Can be avoided - dont think you are using this in the tests?

test/e2e/minimatch/ticket_test.go
```diff
@@ -28,73 +28,91 @@ import (
...
-		if err != nil {
-			assert.Equal(test.code, status.Convert(err).Code())
-		} else {
-			gtResp, err := fe.GetTicket(tc.Context(), &pb.GetTicketRequest{TicketId: ctResp.Ticket.Id})
-			assert.Nil(err)
-			assert.Equal(test.req.Assignment.Connection, gtResp.Assignment.Connection)
-		}
+		t.Run(test.description, func(t *testing.T) {
+			resp, err := be.AssignTickets(tc.Context(), test.req)
+			assert.Equal(t, test.wantResp, resp)
```
sawagh: resp here is not meaningful (always empty) so you could skip that.

test/e2e/minimatch/ticket_test.go
```diff
@@ -28,73 +28,91 @@ import (
...
+			assert.Equal(t, test.wantCode, status.Convert(err).Code())
+			gtResp, err := fe.GetTicket(tc.Context(), &pb.GetTicketRequest{TicketId: ctResp.GetTicket().GetId()})
```
sawagh: got?

PR #558, opened 2019-06-19
Make create ticket returns precondition failure code when receiving n…

yfei1:
…on-number properties and add tests accordingly


internal/app/frontend/frontend_service_test.go
```diff
@@ -32,48 +32,67 @@ import (
...
 			wantErr:     false,
 		},
 	}
 
 	for _, test := range tests {
 		t.Run(test.description, func(t *testing.T) {
-			res, err := doCreateTicket(test.ctx, &pb.CreateTicketRequest{Ticket: testTicket}, store)
+			res, err := doCreateTicket(test.ctx, &pb.CreateTicketRequest{Ticket: test.ticket}, store)
 
 			assert.Equal(t, test.wantErr, err != nil)
```
sawagh: You may want to check the actual failure code too. Always a good practice to validate that it failed with the code you expected it to fail and not something else.

internal/app/frontend/frontend_service_test.go
```diff
@@ -32,48 +32,67 @@ import (
...
 	tests := []struct {
 		description string
+		ticket      *pb.Ticket
 		ctx         context.Context
 		wantErr     bool
 	}{
+		{
+			description: "expect precondition error since open-match does not support properties other than number",
+			ctx:         normalCtx,
+			ticket:      stringTicket,
```
sawagh: simply initialize the ticket objects here instead of having them in named variables just used here. Your description already covers the condition.

internal/statestore/redis.go
```diff
@@ -294,25 +286,34 @@ func (rb *redisBackend) IndexTicket(ctx context.Context, ticket *pb.Ticket) erro
...
 		}
 
-		var d float64
-
 		switch v.Kind.(type) {
 		case *structpb.Value_NumberValue:
-			d = v.GetNumberValue()
+			attributesToValue[attribute] = v.GetNumberValue()
 		default:
-			// TODO: investigate if we need to throw out an exception if attribute is not a value
```
sawagh: Add a todo somewhere indicating that we currently dont support anything but numeric values and in future we might. That way, its clear why we actually have a generically supported proto.

test/e2e/minimatch/minimatch_test.go
```diff
@@ -15,17 +15,13 @@
 package minimatch
```
sawagh: the refactor here is un-related to this PR? Intentionally here?

yfei1: I placed these helper functions in the wrong file in the last PR :( Wanna revert them back in one PR since these moves does not hurt readability that much.

internal/statestore/redis.go
```diff
@@ -294,25 +286,36 @@ func (rb *redisBackend) IndexTicket(ctx context.Context, ticket *pb.Ticket) erro
...
+			// TODO: we currently dont support anything but numeric values,
+			// yet we are investigating whether it is possible to have generic typed value supports in the future.
 			redisLogger.WithFields(logrus.Fields{
 				"attribute": attribute,
 			}).Warning("Attribute not a number.")
+			return status.Errorf(codes.FailedPrecondition, "open-match does not support current value type.")
```
sawagh: nit: replace 'current' with 'given'.

PR #563, opened 2019-06-20
Refactor backend.getClient methods for synchronizer's implementation

yfei1:
Our backend and synchronizer service both need to talk to an external service hosted by users to work properly. This commit refactor `backend.getHTTPClient` and `backend.getGRPCClient` helpers with cache functionality for synchronizer's implementation. @sawagh Let me know if it looks good to you.

### DIFF
--- a/internal/app/backend/backend_service.go
+++ b/internal/app/backend/backend_service.go
@@ -24,12 +24,14 @@ import (
 	"sync"
 
 	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"open-match.dev/open-match/internal/config"
-	"open-match.dev/open-match/internal/rpc"
 	"open-match.dev/open-match/internal/statestore"
 	"open-match.dev/open-match/pkg/pb"
+
+	utilCache "open-match.dev/open-match/internal/util/cache"
 )
 
 // The service implementing the Backend API that is called to generate matches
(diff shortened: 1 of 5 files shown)


internal/app/backend/backend_service.go
```diff
@@ -96,6 +89,7 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
 func doFetchMatchesInChannel(ctx context.Context, cfg config.View, mmfClients *sync.Map, req *pb.FetchMatchesRequest, resultChan chan<- mmfResult) error {
 	var grpcClient pb.MatchFunctionClient
 	var httpClient *http.Client
+	var conn *grpc.ClientConn
```
sawagh: need this as a separate var at top? can just initialize on usage?

internal/app/backend/backend_service.go
```diff
@@ -179,43 +176,6 @@ func doFetchMatchesSendResponse(ctx context.Context, proposals []*pb.Match, send
 	return nil
 }
 
-func getHTTPClient(cfg config.View, mmfClients *sync.Map, funcConfig *pb.FunctionConfig_Rest) (*http.Client, string, error) {
```
sawagh: I have second thoughts about the need for this changes. The Synchronizer (unlike backend) only ever needs to connect to one evaluator - so the repeated code between these components is minimal (the mmfclients map is not relevant to synchronizer). For now, the synchronizer can just initialize the connection during its init - and store it on its service struct - unlike the backend that needs one per MMF. 

The current implementation does not actually abstract the cache anyways. In future, I would expect the 'cache' to actually abstract the map from the caller - thus only supplying the caller with connection. The cache would also handle stale connections and reinitialize them if needed etc. For now, we can let this stay as is.

PR #564, opened 2019-06-20
Simplify mmf config proto definitions

### LINKED ISSUE #562: Mmf config proto struct can be simplified
Instead of having mmf config proto below
```
// Configuration for a GRPC Match Function
message GrpcFunctionConfig {
  string host = 1;
  int32 port = 2;
}

// Configuration for a REST Match Function.
message RestFunctionConfig {
  string host = 1;
  int32 port = 2;
}

// Configuration for the Match Function to be triggered by Open Match to
// generate proposals.
message FunctionConfig {
  // A developer-chosen human-readable name for this Match Function.
  string name = 1;

  // Properties for the type of this function.
  oneof type {
    GrpcFunctionConfig grpc = 10001;
    RestFunctionConfig rest = 10002;
  }
}
```
We can use the following definition
```
// Configuration for the Match Function to be triggered by Open Match to
// generate proposals.
message FunctionConfig {
  string address = 2;
  Type connection = 3;
  enum Type {
    GRPC = 0;
    REST = 1;
  }
}
```

This change can simplify the implementation of `backend.FetchMatches` function and avoid doing the type casting throughout the function call. The name field of the FunctionConfig could also be simplified since calls to the mmfs returns responses with function name in it.

yfei1:
Resolves #562 @sawagh Please let me know you thoughts to this change.


api/backend.proto
```diff
@@ -54,28 +54,14 @@ option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
...
-message GrpcFunctionConfig {
-  string host = 1;
-  int32 port = 2;
-}
-
-// Configuration for a REST Match Function.
-message RestFunctionConfig {
-  string host = 1;
-  int32 port = 2;
-}
```
sawagh: @jeremyje - Is there a recommendation - or a canonical way to accept a GRPC endpoint address? Between passing in entire address v/s accepting host, port and creating the address when generating a client, which one is preferred?

jeremyje: Splitting out host and port seems like the most generic way. I just looked at the java example.

```java
public RouteGuideClient(String host, int port) {
  this(ManagedChannelBuilder.forAddress(host, port).usePlaintext());
}
```

yfei1: I'll update this commit to split out the host and port, yet I think the change of using enum type instead of `oneof` is still valid. This could keep us from casting things around.

PR #565, opened 2019-06-21
Open Match Helm Repository and merge in the Demo.

jeremyje:
* Create a Helm Repository for Open Match
* The Demo chart is now part of the canonical Open Match chart.
* Introduce Chart Tester to validate the Helm chart
* Move configuration deployment to the end of the build.


Makefile
```diff
@@ -330,15 +349,6 @@ install-chart: build/toolchain/bin/helm$(EXE_EXTENSION)
 		--set openmatch.monitoring.stackdriver.enabled=true \
 		--set openmatch.monitoring.stackdriver.gcpProjectId=$(GCP_PROJECT_ID)
 
-install-demo-chart: build/toolchain/bin/helm$(EXE_EXTENSION)
```
yfei1: Should we provide a Makefile commands for open-match install along with the demo for people who wanna get started with using open-match?

install/helm/open-match/templates/demo.yaml
```diff
@@ -37,43 +38,52 @@ metadata:
...
+        {{- include "prometheus.annotations" (dict "port" .Values.openmatch.demo.http.port "prometheus" .Values.openmatch.monitoring.prometheus) | nindent 8 }}
       labels:
         app: {{ template "openmatch.name" . }}
         component: demo
-        chart: {{ template "openmatch.chart" . }}
-        release: {{ .Release.Name }}
-        heritage: {{ .Release.Service }}
+        {{- include "openmatch.chartmeta" (set . "indent" 8) }}
     spec:
+      serviceAccountName: open-match-unprivileged-service
```
yfei1: What is it used for?

Makefile
jeremyje: `make install-chart` will now install the demo by default.

install/helm/open-match/templates/demo.yaml
jeremyje: There's some policies that come along when a service account is bound to the pod. In this case a PodSecurityPolicy that prevents the pod from doing privileged operations.

PR #575, opened 2019-06-24
Implement core synchronizer functionality.

sawagh:
This PR only implements the core Synchronizer functionality. Wiring this up to backend and actually calling out to the User Evaluator will follow in separate PR. This PR has tests for the core functionality providing a Unit Test evaluation method.

### DIFF
--- a/install/helm/open-match/values.yaml
+++ b/install/helm/open-match/values.yaml
@@ -146,8 +146,8 @@ openmatch:
 
         synchronizer:
           enabled: false
-          registrationIntervalMs: 3000
-          proposalCollectionIntervalMs: 2000
+          registrationIntervalMs: 3000ms
+          proposalCollectionIntervalMs: 2000ms
 
         monitoring:
           jaeger:
--- a/install/helm/open-match/values.yaml
+++ b/install/helm/open-match/values.yaml
@@ -146,8 +146,8 @@ openmatch:
 
         synchronizer:
           enabled: false
-          pollIntervalMs: 1000
-          maxWaitMs: 10000
+          registrationIntervalMs: 3000
+          proposalCollectionIntervalMs: 2000
 
         monitoring:
           jaeger:
--- a/internal/app/synchronizer/evaluator_client.go
+++ b/internal/app/synchronizer/evaluator_client.go
@@ -1,0 +1,20 @@
+package synchronizer
+
+import (
+	"open-match.dev/open-match/pkg/pb"
+)
+
+// EvaluatorFunction is the signature of the function that the synchronizer
+// service invokes for evaluation.
+type EvaluatorFunction func([]*pb.Match) []*pb.Match
+
+func evaluate([]*pb.Match) []*pb.Match {
+	// TODO: Add the logic to trigger HTTP / GRPC evaluation client based on user
+	// configuration to trigger the custom evaluator here.
+	// Currently, this hook is only used for unit testing the synchronizer.
+	return nil
+}
+
+func getEvaluator() EvaluatorFunction {
+	return evaluate
+}
--- a/internal/app/synchronizer/evaluator_client.go
+++ b/internal/app/synchronizer/evaluator_client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"sync"
 
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc/codes"
@@ -42,12 +43,15 @@ type evaluatorClient struct {
 	grpcClient pb.EvaluatorClient
 	httpClient *http.Client
 	baseURL    string
+	m          sync.Mutex
 }
 
 // evaluate method triggers execution of user evaluation function. It initializes the configured
 // GRPC or HTTP client. If the client was already initialized, initialize is a no-op.
 func (ec *evaluatorClient) evaluate(proposals []*pb.Match) (result []*pb.Match, err error) {
-	ec.initialize()
+	if err := ec.initialize(); err != nil {
+		return nil, err
+	}
 
 	// Call the appropriate client's evaluation function.
 	switch ec.clType {
@@ -63,34 +67,40 @@ func (ec *evaluatorClient) evaluate(proposals []*pb.Match) (result []*pb.Match, 
 // initialize sets up the configured GRPC or HTTP client the first time it is called. Consequent
 // invocations after a successful client setup are a no-op. It attempts to initialize either GRPC
 // or HTTP clients. If both are defined, it prefers GRPC client.
-func (ec *evaluatorClient) initialize() {
+func (ec *evaluatorClient) initialize() error {
+	ec.m.Lock()
+	defer ec.m.Unlock()
 	if ec.clType != none {
 		// Client already initialized.
-		return
+		return nil
 	}
 
 	// Evaluator client not initialized. Attempt to initialize GRPC client or HTTP client.
 	grpcAddr := fmt.Sprintf("%s:%d", ec.cfg.GetString("api.evaluator.hostname"), ec.cfg.GetInt64("api.evaluator.grpcport"))
 	conn, err := rpc.GRPCClientFromEndpoint(ec.cfg, grpcAddr)
 	if err == nil {
-		evaluatorClientLogger.Infof("Successfully connected to GRPC client at address %v", grpcAddr)
-		// GRPC client connection successful. Initialize client and set type as GRPC.
+		evaluatorClientLogger.WithFields(logrus.Fields{
+			"endpoint": grpcAddr,
+		}).Info("Created a GRPC client for input endpoint.")
 		ec.grpcClient = pb.NewEvaluatorClient(conn)
 		ec.clType = grpcEvaluator
-		return
+		return nil
 	}
 
-	evaluatorClientLogger.Errorf("connecting to GRPC client at %v failed, %v", grpcAddr, err.Error())
+	evaluatorClientLogger.WithError(err).Errorf("failed to get a GRPC client from the endpoint %v", grpcAddr)
 	httpAddr := fmt.Sprintf("%s:%d", ec.cfg.GetString("api.evaluator.hostname"), ec.cfg.GetInt64("api.evaluator.httpport"))
 	client, baseURL, err := rpc.HTTPClientFromEndpoint(ec.cfg, httpAddr)
 	if err != nil {
-		evaluatorClientLogger.Errorf("connecting to HTTP client for %v failed, %v", httpAddr, err.Error())
-		return
+		evaluatorClientLogger.WithError(err).Errorf("failed to get a HTTP client from the endpoint %v", httpAddr)
+		return err
 	}
 
-	evaluatorClientLogger.Infof("Successfully connected to GRPC client at address %v", httpAddr)
+	evaluatorClientLogger.WithFields(logrus.Fields{
+		"endpoint": httpAddr,
+	}).Info("Created a HTTP client for input endpoint.")
 	ec.httpClient = client
 	ec.baseURL = baseURL
+	return nil
 }
 
 func (ec *evaluatorClient) grpcEvaluate(proposals []*pb.Match) (results []*pb.Match, err error) {
@@ -133,13 +143,15 @@ func (ec *evaluatorClient) httpEvaluate(proposals []*pb.Match) (results []*pb.Ma
 
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return nil, status.Errorf(codes.FailedPrecondition, "failed to read from response body for proposals %s: %s", proposalIDs, err.Error())
+		return nil, status.Errorf(codes.FailedPrecondition,
+			"failed to read from response body for proposals %s: %s", proposalIDs, err.Error())
 	}
 
 	pbResp := &pb.EvaluateResponse{}
 	err = json.Unmarshal(body, pbResp)
 	if err != nil {
-		return nil, status.Errorf(codes.FailedPrecondition, "failed to unmarshal response body to response pb for proposals %s: %s", proposalIDs, err.Error())
+		return nil, status.Errorf(codes.FailedPrecondition,
+			"failed to unmarshal response body to response pb for proposals %s: %s", proposalIDs, err.Error())
 	}
 
 	return pbResp.GetMatch(), nil
--- a/internal/app/synchronizer/evaluator_client.go
+++ b/internal/app/synchronizer/evaluator_client.go
@@ -1,20 +1,146 @@
 package synchronizer
 
 import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"open-match.dev/open-match/internal/config"
+	"open-match.dev/open-match/internal/rpc"
 	"open-match.dev/open-match/pkg/pb"
 )
 
-// EvaluatorFunction is the signature of the function that the synchronizer
-// service invokes for evaluation.
-type EvaluatorFunction func([]*pb.Match) []*pb.Match
-
-func evaluate([]*pb.Match) []*pb.Match {
-	// TODO: Add the logic to trigger HTTP / GRPC evaluation client based on user
-	// configuration to trigger the custom evaluator here.
-	// Currently, this hook is only used for unit testing the synchronizer.
-	return nil
+type evaluator interface {
+	evaluate([]*pb.Match) ([]*pb.Match, error)
 }
 
-func getEvaluator() EvaluatorFunction {
-	return evaluate
+type clientType int
+
+// Different type of clients supported for evaluation.
+const (
+	none clientType = iota
+	grpcEvaluator
+	httpEvaluator
+)
+
+var (
+	evaluatorClientLogger = logrus.WithFields(logrus.Fields{
+		"app":       "openmatch",
+		"component": "app.synchronizer.evaluator_client",
+	})
+)
+
+type evaluatorClient struct {
+	clType     clientType
+	cfg        config.View
+	grpcClient pb.EvaluatorClient
+	httpClient *http.Client
+	baseURL    string
+}
+
+// evaluate method triggers execution of user evaluation function. It initializes the configured
+// GRPC or HTTP client. If the client was already initialized, initialize is a no-op.
+func (ec *evaluatorClient) evaluate(proposals []*pb.Match) (result []*pb.Match, err error) {
+	ec.initialize()
+
+	// Call the appropriate client's evaluation function.
+	switch ec.clType {
+	case grpcEvaluator:
+		return ec.grpcEvaluate(proposals)
+	case httpEvaluator:
+		return ec.httpEvaluate(proposals)
+	default:
+		return nil, status.Error(codes.Unavailable, "failed to initialize a connection to the evaluator")
+	}
+}
+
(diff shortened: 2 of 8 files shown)


internal/app/synchronizer/synchronizer_service.go
```diff
@@ -46,6 +132,165 @@ func (s *synchronizerService) Register(ctx context.Context, req *pb.RegisterRequ
...
+	go s.trackProposalWindow()
+}
+
+func (s *synchronizerService) trackProposalWindow() {
+	time.Sleep(s.proposalCollectionInterval())
+	s.stateMutex.Lock()
+	defer s.stateMutex.Unlock()
+	synchronizerServiceLogger.Infof("Changing state from proposalCollection to evaluation")
+	s.state = evaluation
+	s.Evaluate()
```
yfei1: One of my concerns is since we don't know what user would do in the evaluator function, we may have a customized evaluator function that takes more than `s.proposalCollectionInterval` to complete, which results in more and more `Resgiter` calls waiting for the lock and ends up with pod crashes or the evaluator takes too long to respond. Do you think it is worth to consider? Would it be possible to create a new cycleData object per cycle, and execute `s.evalFunc` in a gorountine? This way we could have multiple `s.Evaluate()` running in parallel without hurting the atomicity

install/helm/open-match/values.yaml
```diff
@@ -128,8 +128,8 @@ openmatch:
 
         synchronizer:
           enabled: false
-          pollIntervalMs: 1000
-          maxWaitMs: 10000
+          registrationIntervalMs: 3000
```
jeremyje: Add a todo to make these Durations.

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -46,6 +132,165 @@ func (s *synchronizerService) Register(ctx context.Context, req *pb.RegisterRequ
...
+	// in addProposals while holding the state mutex. The count should be decremented
+	// only after this request completes.
+	err := s.addProposals(req.Id, req.Match)
+	if err != nil {
+		return nil, err
+	}
+
+	// If proposals were successfully added, then the pending request count has been
+	// incremented and should only be reduced when this request is completed.
+	defer s.cycleData.pendingRequests.Done()
```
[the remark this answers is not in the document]
jeremyje: Also because this is crossing method boundaries this may break WaitGroup. Not sure though but there's weird behavior. Make sure there's sufficient test coverage to vet this.

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -46,6 +132,165 @@ func (s *synchronizerService) Register(ctx context.Context, req *pb.RegisterRequ
...
+}
+
+func (s *synchronizerService) fetchResults(id string) []*pb.Match {
+	t := time.Now()
+	var results []*pb.Match
+
+	// fetchMatches will block till evaluation is in progress. The resultsReady channel
+	// is closed only when evaluation is complete. This is used to signal all waiting
+	// requests that the results are ready.
+	<-s.cycleData.resultsReady
```
jeremyje: Add a TODO to support timeout here. I'll add a primitive to handle this. Basically I don't think we should ever wait for channels like this in the serving path. It's prone to denial of service issues.

Also file an issue against me to add it.

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -46,6 +132,165 @@ func (s *synchronizerService) Register(ctx context.Context, req *pb.RegisterRequ
...
+	results = make([]*pb.Match, len(s.cycleData.requests[id].results))
+	copy(results, s.cycleData.requests[id].results)
+	synchronizerServiceLogger.Debugf("Results ready for id: %v, results: %v, elapsed time: %v", id, getMatchIds(results), time.Since(t).String())
```
jeremyje: Add these as context variables via `.WithFields()`

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -46,6 +132,165 @@ func (s *synchronizerService) Register(ctx context.Context, req *pb.RegisterRequ
...
+}
+
+// Evaluate aggregates all the proposals across requests and calls the user configured
+// evaluator with these. Once evaluator returns, it copies results back in respective
+// request data objects and then signals the completion of evaluation. It then waits
+// for all the result processing (by fetchMatches) to be compelted before changing the
+// synchronizer state to idle.
+// NOTE: This method is always called while holding the synchronizer state mutex.
+func (s *synchronizerService) Evaluate() {
+	var aggregateProposals []*pb.Match
```
jeremyje: Use literal init here `aggregateProposals = []*pb.Match{}`

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -46,6 +132,165 @@ func (s *synchronizerService) Register(ctx context.Context, req *pb.RegisterRequ
...
+	proposalMap := make(map[string]string)
+	for id, data := range s.cycleData.requests {
+		aggregateProposals = append(aggregateProposals, data.proposals...)
+		for _, m := range data.proposals {
+			proposalMap[m.MatchId] = id
+		}
+	}
+
+	synchronizerServiceLogger.Infof("Requesting evaluation of %v proposals", len(aggregateProposals))
```
jeremyje: TODO for prom metric here?

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -46,6 +132,165 @@ func (s *synchronizerService) Register(ctx context.Context, req *pb.RegisterRequ
...
+	s.state = idle
+	s.idleCond.Broadcast()
+}
+
+func (s *synchronizerService) trackRegistrationWindow() {
+	time.Sleep(s.registrationInterval())
+	s.stateMutex.Lock()
+	defer s.stateMutex.Unlock()
+	synchronizerServiceLogger.Infof("Changing state from requestRegistration to proposalCollection")
+	s.state = proposalCollection
```
jeremyje: Use a prefix to indicate this is effectively an enum, see: https://golang.org/src/net/http/status.go for example.

internal/app/synchronizer/evaluator_client.go
```diff
@@ -0,0 +1,20 @@
+package synchronizer
+
+import (
+	"open-match.dev/open-match/pkg/pb"
+)
+
+// EvaluatorFunction is the signature of the function that the synchronizer
```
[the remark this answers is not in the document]
sawagh: Since I had evaluator_client implementation ready, rather than adding more detailed TODOS. I added the implementation and made all changes here as to the final state. Please review this file again. Note that some other changes needed for E2E test (actual config values in yaml etc.) will still follow in a separate PR.

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -16,28 +16,114 @@ package synchronizer
...
+// lifespan of this data is same as the lifespan of a request.
+type requestData struct {
+	proposals []*pb.Match // Proposals added by the request for evaluation.
+	results   []*pb.Match // Results from evaluation of the proposals.
+}
+
+// synchronizerData holds the state for a synchronization cycle. This data is
+// reset on every new synchronization cycle.
+type synchronizerData struct {
+	requests        map[string]*requestData // data for all the requests in the current cycle.
```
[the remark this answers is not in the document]
sawagh: Renamed - and added comment. Thanks!

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -46,6 +132,165 @@ func (s *synchronizerService) Register(ctx context.Context, req *pb.RegisterRequ
...
+	const (
+		name                = "synchronizer.proposalCollectionIntervalMs"
+		defaultInterval int = 3000
+	)
+
+	if !s.cfg.IsSet(name) {
+		return time.Duration(defaultInterval)
+	}
+
+	return time.Duration(s.cfg.GetInt(name)) * time.Millisecond
```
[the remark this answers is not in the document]
sawagh: Done!

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -46,6 +132,165 @@ func (s *synchronizerService) Register(ctx context.Context, req *pb.RegisterRequ
...
+	const (
+		name                = "synchronizer.registrationIntervalMs"
+		defaultInterval int = 3000
+	)
+
+	if !s.cfg.IsSet(name) {
+		return time.Duration(defaultInterval)
+	}
+
+	return time.Duration(s.cfg.GetInt(name)) * time.Millisecond
```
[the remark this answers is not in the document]
sawagh: Done

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -46,6 +132,165 @@ func (s *synchronizerService) Register(ctx context.Context, req *pb.RegisterRequ
...
+func (s *synchronizerService) addProposals(id string, proposals []*pb.Match) error {
+	s.stateMutex.Lock()
+	defer s.stateMutex.Unlock()
+
+	if !s.proposalCollection() {
+		return status.Error(codes.DeadlineExceeded, "synchronizer currently not accepting match proposals")
+	}
+
+	if _, ok := s.cycleData.requests[id]; !ok {
+		return status.Error(codes.DeadlineExceeded, "request not present in the current synchronization cycle")
```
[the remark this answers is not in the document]
sawagh: The root cause of this condition is - when the MMF takes too long - and the request to add proposals misses its cycle and shows up in next cycle. At that point, evaluator does not find the id in the current cycle. Hence the DeadlineExceeded error code.

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -16,28 +16,114 @@ package synchronizer
...
+	for (s.state != idle) && (s.state != requestRegistration) {
+		s.idleCond.Wait()
+	}
+
+	defer s.stateMutex.Unlock()
+	if s.state == idle {
+		// After waking up, the first request that encounters the idle state changes
+		// state to requestRegistration and initializes the state for this synchronization
+		// cycle. Consequent registration requests bypass this initialization.
+		synchronizerServiceLogger.Infof("Changing state from idle to requestRegistration")
```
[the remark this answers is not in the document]
sawagh: I definitely plan to move the Info down to Debug very soon - once I've seen this working together in the E2E / Demos (those PRs will come this week). Till then, for the test run you pointed out - the current tracing for state changes is critical (bee super useful in identifying issues). Note that I prefer to minimize (eliminate if possible) Trace logs though unless absolutely necessary.

internal/app/synchronizer/synchronizer_service.go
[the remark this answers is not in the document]
sawagh: Not sure if I understand yufan's concern here. (Add proposal adds to the wait group only when it is past the failure point). After this, if EvaluateProposals fails at any point, the deferred pendingRequests.Done() will reduce this count correctly. The idea is, once there is work queued, we set the defer s.cycleData.pendingRequests.Done() so whenever the method terminates in any way, it will mark the queued work as done. 

@jeremyje - yes, that was a concern. I validated that it indeed works (the wg variable is stored on the service object and is not crossing message boundaries). I ve tests validating that new requests correctly queue behind a set of requests from previous cycle.

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -46,6 +132,165 @@ func (s *synchronizerService) Register(ctx context.Context, req *pb.RegisterRequ
...
+	// Proposal can be added for this request so increment the pending requests
+	s.cycleData.pendingRequests.Add(1)
+
+	// Copy the proposals being added to the request data for the specified id.
+	s.cycleData.requests[id].proposals = make([]*pb.Match, len(proposals))
+	copy(s.cycleData.requests[id].proposals, proposals)
+	return nil
+}
+
+func (s *synchronizerService) proposalCollection() bool {
```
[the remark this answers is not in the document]
sawagh: changed to canAcceptProposal

install/helm/open-match/values.yaml
sawagh: made these Durations.

internal/app/synchronizer/synchronizer_service.go
sawagh: Good point - yes, if the user evaluation blocks this .. it would queue up new registrations.. Added a TODO to timeout here - assigned issue to you. Thanks!

internal/app/synchronizer/synchronizer_service.go
sawagh: Done.

internal/app/synchronizer/synchronizer_service.go
sawagh: Created an issue listing some interesting metrics to be added for synchronizer - including this one. Thanks!

internal/app/synchronizer/synchronizer_service.go
sawagh: added a prefix 'state' to each state.

internal/app/synchronizer/synchronizer_service.go
sawagh: You bring up a good point. See similar comment from Jeremy below. Custom Evaluator taking long would block all pending requests - thereby queueing up any new registrations. I would be concerned with having eval functions run in paralell though - since correctness wise, it would not be right to start registrations when an eval is pending. The timeout suggestion jeremy had might solve this problem - and we can abandon the pending evaluation request. Have created an issue to address this. Thanks!

internal/app/synchronizer/synchronizer_service.go
sawagh: Done. (All places in this file)

internal/app/synchronizer/synchronizer_service.go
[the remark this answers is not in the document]
sawagh: I ve gone back and forth on this. Had offline conversation with both of you. For now, my worry is, in absence of any metrics, the logs are the only way we understand how well these configured timeouts are working. Also, in near future, the most common issue we might see is something is stuck - and we cant repro it. Demo logs from synchronizer would be useful. I am inclined to be more verbose about any such info at the get go - and we can absolutely tone it down once we feel it is unnecessary or too much of an overhead.

internal/app/synchronizer/synchronizer_service.go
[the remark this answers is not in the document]
yfei1: Ah but if add proposals encountered an error, evaluate proposals will return directly at line 142 so that line 147 won’t be executed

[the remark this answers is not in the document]
yfei1: Ah but if add proposals encountered an error, evaluate proposals will return directly at line 142 so that line 147 won’t be executed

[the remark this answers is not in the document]
sawagh: Yes :) and add proposal only calls Add() after all errors return paths are done with. It would have been nice and readable to get the Add right before this defer out of add proposal - but that is a race. We need to hold the state mutex and do the increment to the wg and the actual addition of proposals together. Hence what we do is - wait for all error conditions to be done - then increment the wg and append the list.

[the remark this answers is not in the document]
yfei1: A quick fix is to move line 147 to line 141

sawagh:
I had it thete and that was the race i mentioned to you that -race caught.
I can explain when you are in  but moving to 141 introduces a race yielding
incorrectness. Also, I don't yet see the actual issue in the current logic
that needs to be fixed.

internal/app/synchronizer/synchronizer_service.go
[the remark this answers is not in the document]
sawagh: I had it there and the -race caught a race with it :). I can explain better f2f when you are in but moving to 141 introduces a race yielding incorrectness. Further, it could actually set a defer done before we have really added any proposals and then you have a problem if add fails.  I don't yet see the actual issue in the current logic
that needs to be fixed. We can discuss f2f when you are in.

internal/app/synchronizer/evaluator_client.go
```diff
@@ -0,0 +1,146 @@
...
+		// GRPC client connection successful. Initialize client and set type as GRPC.
+		ec.grpcClient = pb.NewEvaluatorClient(conn)
+		ec.clType = grpcEvaluator
+		return
+	}
+
+	evaluatorClientLogger.Errorf("connecting to GRPC client at %v failed, %v", grpcAddr, err.Error())
+	httpAddr := fmt.Sprintf("%s:%d", ec.cfg.GetString("api.evaluator.hostname"), ec.cfg.GetInt64("api.evaluator.httpport"))
+	client, baseURL, err := rpc.HTTPClientFromEndpoint(ec.cfg, httpAddr)
+	if err != nil {
```
yfei1: Connection is not established yet.
nit: evaluatorClientLogger.Errorf("failed to get a HTTP client from input endpoint")

internal/app/synchronizer/evaluator_client.go
```diff
@@ -0,0 +1,146 @@
...
+		evaluatorClientLogger.Errorf("connecting to HTTP client for %v failed, %v", httpAddr, err.Error())
+		return
```
yfei1: Would it make more sense to fail/panic the program if the initialization failed?

internal/app/synchronizer/synchronizer_service_test.go
```diff
@@ -13,3 +13,221 @@
...
+			testEvaluator: &testEvaluatorData{
+				results: [][]*pb.Match{[]*pb.Match{}, tm[10:20]},
+			},
+			regInterval:  "2000ms",
+			propInterval: "100ms",
+		},
+	}
+
+	for _, tc := range testCases {
+		tc.testEvaluator.eval = &testEvaluatorClient{
```
yfei1: I would prefer using
```
t.Run(tc.description, func(t *testing.T) {
...
})
```
and remove the description log at line 226 since this is the documented pattern for table testing.

Ref: https://github.com/golang/go/wiki/TableDrivenTests

internal/app/synchronizer/evaluator_client.go
```diff
@@ -0,0 +1,146 @@
...
+		ec.grpcClient = pb.NewEvaluatorClient(conn)
+		ec.clType = grpcEvaluator
+		return
+	}
+
+	evaluatorClientLogger.Errorf("connecting to GRPC client at %v failed, %v", grpcAddr, err.Error())
+	httpAddr := fmt.Sprintf("%s:%d", ec.cfg.GetString("api.evaluator.hostname"), ec.cfg.GetInt64("api.evaluator.httpport"))
+	client, baseURL, err := rpc.HTTPClientFromEndpoint(ec.cfg, httpAddr)
+	if err != nil {
+		evaluatorClientLogger.Errorf("connecting to HTTP client for %v failed, %v", httpAddr, err.Error())
```
jeremyje: If you're going to log an error then you can use `.WithError(err)` to capture the full error context. Otherwise the metadata will get dropped.

```golang
evaluatorClientLogger.WithError(err).Errorf("connecting to HTTP client for %s failed", httpAddr)
```

internal/app/synchronizer/evaluator_client.go
sawagh: Done!

internal/app/synchronizer/evaluator_client.go
sawagh: Ah yes. Initially i was calling this in the initialization code for synchronizer - and failing there was not idea. Now that we call it at runtime - its ok to fail since we ll retry in case the failure was transient.

sawagh: So fixed!

internal/app/synchronizer/evaluator_client.go
sawagh: Done.

Issue #576, opened 2019-06-24
CI Deployments Conflict Due to Subnetwork in UPDATING status.


jeremyje:
There's a bug in GKE that causes a create cluster to fail if another cluster is creating on the same subnetwork.

To mitigate this we'll deploy a custom VPC network with 60 subnets for `10.0.X.0/24` CIDR ranges. After that we'll modulo the second the deployment started.

Anyone else have any better ideas here?

jeremyje:
Fixed by https://github.com/googleforgames/open-match/pull/577.

PR #577, opened 2019-06-24
Codify Open Match Continuous Integration as a Terraform template and migrate CI collisions. (#576)

jeremyje:
This change introduces Terraform management of Open Match CI. Right now it'll manage the subnets and service accounts used to allow CI to run. This is GCB/prow agnostic.

Lastly, this change mitigates the subnet collision bug in GKE by provisioning a cluster on it's own subnetwork based on the second the build runs. This isn't a perfect solution but it should mitigate the vast majority of collisions we are seeing.

### DIFF
--- a/install/terraform/open-match-build/README.md
+++ b/install/terraform/open-match-build/README.md
@@ -11,6 +11,24 @@ a migration or emergency.
 If you're making changes to these files you must check in the .tfstate file as
 well as comment the reason why you're enabling a feature or making a change.
 
+## Update Infrastructure
+
+To apply your changes run the following commands:
+
+```bash
+# Get the Terraform tool we use.
+make build/toolchain/bin/terraform
+alias terraform=$PWD/build/toolchain/bin/terraform
+cd install/terraform/open-match-build/
+# Initialize Terraform and download the state of the project.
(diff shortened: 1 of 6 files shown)


install/terraform/open-match-build/README.md
```diff
@@ -0,0 +1,22 @@
+# Open Match Continuous Integration
+
+Open Match runs it's CI in the open-match-build GCP project.
+We primarily use Google Cloud Build to orchestrate our builds and the artifacts are published to GCS and gcr.io/open-match-build.
+
+These Terraform templates codify the infrastructure required to perform CI runs.
+
+```bash
```
sawagh: Are these currently intended only for CIs? Or do we expect demos etc., to use these? I see the project defaulted to "open-match-build" so currently if someone tries these steps manually - would that be a new cluster creation in the project? or would the state of CI cluster be syncd?

jeremyje: This is solely for open-match-build. It'll eventually be possible for someone to take this template and recreate the Open Match CI build pipeline elsewhere. Basically all the tooling we have for this project will be open sourced and checked in.

PR #578, opened 2019-06-25
Add skip ticket checks to FetchMatches and Statestore service

yfei1:


internal/app/backend/backend_service.go
```diff
@@ -94,28 +95,36 @@ func (s *backendService) FetchMatches(ctx context.Context, req *pb.FetchMatchesR
...
+		resp, err = s.synchronizer.Register(ctx, &internalpb.RegisterRequest{})
 		if err != nil {
 			return nil, err
 		}
 
 		syncID = resp.GetId()
 	}
 
-	err := doFetchMatchesInChannel(ctx, s.cfg, s.mmfClients, req, resultChan)
+	err = doFetchMatchesReceiveMmfResult(ctx, s.cfg, s.mmfClients, req, resultChan)
```
sawagh: Currently you discard the proposals that already have spoken for tickets - which leads to work done by MMFs already being discarded. Effectively, till tickets are assigned, MMFs may not be able to make forward progress (they may keep picking the tickets that OM ignores - and OM keeps dropping the results).

The approach we can take (similar to what was in prior OM) is to not return the ignored tickets in the FilterTickets response. Thus the ignored tickets simply act as one more filter for FilterTickets method. 

In the Backend, the only change is that we add the tickets to a ignored ticket list (like you already have). No other proposal check. The other change happens in FilterTickets in the guts of statestore - not exposed through any other API.

internal/app/backend/backend_service.go
```diff
@@ -125,10 +134,15 @@ func (s *backendService) FetchMatches(ctx context.Context, req *pb.FetchMatchesR
 		results = resp.Match
 	}
 
+	err = doFetchMatchesAddSkiplistIds(ctx, s.store, results)
```
sawagh: This is the only task to be done in backend. Btw, why have this helper to 'doFetchMatchesAddSkiplistIds' instead of simply having a loop here that adds tickets to ignored list directly?

install/helm/open-match/values.yaml
```diff
@@ -180,6 +180,8 @@ openmatch:
           # TODO: remove unused config
           results:
             pageSize: 10000
+          skiplist:
```
sawagh: stick to the earlier term 'ignorelist' ?

internal/app/backend/backend_service.go
```diff
@@ -198,6 +212,38 @@ func doFetchMatchesFilterChannel(ctx context.Context, resultChan <-chan mmfResul
...
+}
+
+func doFetchMatchesAddSkiplistIds(ctx context.Context, store statestore.Service, results []*pb.Match) error {
+	ids := []string{}
+	for _, match := range results {
+		for _, ticket := range match.GetTicket() {
+			ids = append(ids, ticket.GetId())
+		}
+	}
+	return store.AddProposedTickets(ctx, ids)
```
sawagh: AddTicketsToIgnoreList ? would clearly communicate the intent.

internal/statestore/public.go
```diff
@@ -50,6 +51,12 @@ type Service interface {
 	// GetAssignments returns the assignment associated with the input ticket id
 	GetAssignments(ctx context.Context, id string, callback func(*pb.Assignment) error) error
 
+	// GetProposedTickets returns the tickets stored under the proposed sorted set withinin TTL period.
+	GetProposedTickets(ctx context.Context, ttl time.Duration) ([]string, error)
```
sawagh: Not needed - the fact that the tickets are ignored can just be internal detail of filter tickets.

internal/app/backend/backend_service.go
yfei1: To make the code more testable. We have an evaluator invocation before this helper call so unit test cant cover this line without hiding the context inside a function.

internal/app/backend/backend_service.go
```diff
@@ -98,10 +101,15 @@ func (s *backendService) FetchMatches(ctx context.Context, req *pb.FetchMatchesR
 		return nil, err
 	}
 
+	err = doFetchMatchesAddIgnoredTickets(ctx, s.store, results)
```
sawagh: Is this helper for testability? otherwise it may not be necessary? 

If you keep it, you may use short form error check - 
if err = do... ; err != nil {

}

internal/statestore/redis.go
```diff
@@ -562,6 +582,40 @@ func (rb *redisBackend) GetAssignments(ctx context.Context, id string, callback
...
 
+// AddProposedTickets appends new proposed tickets to the proposed sorted set with current timestamp
+func (rb *redisBackend) AddTicketsToIgnoreList(ctx context.Context, ids []string) error {
+	redisConn, err := rb.connect(ctx)
+	if err != nil {
+		return err
+	}
+	defer handleConnectionClose(&redisConn)
+
+	err = redisConn.Send("MULTI")
```
sawagh: if err = red...; err != nil {

}

internal/statestore/redis.go
```diff
@@ -423,9 +429,23 @@ func (rb *redisBackend) FilterTickets(ctx context.Context, filters []*pb.Filter,
 		}
 	}
 
+	ttl := rb.cfg.GetDuration("redis.ignoreLists.ttl")
+	curTime := time.Now()
+	curTimeInt := curTime.UnixNano()
+	startTimeInt := curTime.Add(-ttl).UnixNano()
+
+	// Filter out tickets that are fetched but not assigned within ttl time (ms).
+	idsInIgnoreLists, err = redis.Strings(redisConn.Do("ZRANGEBYSCORE", "proposed_ticket_ids", startTimeInt, curTimeInt))
```
sawagh: Do add a TODO somewhere to indicate that we need to clean up the ol ones. Create an issue even better.

Btw, a super cheap way would be to trigger a fire and forget go-routine right here to clean up everything stale in the backend. Its not clean - but for most practical purposes will work well (you can even add that to addignoredlist). Does not need to happen in this PR.

PR #580, opened 2019-06-25
Remove *testing.T dependency from framework structs.

jeremyje:
In preparation for the unified e2e test framework for Minimatch and cluster we'll need to remove the `*testing.T` dependencies since the init code will be moved to `*testing.M`


jeremyje:
Obsolete

PR #587, opened 2019-06-25
Baseline e2e test framework for Kubernetes clusters.

jeremyje:
This is just a baseline implementation of what's to come for the e2e test framework. A later change will unify Minimatch and this.

This change includes a simple health check as well as a proof of concept test to create a ticket to the frontend server based on the ticket_test.go.


jeremyje:
Obsolete

PR #590, opened 2019-06-25
Add evaluator to E2E Minimatch tests

sawagh:


test/e2e/minimatch/minimatch_setup.go
```diff
@@ -39,7 +39,7 @@ const (
 
 // Create a minimatch test service with function bindings from frontend, backend, and mmlogic.
 // Instruct this service to start and connect to a fake storage service.
-func createMinimatchForTest(t *testing.T) *rpcTesting.TestContext {
+func createMinimatchForTest(t *testing.T, evalTc *rpcTesting.TestContext) *rpcTesting.TestContext {
```
jeremyje: Why is this being passed in while the synchronizer does not need to be?

sawagh: This is the host / port for the user defined Evaluation function. Similar to MMF. In case of MMF - we pass the endpoint in each fetchMatch call. As for evaluator, since you only have one, you configure user evaluator endpoint as a configuration on open match. Hence this needs to be passed in for the synchronizer to initialize client to user's evaluator function.

PR #592, opened 2019-06-25
Enable E2E Cluster Tests

jeremyje:
This change enables e2e cluster test to run in CI. Previously it was possible to run e2e against a cluster via `make e2ecluster` but not in CI. This change adds support to scan for a Kubernetes config file via the `KUBECONFIG` environment variable in the test framework and also assign a hard coded path to `KUBECONFIG` in the open-match-build CI docker image.

### DIFF
--- a/internal/testing/e2e/common.go
+++ b/internal/testing/e2e/common.go
@@ -19,23 +19,24 @@ import (
 	"log"
 	pb "open-match.dev/open-match/pkg/pb"
 	"os"
+	"sync"
 	"testing"
 )
 
 // OM is the interface for communicating with Open Match.
 type OM interface {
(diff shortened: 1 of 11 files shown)


internal/testing/e2e/common.go
```diff
@@ -75,3 +76,39 @@ func closeSilently(f func() error) func() {
...
+	m       sync.Mutex
+}
+
+func newMulticloser() *multicloser {
+	return &multicloser{
+		closers: []func(){},
+	}
+}
+
+func (mc *multicloser) add(f func()) {
```
golangcibot: U1000: func `(*multicloser).add` is unused (from `unused`)

jeremyje:
@sawagh this is ready for review.

PR #593, opened 2019-06-25
Reduce Time in CI

jeremyje:
Reorganize some build steps to only download the cached image once.
Change the `make push-helm` to wait for the tiller pod instead of the tiller deployment.


sawagh:
Can you add some description in PR details as to what this change does? How does moving the block lower in cloud build config save CI time?

PR #594, opened 2019-06-26
Enable error handling for evaluation

sawagh:


internal/app/synchronizer/synchronizer_service_test.go
```diff
@@ -215,6 +234,7 @@ func runEvaluationTest(t *testing.T, tc *testData) {
 	for _, c := range tc.testCalls {
 		c := c
 		go func() {
+			ctx := context.Background()
```
jeremyje: It's not clear because of how github does diffing but if you have the RpcTestContext or whatever it's called in scope use the `Context()` method.

internal/app/synchronizer/synchronizer_service_test.go
```diff
@@ -177,14 +176,36 @@ func TestSynchronizerService(t *testing.T) {
...
+		{
+			description: "user evaluator returns an error causing evaluation request to fail",
+			testCalls: []*testCallData{
+				{
+					proposals:          tm[0:5],
+					evaluationrErrCode: codes.ResourceExhausted,
+				},
+				{
+					proposals:          tm[5:10],
+					evaluationrErrCode: codes.ResourceExhausted,
```
jeremyje: `evaluationErrorCode`

internal/app/synchronizer/synchronizer_service_test.go
```diff
@@ -177,14 +176,36 @@ func TestSynchronizerService(t *testing.T) {
...
+				},
+			},
+			testEvaluator: &testEvaluatorData{
+				evalErr: status.Error(codes.ResourceExhausted, "not enough players to make this match"),
```
jeremyje: I'm leaving this as an FYI because I think after reading the docs this is the best error code. Leaving this note anyways to at least document that I've come around to this. The go docs aren't clear on what's the best practice here but the gRPC docs provide better guidance, https://github.com/grpc/grpc/blob/master/doc/statuscodes.md.

> Applications that may wish to retry failed RPCs must decide which status codes on which to retry. As shown in the table above, the gRPC library can generate the same status code for different cases. Server applications can also return those same status codes. **Therefore, there is no fixed list of status codes on which it is appropriate to retry in all applications.** 


`ResourceExhausted` might be fine but `Aborted` might be slightly more suitable.

My concern mainly stems from how `ResourceExhausted` is expected to be handled. The docs aren't clear here but is it retryable or not.

From: https://godoc.org/google.golang.org/grpc/codes#Code

```golang
    // ResourceExhausted indicates some resource has been exhausted, perhaps
    // a per-user quota, or perhaps the entire file system is out of space.
    ResourceExhausted Code = 8

    // FailedPrecondition indicates operation was rejected because the
    // system is not in a state required for the operation's execution.
    // For example, directory to be deleted may be non-empty, an rmdir
    // operation is applied to a non-directory, etc.
    //
    // A litmus test that may help a service implementor in deciding
    // between FailedPrecondition, Aborted, and Unavailable:
    //  (a) Use Unavailable if the client can retry just the failing call.
    //  (b) Use Aborted if the client should retry at a higher-level
    //      (e.g., restarting a read-modify-write sequence).
    //  (c) Use FailedPrecondition if the client should not retry until
    //      the system state has been explicitly fixed. E.g., if an "rmdir"
    //      fails because the directory is non-empty, FailedPrecondition
    //      should be returned since the client should not retry unless
    //      they have first fixed up the directory by deleting files from it.
    //  (d) Use FailedPrecondition if the client performs conditional
    //      REST Get/Update/Delete on a resource and the resource on the
    //      server does not match the condition. E.g., conflicting
    //      read-modify-write on the same resource.
    FailedPrecondition Code = 9

    // Aborted indicates the operation was aborted, typically due to a
    // concurrency issue like sequencer check failures, transaction aborts,
    // etc.
    //
    // See litmus test above for deciding between FailedPrecondition,
    // Aborted, and Unavailable.
    Aborted Code = 10
```

internal/app/synchronizer/synchronizer_service_test.go
sawagh: This is a unit test - so not really initializing any RPC server here.

internal/app/synchronizer/synchronizer_service_test.go
sawagh: Done!

internal/app/synchronizer/synchronizer_service_test.go
sawagh: Updated to Aborted. In this specific case, the error itself is a placeholder - the test only validates whatever error we specify here - is the same that gets returned back as error from evaluation. Nonetheless, good information - Thanks for the pointer!

PR #595, opened 2019-06-27
Add e2eevaluator and e2ematchfunction setup

yfei1:
This commit adds the e2eevaluator and e2ematchfunction to the `make install-chart` command. Will add the helm charts and modify the cloudbuild file in a later PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -351,6 +351,30 @@ install-chart: build/toolchain/bin/helm$(EXE_EXTENSION)
 		--set openmatch.monitoring.stackdriver.enabled=true \
 		--set openmatch.monitoring.stackdriver.gcpProjectId=$(GCP_PROJECT_ID)
 
+install-e2e-chart: build/toolchain/bin/helm$(EXE_EXTENSION)
+	$(HELM) upgrade $(OPEN_MATCH_CHART_NAME) --install --wait --debug install/helm/open-match \
+		--timeout=400 \
+		--namespace=$(OPEN_MATCH_KUBERNETES_NAMESPACE) \
+		--set openmatch.image.registry=$(REGISTRY) \
+		--set openmatch.image.tag=$(TAG) \
+		--set grafana.enabled=false \
+		--set jaeger.enabled=false \
+		--set prometheus.enabled=false \
+		--set redis.enabled=false \
+		--set openmatch.monitoring.stackdriver.enabled=true \
+		--set openmatch.monitoring.stackdriver.gcpProjectId=$(GCP_PROJECT_ID) \
+		--set openmatch.config.install=false \
+		--set openmatch.backend.install=false \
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -349,7 +349,9 @@ install-chart: build/toolchain/bin/helm$(EXE_EXTENSION)
 		--set prometheus.enabled=false \
 		--set redis.enabled=true \
 		--set openmatch.monitoring.stackdriver.enabled=true \
-		--set openmatch.monitoring.stackdriver.gcpProjectId=$(GCP_PROJECT_ID)
+		--set openmatch.monitoring.stackdriver.gcpProjectId=$(GCP_PROJECT_ID) \
+		--set openmatch.e2eevaluator.install=true \
```
jeremyje: I'd prefer not to add these to the default chart because this is the install.yaml file that customers will need to install the baseline Open Match.

I think the best approach is to install-chart first and then install-e2e-chart. You'll need to add install-e2e-chart in the cloudbuild.yaml file to deploy it. Use the deploy chart step to do this.

examples/functions/golang/pool/main.go
```diff
@@ -0,0 +1,35 @@
...
+
+// Package main a sample match function that uses the GRPC harness to set up
+// the match making function as a service. This sample is a reference
+// to demonstrate the usage of the GRPC harness and should only be used as
+// a starting point for your match function. You will need to modify the
+// matchmaking logic in this function based on your game's requirements.
+package main
+
+import (
+	pool "open-match.dev/open-match/examples/functions/golang/pool/mmf"
```
jeremyje: Why are these imports aliased?

install/helm/open-match/values.yaml
```diff
@@ -31,67 +31,82 @@ openmatch:
...
+    replicas: 1
     grpc:
-      port: 50505
+      port: 50501
     http:
-      port: 51505
+      port: 51501
     porttype: LoadBalancer
-  frontend:
+  backend:
```
jeremyje: Why are all these port numbers changing?

examples/functions/golang/pool/main.go
yfei1: I think it might help to better distinguish different mmfs

install/helm/open-match/values.yaml
yfei1: Just wanna make sure we are defining the port sequentially in this yaml file, we have a bunch of different components defined in this file and it is unclear what port we are not using when we wanna add something new if things are defined without any orders. The other point of this change is to use port with different postfix for the demo and e2e components (e.g 5050x for open match core components, 5051x for demo components and 5052x for e2e stuffs.)

jeremyje: This will break cmd/swaggerui

PR #600, opened 2019-06-28
Implements a score-based evaluator for end to end test

yfei1:


examples/consts.go
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package examples defines the constants that some of the examples may share.
+package examples
+
+const (
+	MatchScore = "match_score"
```
jeremyje: FYI: This was introduced as a pattern for `var` that's effectively a const but cannot be declared as one because of how golang treats the type.

examples/evaluator/golang/simple/evaluate/evaluator.go
```diff
@@ -11,14 +11,48 @@
...
 	harness "open-match.dev/open-match/pkg/harness/evaluator/golang"
 	"open-match.dev/open-match/pkg/pb"
 )
 
 // Evaluate is where your custom evaluation logic lives.
+// This sample evaluator sorts and deduplicates the input matches.
 func Evaluate(p *harness.EvaluatorParams) ([]*pb.Match, error) {
-	return p.Matches, nil
+	// Sort in descending numerical order
+	scoreBy := func(a, b *pb.Match) bool {
```
jeremyje: `scoreBy` is not very descriptive. `scoreInDescendingOrder`

That way it reads `by(scoreInDescendingOrder).Sort(p.Matches)` which is still awkward but more descriptive.

examples/evaluator/golang/simple/evaluate/evaluator.go
```diff
@@ -11,14 +11,48 @@
...
+	"open-match.dev/open-match/examples"
 	harness "open-match.dev/open-match/pkg/harness/evaluator/golang"
 	"open-match.dev/open-match/pkg/pb"
 )
 
 // Evaluate is where your custom evaluation logic lives.
+// This sample evaluator sorts and deduplicates the input matches.
 func Evaluate(p *harness.EvaluatorParams) ([]*pb.Match, error) {
-	return p.Matches, nil
+	// Sort in descending numerical order
```
jeremyje: If you make the just below you don't need this comment.

examples/evaluator/golang/simple/evaluate/evaluator_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}{
+		{
+			description: "test deduplicates and expect the one with higher score",
+			testMatches: []*pb.Match{ticket12Score1, ticket12Score10},
+			wantMatches: []*pb.Match{ticket12Score10},
+		},
+		{
+			description: "test first returns matches with higher score",
+			testMatches: []*pb.Match{ticket123Score5, ticket12Score10},
+			wantMatches: []*pb.Match{ticket12Score10, ticket123Score5},
```
sawagh: Here the wantMatches is just ticket12Score10. The Evaluator has to de-dupe and pick only one match that uses a ticket. The goal of deduping should be - any ticket that has appeared in a 'result' should not reappear in across (hence we add it to ignore list to even sit it out of future rounds)

examples/evaluator/golang/simple/evaluate/evaluator_test.go
```diff
@@ -0,0 +1,86 @@
...
+	ticket123Score5 := &pb.Match{
+		Ticket: []*pb.Ticket{ticket1, ticket2, ticket3},
+		Properties: &structpb.Struct{
+			Fields: map[string]*structpb.Value{
+				examples.MatchScore: {Kind: &structpb.Value_NumberValue{NumberValue: 5}},
+			},
+		},
+	}
+
+	tests := []struct {
```
sawagh: You have a good setup to add more tests easily. Would love to see a few other test cases to make this comprehensive:
0 matches, 1 match, 3 matches no collisions, 3 matches - 1 collision

examples/evaluator/golang/simple/evaluate/evaluator.go
```diff
@@ -11,14 +11,46 @@
...
+	}
+	by(scoreInDescendingOrder).Sort(p.Matches)
+
+	results := []*pb.Match{}
+
+	dedup := map[string]bool{}
+	for _, match := range p.Matches {
+		valid := true
+		ids := []string{}
+		for i := 0; valid && i < len(match.GetTicket()); i++ {
```
sawagh: nit feedback on reorganizing code here: 

for _, match := range p.Matches {
    if isNonCollidingMatch(match, selectedTickets) {
        // Add tickets to dedup
        // Add to match to results
   }
}
func isNonCollidingMatch(...) {
    for _, ticket := range match.GetTicket() {
    // If ticket id belongs to the selected players set, return false
   }
}

PR #603, opened 2019-07-01
Add post validation to keep backend from sending matches with empty tickets

yfei1:


api/backend.proto
```diff
@@ -75,7 +75,7 @@ message FetchMatchesRequest {
 }
 
 message FetchMatchesResponse {
-  // Result Match for the requested MatchProfile.
+  // Result Match for the requested MatchProfile. Note that open-match will validate the proposals returned from MMF and skip the ones with empty tickets.
```
sawagh: nit: Split comment to next line. also can simply state - a valid match should contain atleast one ticket.

Also, We should add that comment actually to the Match proto too.

internal/app/backend/backend_service.go
```diff
@@ -163,17 +163,24 @@ func doFetchMatchesReceiveMmfResult(ctx context.Context, cfg config.View, mmfCli
...
+			// Check if mmf responds with any errors
 			if result.err != nil {
 				return nil, result.err
 			}
-			proposals = append(proposals, result.matches...)
+
+			// Check if mmf returns a match with no tickets in it
+			for _, match := range result.matches {
+				if len(match.GetTicket()) != 0 {
+					proposals = append(proposals, match)
```
sawagh: Isnt this identical to before? I thought we said we would fail the request if invalid match was returned?

PR #604, opened 2019-07-02
Updater logic for the demo

Laremere:
Updater allows concurrent processes to update different fields on a json object
which is serialized and passed to a func([]byte).  In the demo, different SetFunc
for different fields on the base json object will be passed to the relevant components 
(clients, director, game servers, etc).  These components will run and pass their state
to the updater.

This updater will be combined with bytesub by passing bytesub's AnnounceLatest
method into the base updater New.  This way the demo state of each component
will be passed to all current dashboard viewers.


examples/demo/updater/updater.go
```diff
@@ -0,0 +1,133 @@
...
+	set      SetFunc
+}
+
+// SetFunc serializes the value passed in into json and sets the associated field
+// to that value.  If nil is passed (BUT NOT a nil value of an interface), the
+// field will be removed from the Updater's json object.
+type SetFunc func(v interface{})
+
+// New creates an Updater, which will call set with the json serialized value
+// set the Updater's fields.
```
Laremere: Comment needs improvement to be English.

examples/demo/updater/updater_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	for i := 0; i < 100; i++ {
+		set := u.ForField(strconv.Itoa(i))
+		go func() {
+			set("Hi")
+			wg.Done()
+		}()
+	}
+
+	// Blocking call ensures that canceling the context will clean up the itnernal go routine.
```
Laremere: typo

examples/demo/updater/updater_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	latest := make(chan string)
+
+	base := New(ctx, func(b []byte) {
+		latest <- string(b)
+	})
+
+	child := NewNested(ctx, base.ForField("Foo"))
+
+	l := <-latest
```
sawagh: move this above line 35? This should be valid even before creating the new nested one right?

sawagh: If something was wrong and the value was not being set - wouldnt this block instead of failing?

examples/demo/updater/updater_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	l = <-latest
+	if l != "{\"Foo\":{}}" {
+		t.Errorf("Got %s, expected %s", l, "{\"Foo\":{}}")
+	}
+
+	child.ForField("Bar")(interface{}((*int)(nil)))
+
+	l = <-latest
+	if l != "{\"Foo\":{\"Bar\":null}}" {
```
sawagh: use the assert (or require) library being used in the rest of the testing for Open Match?

examples/demo/updater/updater_test.go
Laremere: Yup.  Fixed.

Laremere: The test would deadlock and therefor fail.

examples/demo/updater/updater_test.go
Laremere: Stylistically, I prefer not to, and I haven't seen any guidance on this yet.  Is the assert library something we intentionally chose after reviewing options, or just something that someone chose at some point?

examples/demo/updater/updater_test.go
sawagh: Would you rather reasonably timeout and fail? than deadlock and fail? I think deadlock and fail will have no good context on why it failed - and am not sure of how long before the framework decides it failed - for our CIs etc.

Laremere: Deadlocks have nice stack traces: https://play.golang.org/p/u7hcDtANkNl

jeremyje:

PR #607, opened 2019-07-02
Add mmf and evaluator setup for e2e in cluster tests

yfei1:
Will remove the setup codes under the test/e2e directory in a later PR.


install/helm/open-match/values.yaml
```diff
@@ -269,6 +269,12 @@ openmatch:
     demo:
       name: openmatch-demo
       pullPolicy: Always
+    e2ematchfunction:
```
sawagh: May be a future improvement comment - but should we consider going back to the model where there is a core Open Match services installer chart - and then one for the demo / separate one for test? Lets discuss when @jeremyje is back (since he merged them into one)

internal/testing/e2e/cluster.go
```diff
@@ -78,22 +79,34 @@ func (com *clusterOM) MustMmLogicGRPC() pb.MmLogicClient {
...
+func (com *clusterOM) MustMmfConfigGRPC() *pb.FunctionConfig {
+	host, port := com.getAddressFromServiceName("om-e2ematchfunction")
+	return &pb.FunctionConfig{
+		Host: host,
+		Port: port,
+		Type: pb.FunctionConfig_GRPC,
+	}
+}
+
+func (com *clusterOM) getAddressFromServiceName(serviceName string) (string, int32) {
```
sawagh: if this helper was needed only within getGRPCClientFromServiceName - why refactor it out of that method? (unless you plan to reuse it?)

internal/testing/e2e/in_memory.go
```diff
@@ -123,3 +146,32 @@ func createMinimatchForTest(t *testing.T) *rpcTesting.TestContext {
 	tc.AddCloseFunc(closer)
 	return tc
 }
+
+// Create a mmf service using a started test server.
+// Inject the port config of mmlogic using that the passed in test server
+func createMatchFunctionForTest(t *testing.T, c *rpcTesting.TestContext) *rpcTesting.TestContext {
```
sawagh: does this render the setup files in current e2e minimatch tests to setup mmf, evaluator obsolete? Are we not removing them for now because there are other tests using them that need to be ported?

internal/testing/e2e/in_memory.go
```diff
@@ -123,3 +146,32 @@ func createMinimatchForTest(t *testing.T) *rpcTesting.TestContext {
...
+
+		assert.Nil(t, mmfHarness.BindService(p, cfg, &mmfHarness.FunctionSettings{
+			Func: pool.MakeMatches,
+		}))
+	})
+	return tc
+}
+
+// Create an evaluator service that will be used by the minimatch tests.
+func createEvaluatorForTest(t *testing.T) *rpcTesting.TestContext {
```
sawagh: Shouldnt we be setting the ports that the evaluator got started on in createMinimatchForTest for the synchronizer to know where to find the evaluator? Like the below:
	cfg.Set("api.synchronizer.hostname", tc.GetHostname())
	cfg.Set("api.synchronizer.grpcport", tc.GetGRPCPort())
	cfg.Set("api.synchronizer.httpport", tc.GetHTTPPort())

internal/testing/e2e/cluster.go
yfei1: This helper is also used by `MustMmfConfigGRPC`

internal/testing/e2e/in_memory.go
yfei1: My plan is to migrate the setup from `test/e2e` folder to `internal/testing/e2e` folder since it abstract the implementation details. Starting from this PR, all the end to end tests will use the setup code in this file. I'll obsolete the original setup codes in a later PR.

install/helm/open-match/values.yaml
yfei1: I'll open an issue to track whether we need to split up the configmap and helm chart.

PR #610, opened 2019-07-03
FetchMatches e2e tests

yfei1:
Depends on #607


test/e2e/tickets/fetch_matches_test.go
```diff
@@ -16,21 +16,29 @@
...
 	"github.com/stretchr/testify/assert"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	statestoreTesting "open-match.dev/open-match/internal/statestore/testing"
 	"open-match.dev/open-match/internal/testing/e2e"
 	"open-match.dev/open-match/pkg/pb"
 )
 
 func TestGameMatchWorkFlow(t *testing.T) {
+	var gotFmResp *pb.FetchMatchesResponse
```
sawagh: need to declare all of these upfront? check if avoidable - if not, may be declare close to first use? Havent see golang code have all variables declared uptop a function.

yfei1: Sometimes I have to declare these variables instead of using inline declaration because of golangci's shadow check, yet it looks cumbersome both having the `var` declarations with inline ones in one function so I just move everything upfront and try to use `var` instead. I'll try to figure out an alternative.

PR #611, opened 2019-07-03
Open Match demo creating tickets, matches, and assignments

Laremere:
This is the working end to end demo!

There are 3 components to the demo:
- Uptime just counts up once every second.
- Clients simulates 5 game clients, which create a ticket, then wait for assignment on that ticket.
- Director simulates a single director, requesting matches and giving fake assignments to all of the tickets.

To run: make create-gke-cluster push-helm push-images install-chart proxy-demo


examples/demo/components/director/director.go
```diff
@@ -0,0 +1,150 @@
...
+	var matches []*pb.Match
+	{
+		req := &pb.FetchMatchesRequest{
+			Config: &pb.FunctionConfig{
+				Host: "om-demofunction",
+				Port: 50502,
+				Type: pb.FunctionConfig_GRPC,
+			},
+			Profile: []*pb.MatchProfile{
+				&pb.MatchProfile{
```
yfei1: This is enough. Run `make presubmit` to fix it.
```suggestion
				{
```

examples/demo/components/director/director.go
```diff
@@ -0,0 +1,150 @@
...
+
+	//////////////////////////////////////////////////////////////////////////////
+	s.Status = "Match Match: Sending Request"
+	ds.Update(s)
+
+	var matches []*pb.Match
+	{
+		req := &pb.FetchMatchesRequest{
+			Config: &pb.FunctionConfig{
+				Host: "om-demofunction",
```
yfei1: Would it be possible to read these configurations from a file or something? If not, I prefer using patterns like [this](https://github.com/googleforgames/open-match/blob/master/internal/testing/e2e/cluster.go#L91) rather than hardcoding the configs

examples/demo/components/director/director.go
```diff
@@ -0,0 +1,150 @@
...
+				Connection: fmt.Sprintf("%d.%d.%d.%d:2222", rand.Intn(256), rand.Intn(256), rand.Intn(256), rand.Intn(256)),
+			},
+		}
+
+		resp, err := be.AssignTickets(ds.Ctx, req)
+		if err != nil {
+			panic(err)
+		}
+
+		_ = resp
```
yfei1:
```suggestion
```

examples/demo/main.go
```diff
@@ -59,22 +66,40 @@ func main() {
...
 		bs.Subscribe(ws.Request().Context(), ws)
 	}))
 
 	logger.Info("Starting Server")
 
+	go startComponents(cfg, u)
+
 	// TODO: Other services read their port from the common config map, how should
 	// this be choosing the ports it exposes?
 	err = http.ListenAndServe(":51507", nil)
```
yfei1: read the port from kubectl client

examples/demo/components/director/director.go
Laremere: If I don't include this line, resp becomes unused, and throws an error.  I don't want not create resp in the first place because I want to be explicit here that we're doing nothing with the response.

examples/demo/main.go
Laremere: kubectl isn't included in our images?  How are you expecting me to do this?  (Also, this should maybe be a different pr?)

jeremyje:

PR #612, opened 2019-07-03
Use struct in assignment properties

Laremere:
Fixing because I noticed that we were still using string here.
Also properties was stating that it was optional and open match didn't interpret the contents, which is true for all of Assignments fields, so clarified that.


jeremyje:
Can you run `make presubmit` to update the generated files.

PR #618, opened 2019-07-08
Add error to MMF harness signature.

jeremyje:
Add `error` to the MMF function signature because the inputs could be malformed.


examples/functions/golang/soloduel/mmf/matchfunction_test.go
```diff
@@ -39,8 +39,9 @@ func TestMakeMatchesDeduplicate(t *testing.T) {
 		PoolNameToTickets: poolNameToTickets,
 	}
 
-	matches := MakeMatches(p)
+	matches, err := MakeMatches(p)
 	assert.Equal(len(matches), 0)
+	assert.Nil(err)
```
sawagh: nit - this check before the matches length check? if assert was set to use require library, then the test log would miss the real failure.

PR #621, opened 2019-07-08
Update Terraform setup and README

yfei1:
This commit does the following things:
1. Add GCP service account setup to Terraform README.
2. Enable GCP API using Terraform
3. Fix some errors in `open-match/secure-gke.tf` file - Saw an error `Zone us-west1-a doesn't support Haswell Platform`


install/terraform/open-match/secure-gke.tf
```diff
@@ -59,6 +59,22 @@ variable "gcp_machine_type" {
...
+# Enable Kubernetes API
+resource "google_project_service" "kubernetes_api" {
+  project     = "${var.gcp_project_id}"
+  service     = "container.googleapis.com"
+  
+  disable_dependent_services= true
+}
+
+# Enable Cloud Resource Manager API
+resource "google_project_service" "cloud_resource_manager_api" {
```
jeremyje: Use `google_project_services` instead https://www.terraform.io/docs/providers/google/r/google_project_services.html.

install/terraform/open-match/secure-gke.tf
```diff
@@ -59,6 +59,22 @@ variable "gcp_machine_type" {
...
+  project     = "${var.gcp_project_id}"
+  service     = "cloudresourcemanager.googleapis.com"
+  
+  disable_dependent_services= true
```
jeremyje: `disable_dependent_services` looks very risky. I'm not sure we should enable this because it can lead to data loss.

install/terraform/open-match/secure-gke.tf
```diff
@@ -265,6 +271,8 @@ resource "google_container_node_pool" "om-services" {
   node_count = 5
   project = "${google_container_cluster.primary.project}"
   version = "1.13"
+
+  depends_on = [google_project_service.cloud_resource_manager_api, google_project_service.kubernetes_api]
```
jeremyje: `google_project_services.gcp_apis`

install/terraform/open-match-build/README.md
```diff
@@ -11,6 +11,17 @@ a migration or emergency.
 If you're making changes to these files you must check in the .tfstate file as
 well as comment the reason why you're enabling a feature or making a change.
 
+## GCP Service Account Setup
+To use the terraform templates when developing Open Match, you need to have a [GCP service account](https://cloud.google.com/docs/authentication/getting-started) associated with your Open Match project and generate a key file. The terraform templates would read the credentials from `./cred.json` file to authenticate with GCP.
```
jeremyje: I think pointing to this documentation is better:
https://www.terraform.io/docs/providers/google/provider_reference.html#credentials-1

Also we should remove creds.json from the file. It makes it harder to do automation.

PR #633, opened 2019-07-10
Obsolete e2e test setup files

yfei1:
This commit obsoletes the setup files under the `test/e2e/minimatch` directory and instruct the tests to use the test setup code under the`internal/testing` directory instead.


internal/testing/e2e/cluster.go
```diff
@@ -80,27 +80,43 @@ func (com *clusterOM) MustMmLogicGRPC() pb.MmLogicClient {
...
+		if servicePort.Name == portName {
+			port = servicePort.Port
+		}
+	}
+	return svc.Status.LoadBalancer.Ingress[0].IP, port
 }
 
 func (com *clusterOM) getGRPCClientFromServiceName(serviceName string) (*grpc.ClientConn, error) {
-	ipAddress, port := com.getAddressFromServiceName(serviceName)
+	ipAddress, port := com.getAddressFromServiceName(serviceName, "grpc")
```
jeremyje: Use separate names for the methods like `getGRPCAddressFromServiceName`.

test/e2e/match_test.go
```diff
@@ -53,7 +49,7 @@ func TestFetchMatches(t *testing.T) {
 		{
 			"expects unavailable code since there is no mmf being hosted with given function config",
 			&pb.FunctionConfig{
-				Host: mmfTc.GetHostname(),
+				Host: "om-matchfunction",
```
jeremyje: Why is this hard coded now? The `GetHostname` may return "localhost" for minimatch.

yfei1: Cuz OMCluster does not expose mmf text context. In addition, this test cases expect an invalid FunctionConfig, which is okay even if we pass in a random string.

PR #639, opened 2019-07-10
Simplify proxy commands in makefile

Laremere:
I manually tested all of the ones I changed.  I did not test all of the listed endpoints though. (the rpcz and tracez stuff doesn't work I think.)

Didn't update `proxy-dashboard` because I couldn't get it to work pre-change and I don't know what deployment it is trying to reach.


Laremere:
Sorry, not ready for review quite yet, realized port-forward also supports services, which is what we really want to do.

PR #643, opened 2019-07-11
Add metrics for Open Match

jeremyje:
This is a first pass and very rough attempt at adding specific metrics to Open Match. Many of these metrics may not be immediately useful but we need something in for now so that we can build dashboards and analyze key metrics based on stress testing.

This change is also removing some logger instances. There will be a future changes to make 1 logger per package unless there's a reason to keep a separate one for example, delegated client calls.

### DIFF
--- a/internal/app/backend/backend.go
+++ b/internal/app/backend/backend.go
@@ -23,30 +23,23 @@ import (
 	"open-match.dev/open-match/pkg/pb"
 )
 
-var (
-	backendLogger = logrus.WithFields(logrus.Fields{
-		"app":       "openmatch",
-		"component": "backend",
-	})
-)
-
 // RunApplication creates a server.
 func RunApplication() {
 	cfg, err := config.Read()
 	if err != nil {
-		backendLogger.WithFields(logrus.Fields{
+		logger.WithFields(logrus.Fields{
 			"error": err.Error(),
 		}).Fatalf("cannot read configuration.")
 	}
 	p, err := rpc.NewServerParamsFromConfig(cfg, "api.backend")
 	if err != nil {
-		backendLogger.WithFields(logrus.Fields{
+		logger.WithFields(logrus.Fields{
 			"error": err.Error(),
 		}).Fatalf("cannot construct server.")
 	}
 
 	if err := BindService(p, cfg); err != nil {
-		backendLogger.WithFields(logrus.Fields{
+		logger.WithFields(logrus.Fields{
 			"error": err.Error(),
 		}).Fatalf("failed to bind backend service.")
 	}
--- a/internal/app/frontend/frontend_service.go
+++ b/internal/app/frontend/frontend_service.go
@@ -24,6 +24,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"open-match.dev/open-match/internal/config"
+	"open-match.dev/open-match/internal/monitoring"
 	"open-match.dev/open-match/internal/statestore"
 	"open-match.dev/open-match/pkg/pb"
 )
@@ -36,10 +37,14 @@ type frontendService struct {
 }
 
 var (
-	frontendServiceLogger = logrus.WithFields(logrus.Fields{
+	logger = logrus.WithFields(logrus.Fields{
 		"app":       "openmatch",
-		"component": "app.frontend.frontend_service",
+		"component": "app.frontend",
 	})
+	mTicketsCreated             = monitoring.Counter("frontend/tickets_created", "tickets created")
+	mTicketsDeleted             = monitoring.Counter("frontend/tickets_deleted", "tickets deleted")
+	mTicketsRetrieved           = monitoring.Counter("frontend/tickets_retrieved", "tickets retrieved")
+	mTicketAssignmentsRetrieved = monitoring.Counter("frontend/tickets_assignments_retrieved", "ticket assignments retrieved")
 )
 
 // CreateTicket will create a new ticket, assign an id to it and put it in state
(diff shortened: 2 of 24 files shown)


internal/app/backend/backend.go
```diff
@@ -23,30 +23,23 @@ import (
 	"open-match.dev/open-match/pkg/pb"
 )
 
-var (
-	backendLogger = logrus.WithFields(logrus.Fields{
```
sawagh: I remember you setting this up earlier with an intent to have one per file - now we are switching to a logger per package? What is the intent and generally the recommendation for Open Match components going forward?

internal/statestore/public.go
```diff
@@ -59,5 +60,11 @@ type Service interface {
 
 // New creates a Service based on the configuration.
 func New(cfg config.View) Service {
-	return newRedis(cfg)
+	s := newRedis(cfg)
+	if cfg.GetBool(monitoring.ConfigNameEnableMetrics) {
+		return &instrumentedService{
```
sawagh: Rather than creating an instrumented service, can we rather have a shim on the record that checks for the enabled setting to decide whether to record or not?

internal/app/backend/backend.go
jeremyje: Originally the intent was to mirror what Open Match had before for consistency. After looking at how the server runs it's not very helpful to distinguish exactly which file the log statement comes from and there's options to provide better feedback here. Also this is separation will be further diminished when we fix up the error handling.

internal/statestore/public.go
jeremyje: I wanted something that when/if we move to another backend we can use the same wrapper instead of re-adding the logic there.

sawagh: Ah. I misunderstood. Do we need to config control it at all then? Why not just have public implement the wrapper itself - so you always return the wrapped implementation?

internal/statestore/instrumented.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"context"
+
+	"open-match.dev/open-match/internal/monitoring"
+	"open-match.dev/open-match/pkg/pb"
+)
+
+var (
+	mStateStoreCreateTicket           = monitoring.Counter("statestore/createticket", "tickets created")
```
golangcibot: `mStateStoreCreateTicket` is a global variable (from `gochecknoglobals`)

internal/statestore/instrumented.go
```diff
@@ -0,0 +1,116 @@
...
+	mStateStoreGetTicket              = monitoring.Counter("statestore/getticket", "tickets retrieve")
```
golangcibot: `mStateStoreGetTicket` is a global variable (from `gochecknoglobals`)

internal/statestore/instrumented.go
```diff
@@ -0,0 +1,116 @@
...
+	mStateStoreDeleteTicket           = monitoring.Counter("statestore/deleteticket", "tickets deleted")
```
golangcibot: `mStateStoreDeleteTicket` is a global variable (from `gochecknoglobals`)

internal/statestore/instrumented.go
```diff
@@ -0,0 +1,116 @@
...
+	mStateStoreIndexTicket            = monitoring.Counter("statestore/indexticket", "tickets indexed")
```
golangcibot: `mStateStoreIndexTicket` is a global variable (from `gochecknoglobals`)

internal/statestore/instrumented.go
```diff
@@ -0,0 +1,116 @@
...
+	mStateStoreDeindexTicket          = monitoring.Counter("statestore/deindexticket", "tickets deindexed")
```
golangcibot: `mStateStoreDeindexTicket` is a global variable (from `gochecknoglobals`)

internal/statestore/instrumented.go
```diff
@@ -0,0 +1,116 @@
...
+	mStateStoreFilterTickets          = monitoring.Counter("statestore/filterticket", "tickets that were filtered and returned")
```
golangcibot: `mStateStoreFilterTickets` is a global variable (from `gochecknoglobals`)

internal/statestore/instrumented.go
```diff
@@ -0,0 +1,116 @@
...
+	mStateStoreUpdateAssignments      = monitoring.Counter("statestore/updateassignment", "tickets assigned")
```
golangcibot: `mStateStoreUpdateAssignments` is a global variable (from `gochecknoglobals`)

internal/statestore/instrumented.go
```diff
@@ -0,0 +1,116 @@
...
+	mStateStoreGetAssignments         = monitoring.Counter("statestore/getassignments", "ticket assigned retrieved")
```
golangcibot: `mStateStoreGetAssignments` is a global variable (from `gochecknoglobals`)

internal/statestore/instrumented.go
```diff
@@ -0,0 +1,116 @@
...
+	mStateStoreAddTicketsToIgnoreList = monitoring.Counter("statestore/addticketstoignorelist", "tickets moved to ignore list")
```
golangcibot: `mStateStoreAddTicketsToIgnoreList` is a global variable (from `gochecknoglobals`)

internal/app/frontend/frontend_service.go
```diff
@@ -36,10 +37,12 @@ type frontendService struct {
 }
 
 var (
-	frontendServiceLogger = logrus.WithFields(logrus.Fields{
+	logger = logrus.WithFields(logrus.Fields{
 		"app":       "openmatch",
-		"component": "app.frontend.frontend_service",
+		"component": "app.frontend",
 	})
+	mTicketsCreated = monitoring.Counter("frontend/tickets_created", "tickets created")
```
golangcibot: `mTicketsCreated` is a global variable (from `gochecknoglobals`)

internal/app/frontend/frontend_service.go
```diff
@@ -36,10 +37,12 @@ type frontendService struct {
...
+	mTicketsDeleted = monitoring.Counter("frontend/tickets_deleted", "tickets deleted")
```
golangcibot: `mTicketsDeleted` is a global variable (from `gochecknoglobals`)

PR #644, opened 2019-07-11
Add TLS support to the Helm chart.

jeremyje:
This change allows someone to deploy Open Match with TLS encryption.

This mode will not be enabled by default (for now) because of the added configuration complexity. The demo and swaggerui will still serve on HTTP (because there's no compelling reason to do HTTPS for them and they don't use the `internal/rpc/` package for serving). They are however communicating to their backends using HTTPS.

### DIFF
--- a/install/helm/open-match/templates/backend.yaml
+++ b/install/helm/open-match/templates/backend.yaml
@@ -71,5 +71,5 @@ spec:
           containerPort: {{ .Values.openmatch.backend.grpc.port }}
         - name: http
           containerPort: {{ .Values.openmatch.backend.http.port }}
-        {{- include "kubernetes.probe" (dict "port" .Values.openmatch.backend.http.port) | nindent 8 }}
+        {{- include "kubernetes.probe" (dict "port" .Values.openmatch.backend.http.port "isHTTPS" .Values.openmatch.tls.enabled) | nindent 8 }}
 {{- end }}
--- a/internal/rpc/clients.go
+++ b/internal/rpc/clients.go
@@ -38,7 +38,9 @@ import (
 
 const (
 	configNameEnableRPCLogging = "logging.rpc"
-	configNameEnableMetrics    = "monitoring.prometheus.enabled"
+	configNameEnableMetrics    = "monitoring.prometheus.enable"
+	// configNameClientTrustedCertificatePath is the same as the root CA cert that the server trusts.
+	configNameClientTrustedCertificatePath = configNameServerRootCertificatePath
 )
 
 var (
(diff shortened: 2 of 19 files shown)


install/helm/open-match/templates/backend.yaml
```diff
@@ -71,5 +71,5 @@ spec:
           containerPort: {{ .Values.openmatch.backend.grpc.port }}
         - name: http
           containerPort: {{ .Values.openmatch.backend.http.port }}
-        {{- include "kubernetes.probe" (dict "port" .Values.openmatch.backend.http.port) | nindent 8 }}
+        {{- include "kubernetes.probe" (dict "port" .Values.openmatch.backend.http.port "https" .Values.openmatch.tls.enabled) | nindent 8 }}
```
yfei1: Maybe something like
```suggestion
        {{- include "kubernetes.probe" (dict "port" .Values.openmatch.backend.http.port "isHTTPS" .Values.openmatch.tls.enabled) | nindent 8 }}
```

internal/rpc/clients.go
```diff
@@ -39,6 +39,8 @@ import (
 const (
 	configNameEnableRPCLogging = "logging.rpc"
 	configNameEnableMetrics    = "monitoring.prometheus.enabled"
+	// configNameClientTrustedCertificatePath is the same as the root CA cert that the server trusts.
+	configNameClientTrustedCertificatePath = configNameServerRootCertificateFile
```
yfei1:
```suggestion
	configNameClientTrustedCertificatePath = configNameServerRootCertificatePath
```

internal/rpc/clients.go
```diff
@@ -71,14 +73,14 @@ func GRPCClientFromConfig(cfg config.View, prefix string) (*grpc.ClientConn, err
 	}
 
 	// If TLS support is enabled in the config, fill in the trusted certificates for decrpting server certificate.
-	if cfg.GetBool("tls.enabled") {
-		_, err := os.Stat(cfg.GetString("tls.trustedCertificatePath"))
+	if cfg.GetString(configNameClientTrustedCertificatePath) != "" {
```
yfei1: cfg.IsSet(configNameClientTrustedCertificatePath)

tools/certgen/certgen.go
```diff
@@ -27,13 +27,27 @@ import (
 )
 
 var (
+	// serviceAddresses is a list of all the HTTP hostname:port combinations for generating a TLS certificate.
+	// It appears that gRPC does not care about validating the port number so we only add the HTTP addresses here.
+	serviceAddressList = []string{
```
yfei1: I'm working on a change to add the hostname to the config file. Do you mind adding a TODO at here so that we could generate the certificates based on values from the config in the future.

internal/rpc/clients.go
jeremyje: IsSet won't work. That's if the key doesn't exist. If it's "" it'll try to do TLS mode which will be bad.

tools/certgen/certgen.go
jeremyje: I think the more sane approach is to specify the flag value if you're going to do auto generation.

Issue #623, opened 2019-07-08
Simplify Configuration


jeremyje:
The configuration is pretty large. There's a few TODOs related to it as well.

We should fix them:
* Unify pullPolicy for images.
* Add relevant timeouts to RPC calls.
* Use constants for configuration aliases.
* Stop using environment variables for Redis connection.

yfei1:
Aiming to close this issue next week. Also found something that might be useful in helm3
https://github.com/helm/community/blob/master/helm-v3/001-charts.md#library-charts

PR #647, opened 2019-07-12
Make config names constants. (#623)

jeremyje:
Make the configuration names constants so that there's 1 label.
Tests are excluded if it would cause the config name to be exported.

### DIFF
--- a/internal/statestore/redis.go
+++ b/internal/statestore/redis.go
@@ -31,6 +31,26 @@ import (
 	"open-match.dev/open-match/pkg/pb"
 )
 
+const (
+	configNameRedisUser                   = "redis.user"
+	configNameRedisPassword               = "redis.password"
+	configNameRedisHostname               = "redis.hostname"
+	configNameRedisPort                   = "redis.port"
+	configNameRedisPoolIdleTimeout        = "redis.pool.idleTimeout"
+	configNameRedisPoolMaxActive          = "redis.pool.maxActive"
+	configNameRedisPoolMaxIdle            = "redis.pool.maxIdle"
+	configNameRedisPoolHealthCheckTimeout = "redis.pool.healthCheckTimeout"
+
+	configNameRedisExpiration     = "redis.expiration"
+	configNameRedisIgnoreListsTTL = "redis.ignoreLists.ttl"
+
+	configNameBackoffInitialInterval = "backoff.initialInterval"
+	configNameBackoffRandFactor      = "backoff.randFactor"
+	configNameBackoffMultiplier      = "backoff.multiplier"
+	configNameBackoffMaxInterval     = "backoff.maxInterval"
+	configNameBackoffMaxElapsedTime  = "backoff.maxElapsedTime"
+)
+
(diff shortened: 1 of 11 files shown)


internal/statestore/redis.go
```diff
@@ -31,6 +31,26 @@ import (
...
+	configNameRedisPoolIdleTimeout        = "redis.pool.idleTimeout"
+	configNameRedisPoolMaxActive          = "redis.pool.maxActive"
+	configNameRedisPoolMaxIdle            = "redis.pool.maxIdle"
+	configNameRedisPoolHealthCheckTimeout = "redis.pool.healthCheckTimeout"
+
+	configNameRedisExpiration     = "redis.expiration"
+	configNameRedisIgnoreListsTTL = "redis.ignoreLists.ttl"
+
+	configNameBackoffInitialInterval = "backoff.initialInterval"
+	configNameBackoffRandFactor      = "backoff.randFactor"
```
golangcibot: U1000: const `configNameBackoffRandFactor` is unused (from `unused`)

internal/statestore/redis.go
```diff
@@ -31,6 +31,26 @@ import (
...
+	configNameBackoffMultiplier      = "backoff.multiplier"
```
golangcibot: U1000: const `configNameBackoffMultiplier` is unused (from `unused`)

internal/statestore/redis.go
```diff
@@ -31,6 +31,26 @@ import (
...
+	configNameBackoffMaxInterval     = "backoff.maxInterval"
```
golangcibot: U1000: const `configNameBackoffMaxInterval` is unused (from `unused`)

internal/statestore/redis.go
```diff
@@ -31,6 +31,26 @@ import (
...
+	configNameBackoffMaxElapsedTime  = "backoff.maxElapsedTime"
```
golangcibot: U1000: const `configNameBackoffMaxElapsedTime` is unused (from `unused`)

Laremere:
From a coding style standpoint, I'm not a fan of this change.  It adds a layer of indirection, and config values are generally consumed once (or twice in adjacent lines if it checks the value being set);  So I don't this improving code correctness.

I don't care very strongly either way though, but that's my thoughts.

Laremere:
Oh, and not exporting the values because of tests seems off to me as well:  The main reason you'd use constants for values like this is to ensure consistency between different call sites, and you're excluding a whole set of them.  You're only ensuring consistency between calls within the same package, where it's easiest to see consistency.

jeremyje:
> Oh, and not exporting the values because of tests seems off to me as well: The main reason you'd use constants for values like this is to ensure consistency between different call sites, and you're excluding a whole set of them. You're only ensuring consistency between calls within the same package, where it's easiest to see consistency.

I wrote this change because the configuration had fallen out quite a bit. I'll shelve this for now and make a change to reduce the inconsistencies instead.

PR #648, opened 2019-07-12
Fix metrics flushing, add OC agent, and refactor multi-closing.

jeremyje:
This change originally was to add OC agent but I noticed a bug with how we flush monitoring and...
This change refactors MultiCloser into the utils package.

### DIFF
--- a/internal/util/multiclose.go
+++ b/internal/util/multiclose.go
@@ -1,0 +1,71 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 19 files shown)


internal/util/multiclose.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"sync"
+
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	logger = logrus.WithFields(logrus.Fields{
+		"app":       "openmatch",
+		"component": "util",
```
sawagh: nit - given util will have a wide range of sub components, may be useful for this to imply util.multiclose?

jeremyje: I think `util` is fine for now. The content of logs should sufficiently determine which file is relevant.

PR #649, opened 2019-07-12
Merge logrus loggers

jeremyje:
We are moving to a model where there's 1 logger per package. Some extra loggers are ok if they make sense. Like the rpc package has a client and server logger.

This change does not include the logger merge for monitoring package. That's in https://github.com/googleforgames/open-match/pull/648

### DIFF
--- a/internal/rpc/tls_server.go
+++ b/internal/rpc/tls_server.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"open-match.dev/open-match/internal/monitoring"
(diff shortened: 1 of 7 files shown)


internal/rpc/server.go
```diff
@@ -225,20 +225,20 @@ func startServingIndefinitely(params *ServerParams) (func(), func(), error) {
 	// Start serving traffic.
 	waitForStart, err := s.Start(params)
 	if err != nil {
-		serverLogger.WithFields(logrus.Fields{
+		logger.WithFields(logrus.Fields{
```
golangcibot: undeclared name: `logger` (from `typecheck`)

internal/rpc/tls_server.go
```diff
@@ -96,10 +88,10 @@ func (s *tlsServer) start(params *ServerParams) (func(), error) {
 	serverStartWaiter.Add(1)
 	go func() {
 		serverStartWaiter.Done()
-		tlsServerLogger.Infof("Serving gRPC: %s", s.grpcLh.AddrString())
+		serverLogger.Infof("Serving gRPC-TLS: %s", s.grpcLh.AddrString())
```
sawagh: Any reason to use different loggers for server.go (logger) and for the tls / insecure files (serverlogger) ?

jeremyje: I didn't submit that change to upstream by accident. It's the way you think it should be.

PR #660, opened 2019-07-16
Open local TCP ports in tests to prevent firewall dialogs.

jeremyje:
In unit tests, Open Match serves on random TCP ports that accept traffic anywhere. This causes a lot of firewall alerts to popup when running the unit tests. This change causes Open Match to serve on `localhost:$PORT` when running in unit test mode. (Really when a random port is used.)

### DIFF
--- a/tools/reaper/internal/internal_test.go
+++ b/tools/reaper/internal/internal_test.go
@@ -28,7 +28,7 @@ import (
 )
 
 func TestServeAndClose(t *testing.T) {
-	conn, err := net.Listen("tcp", ":0")
+	conn, err := net.Listen("tcp", "localhost:0")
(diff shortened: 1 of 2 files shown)


tools/reaper/internal/internal_test.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 func TestServeAndClose(t *testing.T) {
-	conn, err := net.Listen("tcp", ":0")
```
yfei1: Just curious, does it use 127.0.0.1 instead without specifying locahost?

jeremyje: No it uses 0.0.0.0 which means to serve everyone.

PR #662, opened 2019-07-16
Add /help, /configz, /debug/* pages.

jeremyje:


internal/telemetry/probe.go
```diff
@@ -30,6 +32,11 @@ const (
 	healthStateUnhealthy  = int32(2)
 )
 
+var (
+	successKey       = tag.MustNewKey("success")
+	mReadinessProbes = Counter("health/readiness", "readiness probes", successKey)
```
golangcibot: `mReadinessProbes` is a global variable (from `gochecknoglobals`)

internal/telemetry/configz.go
```diff
@@ -0,0 +1,94 @@
...
+{{ range $key, $value := . }}
+<tr><td>{{ $value.Key }}</td><td>{{ $value.Value }}</td></tr>
+{{ end }}
+</table>
+</body>
+`
+)
+
+var (
+	configPageTemplate = template.Must(template.New(configZTemplateName).Parse(configPage))
```
golangcibot: `configPageTemplate` is a global variable (from `gochecknoglobals`)

internal/telemetry/configz.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+type configZValue struct {
+	Key   string
+	Value interface{}
+}
+
+// ServeHTTP serves the /configz endpoint that allows a user to view the configuration of the server.
+func (cz *configz) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	vCfg, ok := cz.cfg.(*viper.Viper)
```
yfei1: Can we stick to the `cfg` naming?

internal/telemetry/probe.go
```diff
@@ -51,9 +58,11 @@ func (sp *statefulProbe) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 					logger.WithError(err).Warningf("%s health check failed. The server will terminate if this continues to happen.", HealthCheckEndpoint)
 				}
 				http.Error(w, err.Error(), http.StatusServiceUnavailable)
+				IncrementCounter(req.Context(), mReadinessProbes, tag.Insert(successKey, "false"))
 				return
 			}
 		}
+		IncrementCounter(req.Context(), mReadinessProbes, tag.Insert(successKey, "true"))
 		old := atomic.SwapInt32(sp.healthState, healthStateHealthy)
```
yfei1: It looks like ServeHTTP will respond "ok" even if there is no probe handler defined for a server. It is the expected behavior?

jeremyje: If the readiness probe list is empty the behavior is to return ok. That's intentional, otherwise Kubernetes will kill the pod after a few failed calls.

PR #663, opened 2019-07-16
Instrument HTTP clients and servers.

jeremyje:
This change introduces instrumentation for HTTP servers and clients.


internal/rpc/server.go
```diff
@@ -258,6 +260,16 @@ func MustServeForever(params *ServerParams) {
 	serveUntilKilledFunc()
 }
 
+func instrumentHTTPHandler(handler http.Handler, params *ServerParams) http.Handler {
```
sawagh: null check for params?

jeremyje: Not needed. Don't pass null like that and it's required for it to work.

PR #665, opened 2019-07-16
Merge all docker RUN commands to improve caching of CI image.

jeremyje:
This is an attempt to improve the Kaniko cache step by having having fewer layers in hopes it reduces the number of round trips to verify that the cache is still good.


jeremyje:
There's basically no improvement.

PR #668, opened 2019-07-17
Remove cloudbuild artifacts

yfei1:
The binaries generated in cloudbuild can only be used under kubernetes environment as they requires reading `match_maker.yaml` config from `/app/config` directory - which get created by volume mount. Therefore, uploading these artifacts to our gcs bucket only makes our CI slower. This commit removes the step for storing the binary artifacts to speed up our CI.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -168,12 +168,6 @@ artifacts:
     objects:
         location: gs://open-match-build-artifacts/output/
         paths:
-            - cmd/backend/backend
-            - cmd/frontend/frontend
-            - cmd/mmlogic/mmlogic
-            - cmd/synchronizer/synchronizer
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -168,24 +168,13 @@ artifacts:
     objects:
         location: gs://open-match-build-artifacts/output/
         paths:
-            - cmd/backend/backend
-            - cmd/frontend/frontend
-            - cmd/mmlogic/mmlogic
-            - cmd/synchronizer/synchronizer
-            - cmd/minimatch/minimatch
```
jeremyje: This is ok for now but I may bring back minimatch and certgen later with their Win64 and OSX variants when we have a multi language support story.

PR #669, opened 2019-07-17
Let config.Read() reads from two files

yfei1:
This commit does two things:
1. config.Read() already provides verbose error logs and is used once per service so I cleaned up the logging from service level to reduce the log spam.
2. Let config.Read() reads from `matchmaker_config.yaml` and `global_config.yaml` such that we are able to throw monitoring and tls configs into `global_config.yaml` and keep `matchmaker_config.yaml` clean

### DIFF
--- a/examples/demo/main.go
+++ b/examples/demo/main.go
@@ -19,9 +19,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
+
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/websocket"
-	"net/http"
 	"open-match.dev/open-match/examples/demo/bytesub"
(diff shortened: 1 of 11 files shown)


examples/demo/main.go
```diff
@@ -41,12 +42,8 @@ var (
 )
 
 func main() {
-	cfg, err := config.Read()
-	if err != nil {
-		logger.WithFields(logrus.Fields{
-			"error": err.Error(),
-		}).Fatalf("cannot read configuration.")
-	}
+	cfg := config.Read()
```
jeremyje: This causes a significant reliability issue where pushing a bad config will take down your entire cluster. On startup it makes sense to fail because you cannot read the config but existing servers should report a read failure and continue running with the old configuration.

PR #673, opened 2019-07-18
Add build/cmd to Makefile

Laremere:
This is part of the larger goal of simplifying and speeding up builds.

General strategy here:
1. Add make build/cmd target, to build what is currently in /cmd. <- this PR
2. In the basebuild Dockerfile, run make build/cmd, and replace the seperate Dockerfiles for the services in cmd/ with one Dockerfile (with an arg for which service) which copies from build/cmd/<servicename>/ to the Dockerfile and runs it.
3. Reconcile the Docker images not included in the /cmd to follow this pattern as well (in individual PRs.)
4. Clean up redundant ways to build things.
5. (If it improves speed) add local build variation for faster builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -791,6 +791,28 @@ lint: fmt vet golangci lint-chart terraform-lint
 
 assets: all-protos tls-certs third_party/ build/chart/
 
+# CMDS is a list of all folders in cmd/
+CMDS = $(notdir $(wildcard cmd/*))
+CMDS_BUILD_FOLDERS = $(foreach CMD,$(CMDS),build/cmd/$(CMD))
+
+build/cmd: $(CMDS_BUILD_FOLDERS)
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -791,6 +791,28 @@ lint: fmt vet golangci lint-chart terraform-lint
...
+
+build/cmd/%/BUILD_PHONY: all-protos
+	mkdir -p $(BUILD_DIR)/cmd/$*
+	$(GO) build -o $(BUILD_DIR)/cmd/$*/$* open-match.dev/open-match/cmd/$*
+
+# Default is that nothing needs to be copied into the direcotry
+build/cmd/%/COPY_PHONY:
+	#
+
+build/cmd/swaggerui/COPY_PHONY:
```
jeremyje: What's COPY_PHONY? Typically for these rules there's an actual file backing up the target path so Make knows skip building if it already exists.

Laremere: Added a comment explaining what's going on.

PR #674, opened 2019-07-18
Make load testing upload test results to GCS

yfei1:
This commits enabled auto uploading the stress test result to the GCS bucket with the workload identity authorized by `stress-test-uploader@open-match-build.iam.gserviceaccount.com` service account.


jeremyje:
I think you'll need to do the following:

 1. Enable Workload Identity on the cluster.
 1. Create a IAM Service Account with the policy bindings to write to GCS. You'll need to do this in the open-match-build terraform.
 1. Create a k8s ServiceAccount and have the GCS upload agent run as it.
 1. Add an IAM policy binding where the k8s ServiceAccount can authenticate as the IAM service account.

PR #676, opened 2019-07-22
Use templated Dockerfile and make for cmd images

Laremere:
This leads up to being able to swap out the standard dockerfile for one which uses a locally built binary.  Also is just cleaner as there is less redundancy across dockerfiles.

Disables cgo for builds, as it doesn't work with distroless.

Stops relying on phony protos target which causes extra rebuilds of all the protos.  (Using variable expansion should fix this issue in a seperate PR)

All commands are now named run, because ARG arbitrarily doesn't work for ENTRYPOINT.


Makefile
```diff
@@ -803,9 +809,10 @@ build/cmd: $(CMDS_BUILD_FOLDERS)
 # files to be included in the image.
 $(CMDS_BUILD_FOLDERS): build/cmd/%: build/cmd/%/BUILD_PHONY build/cmd/%/COPY_PHONY
 
-build/cmd/%/BUILD_PHONY: all-protos
+
+build/cmd/%/BUILD_PHONY:
 	mkdir -p $(BUILD_DIR)/cmd/$*
-	$(GO) build -o $(BUILD_DIR)/cmd/$*/$* open-match.dev/open-match/cmd/$*
+	CGO_ENABLED=0 $(GO) build -a -installsuffix cgo -o $(BUILD_DIR)/cmd/$*/run open-match.dev/open-match/cmd/$*
```
jeremyje: At this point does the `GO_BUILD_COMMAND` macro go away?

Dockerfile.cmd
```diff
@@ -14,20 +14,23 @@
...
+ARG IMAGE_TITLE
+
+RUN make "build/cmd/${IMAGE_TITLE}"
 
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /app/
-COPY --from=builder --chown=nonroot /go/src/open-match.dev/open-match/cmd/minimatch/minimatch /app/
 
-ENTRYPOINT ["/app/minimatch"]
+COPY --from=builder --chown=nonroot "/go/src/open-match.dev/open-match/build/cmd/${IMAGE_TITLE}/" "/app/"
```
jeremyje: Is there a way to not make this `run`?

Makefile
Laremere: Once other pieces are in cmd, the old commands can be cleaned up and GO_BUILD_COMMAND can be removed.

Dockerfile.cmd
Laremere: This seems to be the most reasonable way.  Entrypoint doesn't use parameters from the build.

PR #687, opened 2019-07-24
Replace context.Background() in tests to prepare for multi-tenancy.

jeremyje:
Calls to Open Match may have HTTP headers or gRPC metadata now and for that to work each test will produce randomized headers to confirm that calls are made appropriately.

### DIFF
--- a/internal/util/testing/context.go
+++ b/internal/util/testing/context.go
@@ -12,31 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package testing provides utility methods for testing.
 package testing
 
 import (
+	"context"
 	"testing"
-
-	"github.com/spf13/viper"
-	"github.com/stretchr/testify/assert"
-	"open-match.dev/open-match/internal/statestore"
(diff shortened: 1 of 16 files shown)


internal/util/testing/context.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"context"
+	"testing"
+)
+
+type contextTestKey string
+
+// NewContextForUnitTest returns a context appropriate for a service test that is not in test/e2e/.
+func NewContextForUnitTest(t *testing.T) context.Context {
```
Laremere: Naming Suggestion: TestingContext().

Also, maybe with a shorter name you can do:
```
			ctx, cancel := context.WithCancel(utilTesting.TestingContext())
```

Instead of:
```
			ctx := utilTesting.NewContextForUnitTest(t)
			ctx, cancel := context.WithCancel(ctx)
```
In a few places.

jeremyje: Renamed to NewContext(). In golang they discourage "stuttering" names. So `utilTesting.NewContext()` seems like the best here.

PR #688, opened 2019-07-24
Add HTTP client logging support and flags for test.

jeremyje:
```yaml
logging:
  rpc: true
```
will now enable HTTP client logging.

Also there's now `--test_only_rpc_logging=true` and `--test_only_metrics=true` for running tests with these options enabled.

### DIFF
--- a/internal/rpc/clients.go
+++ b/internal/rpc/clients.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/http/httputil"
 	"net/url"
 	"os"
 	"strings"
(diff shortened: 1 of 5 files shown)


internal/rpc/clients.go
```diff
@@ -39,7 +40,8 @@ import (
 )
 
 const (
-	configNameEnableRPCLogging = "logging.rpc"
+	// ConfigNameEnableRPCLogging is the config name for enabling RPC logging.
```
yfei1: Could we have an internal/rpc/const.go file for these package level constants?

jeremyje: I'll do that in my follow up change.

PR #694, opened 2019-07-25
Override containers' resource request values to have Open Match setup in GKE's default cluster

yfei1:
Redis requests 16Gi memory by default. When users wanna try out Open Match with GKE's default instance type n1-standard-1, GKE may unable to schedule Open Match's deployment as it run out of memory&CPU. This commit overrides Redis's memory request and change the cluster instance type in CI to n1-standard-2 to speed up CI and enable deploying Open Match in GKE's default cluster.


install/helm/open-match/templates/_helpers.tpl
```diff
@@ -45,7 +45,7 @@ imagePullPolicy: {{ .Values.image.pullPolicy }}
 resources:
   requests:
     memory: 100Mi
-    cpu: 100m
+    cpu: 50m
```
Laremere: This seems like a very small amount of cpu.  How did you arrive at this number?

yfei1: This only matter at cluster startup time. The process can still auto request more resource if it needs more quota.

See this link for more details https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/

yfei1:
> If we're switching to a persistent cluster, changing the gke size won't matter. The cluster seems reasonably sized for dev work.

The goal of this change is to speed up current CI and have a better story for first-time users of Open Match. We don't have a timeline for Prow yet so I think this change is valid.

jeremyje:
> The goal of this change is to speed up current CI and have a better story for first-time users of Open Match. We don't have a timeline for Prow yet so I think this change is valid.

Prow replaces `cloudbuild.yaml` to `config.yaml`. The formats of those files are different but I expect the `Makefile` to be relatively unchanged.

install/helm/open-match/templates/_helpers.tpl
Laremere: Still... where did the number come from?

yfei1: The idea is to use the minimum resource when setting up the environment. I was tempted to set it to 1m actually. 50m is a number that allows users having a functional Open Match deployed given GKE's default pool size with n1-standard-1 instance type.

yfei1:
> I think the question of whether this makes a difference with a persistent cluster remains?

The goal of this PR is to improve user experience. Renamed the title of this PR.

Issue #569, opened 2019-06-21
Stack Driver Metrics Export rate is too high.


jeremyje:
2019/06/21 19:29:52 Failed to export to Stackdriver: rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: One or more points were written more frequently than the maximum sampling period configured for the metric. {Metric: custom.googleapis.com/opencensus/grpc.io/client/completed_rpcs, Timestamps: {Youngest Existing: '2019/06/21-12:29:46.439', New: '2019/06/21-12:29:51.439'}}: timeSeries[6-10]; One or more points were written more frequently than the maximum sampling period configured for the metric. {Metric: custom.googleapis.com/opencensus/grpc.io/client/received_bytes_per_rpc, Timestamps: {Youngest Existing: '2019/06/21-12:29:46.439', New: '2019/06/21-12:29:51.439'}}: timeSeries[11-13]; One or more points were written more frequently than the maximum sampling period configured for the metric. {Metric: custom.googleapis.com/opencensus/grpc.io/client/received_messages_per_rpc, Timestamps: {Youngest Existing: '2019/06/21-12:29:46.439', New: '2019/06/21-12:29:51.439'}}: timeSeries[14-16]; One or more points were written more frequently than the maximum sampling period configured for the metric. {Metric: custom.googleapis.com/opencensus/grpc.io/client/roundtrip_latency, Timestamps: {Youngest Existing: '2019/06/21-12:29:46.439', New: '2019/06/21-12:29:51.439'}}: timeSeries[17-19]; One or more points were written more frequently than the maximum sampling period configured for the metric. {Metric: custom.googleapis.com/opencensus/grpc.io/client/sent_bytes_per_rpc, Timestamps: {Youngest Existing: '2019/06/21-12:29:46.439', New: '2019/06/21-12:29:51.439'}}: timeSeries[0-2]; One or more points were written more frequently than the maximum sampling period configured for the metric. {Metric: custom.googleapis.com/opencensus/grpc.io/client/sent_messages_per_rpc, Timestamps: {Youngest Existing: '2019/06/21-12:29:46.439', New: '2019/06/21-12:29:51.439'}}: timeSeries[3-5]"

yfei1:
Investigated this issue a bit. Found the sentence below from [Creating custom metrics in GCP](https://cloud.google.com/monitoring/custom-metrics/creating-metrics)
[...]

PR #698, opened 2019-07-29
Unify server setup and call log configuration (#695)

yfei1:
This commits is a hotfix for the stackdriver bug #569. It also enables the log trace config in Open Match. One thing to note is #695 is based on the new helm chart that we will be using in v.7 so I made [a small twist](https://github.com/yfei1/open-match/blob/55e8793898011471469c2b1d155fe98df14c6b4f/install/helm/open-match/values.yaml#L138) in `values.yaml` file to make the go template works for v0.6. This workaround will go away in v.7.


install/helm/open-match/values.yaml
```diff
@@ -135,7 +135,7 @@ openmatch:
       matchmaker_config.yaml:
         logging:
           level: debug
-          format: text
+          format: "{{ternary \"stackdriver\" \"text\" .Values.openmatch.telemetry.stackdriver.enabled}}"
```
sawagh: How to interpret this? how does this get set?

internal/app/appmain.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		logger.WithFields(logrus.Fields{
+			"error": err.Error(),
+		}).Fatalf("cannot construct server.")
+	}
+
+	if err := bindService(p, cfg); err != nil {
+		logger.WithFields(logrus.Fields{
+			"error": err.Error(),
+		}).Fatalf("failed to bind backend service.")
```
sawagh: add the service name here than just backend service.

install/helm/open-match/values.yaml
yfei1: This calls the ternary function defined in the gotpl package - http://masterminds.github.io/sprig/defaults.html#ternary. However, helm does not allow templating in the values.yaml file so what we does in v.6 is to quote the templates and pass the strings to the `tpl` function defined in om-configmap.yaml, which makes the `values.yaml` file cumbersome and hard to modify - I have to quote and escape the template to make it works.

This restriction will go away in v0.7 as we introduced the new helm chart.

internal/app/appmain.go
yfei1: Done.

PR #703, opened 2019-07-30
Replace all-proto with variables containing all proto files

Laremere:
This moves the all-proto target from being phony, to containing a concrete lists of targets.  This means other targets can depend on $(ALL_PROTO) without becoming phony itself.


yfei1:
Having the target from being phony is an improvement, but using variable as the target makes the Makefile less readable. Is there any Makefile conventions/Best practices that we could follow?

Laremere:
> Having the target from being phony is an improvement, but using variable as the target makes the Makefile less readable. Is there any Makefile conventions/Best practices that we could follow?

This seems to be a reasonably idiomatic method: http://web.mit.edu/gnu/doc/html/make_2.html

I don't know the conventions very well though.

PR #706, opened 2019-07-30
Let end-to-end tests run in-cluster

yfei1:
This commit change service.portType from LoadBalancer to ClusterIP to keep CI from wasting time in creating Load Balancers. It also updates the end-to-end config in `cluster.go` file such that we can still have the end-to-end in-cluster tests running in the CI. Since we are using unique namespace per CI run, and `rest.InClusterConfig` does not know the namespace of the service it lives in, I have a workaround in the `e2e-job.yaml` that pass in the namespace of the current release to the e2e job such that `cluster.go` can get the namespace from [here](https://github.com/googleforgames/open-match/blob/e6d470a6f0e6b5ed0f10649e45660d909c2dfbd5/internal/testing/e2e/cluster.go#L183).


internal/testing/e2e/cluster.go
```diff
@@ -170,35 +166,20 @@ func fileExists(name string) bool {
...
-	if !fileExists(kubeconfig) && fileExists(kubeconfigFromEnv) {
-		kubeconfig = kubeconfigFromEnv
-	}
-	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+func createZygote(m *testing.M) (OM, error) {
+	// creates the in-cluster config
+	kubeconfig, err := rest.InClusterConfig()
 	if err != nil {
-		return nil, errors.Wrapf(err, "building Kubernetes config from flags %s failed", kubeconfig)
+		panic(err.Error())
```
sawagh: log.Fatalf?

internal/testing/e2e/cluster.go
```diff
@@ -170,35 +166,20 @@ func fileExists(name string) bool {
...
 }
 
-func newClusterOM(kubeconfig string, kubeconfigFromEnv string, namespace string) (*clusterOM, error) {
-	if !fileExists(kubeconfig) && fileExists(kubeconfigFromEnv) {
-		kubeconfig = kubeconfigFromEnv
-	}
-	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+func createZygote(m *testing.M) (OM, error) {
+	// creates the in-cluster config
+	kubeconfig, err := rest.InClusterConfig()
```
sawagh: For my understanding - can we simplify this to in cluster config because now we are already in context of the created CI cluster?

install/helm/open-match/subcharts/open-match-test/templates/e2e-job.yaml
```diff
@@ -0,0 +1,46 @@
...
+  annotations: {{- include "openmatch.chartmeta" . | nindent 4 }}
+  labels:
+    app: {{ template "openmatch.name" . }}
+    component: e2e-job
+    release: {{ .Release.Name }}
+spec:
+  selector:
+    app: {{ template "openmatch.name" . }}
+    component: e2e-job
+  # Specifies the number of retries before marking this job failed. Defaults to 6
```
sawagh: 6? or 3?

install/helm/open-match/subcharts/open-match-test/templates/stress-master.yaml
```diff
@@ -26,7 +26,7 @@ spec:
   selector:
     app: {{ template "openmatch.name" . }}
     component: locust-master
-  type: {{ .Values.stresstest.portType }}
+  type: {{ coalesce .Values.global.kubernetes.service.portType .Values.stresstest.portType }}
```
sawagh: so value specified in stresstest will trump the global default (if sepcified?)

install/helm/open-match/values.yaml
```diff
@@ -19,35 +19,35 @@
...
   httpPort: 51500
-  portType: LoadBalancer
+  portType: ClusterIP
   replicas: 3
   image: openmatch-swaggerui
 mmlogic: &mmlogic
   hostName: om-mmlogic
   grpcPort: 50503
   httpPort: 51503
-  portType: LoadBalancer
```
sawagh: in the Helm chart documentation, did we specify how to actually switch between these port types? I do see folks needing the LB type in their E2E and would prefer not requiring the understanding of Helm chart / subchart value coalescing to achieve that.

install/helm/open-match/subcharts/open-match-test/templates/e2e-job.yaml
yfei1: Oh I mean default is 6 if we don't specify the number of retries at here

internal/testing/e2e/cluster.go
yfei1: I am using the in cluster config by calling `rest.InClusterConfig()`.

install/helm/open-match/subcharts/open-match-test/templates/stress-master.yaml
yfei1: `coalesce` mean pick the first value that is not empty, so value specified in stresstest will be overrided if we have `.Values.global.kubernetes.service.portType` being set.

install/helm/open-match/values.yaml
yfei1: Yup, #713 already covered this part.

PR #708, opened 2019-07-31
Makefile bug fixes

yfei1:
This commit does the following things:
1. Fixed the selectors in `make proxy-xxx` commands
2. Fixed the hostname of the match function defined in `internal/testing/e2e/cluster.go` file
3. `make install-ci-chart` does not support end to end test with matchfunction and evaluator yet so disable the customize components for now
4. Let CI pushes the base-build image to the gcr repo since the in-cluster end-to-end test needs contexts in that image to run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -282,7 +282,7 @@ build-stress-test-images: build-stress-frontend-image
 
 # Include all-protos here so that all dependencies are guaranteed to be downloaded after the base image is created.
 # This is important so that the repository does not have any mutations while building individual images.
-build-base-build-image: docker all-protos
+build-base-build-image: docker $(ALL_PROTOS)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -282,8 +282,8 @@ build-stress-test-images: build-stress-frontend-image
 
 # Include all-protos here so that all dependencies are guaranteed to be downloaded after the base image is created.
 # This is important so that the repository does not have any mutations while building individual images.
-build-base-build-image: docker $(ALL_PROTOS)
-	docker build -f Dockerfile.base-build -t open-match-base-build .
+build-base-build-image: docker all-protos
```
Laremere: $(ALL_PROTOS) should be correct.  Is there a problem with it?

PR #705, opened 2019-07-30
Refactor ci-reap-cluster to ci-reap-namespaces

yfei1:
This commit together with #701 removes the `Test: Create Cluster` step from cloudbuild to speed up CI. ci-reaper is now responsible for deleting staled namespaces to make sure resources in a broken build will be freed

PR #701, opened 2019-07-29
Let CI runs in one cluster under unique namespaces

yfei1:

PR #709, opened 2019-07-31
Reflects IAM role changes in terraform config

yfei1:
1. Rename Cluster Reaper to a more general name - CI Reaper.
2. Grant Cluster Reaper `container.namespaces` permissions in case we need them in Cloud Run

Will apply and update the tfstate file once this commit get approved


sawagh:
Does this still reap clusters at all or only reaps namespaces? If it does reap clusters too - lets still keep it cluster-reaper? still more informative than ci-reaper :)

yfei1:
> Does this still reap clusters at all or only reaps namespaces? If it does reap clusters too - lets still keep it cluster-reaper? still more informative than ci-reaper :)

#705 and #701 let our CI runs under unique namespaces in the `open-match-ci` cluster, so the reaper is now only responsible for deleting the staled namespaces. I changed the name to ci-reaper since I'm not sure if it role will change in the future.

PR #713, opened 2019-08-01
Helm README

yfei1:


install/helm/README.md
```diff
@@ -0,0 +1,64 @@
+### Open Match Helm Chart Templates
+This directory contains the [helm](https://helm.sh/ "helm") chart templates used to customize and deploy Open Match at ease.
```
sawagh: remove _at ease_

install/helm/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+Templates under the `templates/` directory are for the core components in Open Match - e.g. backend, frontend, mmlogic, synchronizor, some security policies, and configmaps are defined under this folder.
+
+Open Match also provides templates for optional components that are disabled by default under the `subcharts/` directory.
+1. `open-match-demo` contains the chart definitions of an example director.
```
sawagh: since we are using the word template elsewhere below, this can be:
_contains the template for a sample director_

install/helm/README.md
```diff
@@ -0,0 +1,64 @@
...
++ - mysecondice
++ - ...
+
+# Overrides spec.type of a specific Kubernetes Service
+# Equivalent helm cli flag --set swaggerui.portType=LoadBalancer
+swaggerui:
+-  portType: ClusterIP
++  portType: LoadBalancer
+
+# Overrides spec.type of all Open Match components
```
sawagh: Are these values shared across subcharts? So which all components does this service portType impact? All across all core charts and subcharts? 

Also, does synchronizer use this? if not, might be best to call out which components can be configured using this.

install/helm/README.md
```diff
@@ -0,0 +1,64 @@
...
+# Equivalent helm cli flag --set open-match-customize.image.registry=[XXX],open-match-customize.image.tag=[XXX]
+open-match-customize:
+  enabled: true
++   image:
++     registry: [YOUR_REGISTRY_URL]
++     tag: [YOUR_IMAGE_TAG]
++   function:
++     image: [YOUR_MATCHFUNCTION_IMAGE_NAME]
++   evaluator:
++     image: [YOUR_EVALUATOR_IMAGE_NAME]
```
sawagh: Add comment for rpc logging / enable debug logging please? i think a lot of folks need that. also, same question - would changing the value there impact all subcharts?

PR #714, opened 2019-08-01
Autogenerate image commands based on a single list

Laremere:
With this change, anything added to the cmd/ folder is automatically
made into an image and included the image commands.  Additional
images are also included.  This does remove some commands that are
for pushing specific sets of images, but it seems rare/never (asking yfei1
and sawagh) that anyone is actually using these commands.

Includes some commenting to hopefully alleviate the magic being added.


sawagh:
_This does remove some commands that are for pushing specific sets of images, but it seems rare/never (asking yfei1 and sawagh) that anyone is actually using these commands._

I am not so concerned here as i dont foresee just pushing the sets (all sample images only for instance). Ok losing that ability so long as we have the ability to still push only one image - which this change seems to retain.

Makefile
```diff
@@ -202,75 +203,39 @@ SWAGGER_JSON_DOCS = api/frontend.swagger.json api/backend.swagger.json api/mmlog
 
 ALL_PROTOS = $(GOLANG_PROTOS) $(SWAGGER_JSON_DOCS)
 
+# CMDS is a list of all folders in cmd/
+CMDS = $(notdir $(wildcard cmd/*))
```
sawagh: This assumes every folder under cmd will be build into an image? 
- Could in future there be a folder which is some cmd line tool that we dont wish to have an image for.
- While WIP, i create a folder, write code in it - still dont have my dockerfile etc., ready, will make all fail?

I like the approach of making the rest of the commands more concise - how about have a specified list of folders here that we wish to apply all the build / push / retag commands on - with a comment stating, any new component we add under cmd - just to be added to this list here to get all the build functionality?

Laremere: I actually moved this from further on in the file.  The logic that all cmds are images is from before this PR.  I think this is a safe assumption.  In response to your concerns:
- If we have command line tools, they can still be made using `make build/cmd/<name>` (this is what the building Dockerfile calls).  I think any tools we don't want to be images is unlikely, as even if it is usable outside of an image, it would still be useful to be able to run it inside a kube cluster using `kubectl run` or similar.  Especially because then people wouldn't need to build or install the command to use it.
- There is only one shared Dockerfile for images build from cmd. (Dockerfile.cmd in the root directory)  The code in cmd (following convention) is only a short main function.  In a sense, adding it to cmd is the final step in saying everything is ready.

I think it's best to assume regularity here, and if we really find that the regularity isn't correct, we adjust from there.  It's always fairly easy to turn this into a concrete list, or use make's `filter-out` function to add exceptions.

PR #715, opened 2019-08-01
Remove unnecessary dependencies to speed up CI

yfei1:
This commit cleans up the Makefile and removes some duplicated dependencies from cloudbuild to speed up CI.


yfei1:
> Looks like CI failure in a PR to remove CI dependencies :)

Resolved by #717

PR #717, opened 2019-08-02
Skip using PodSecurityPolicy in CI runs

yfei1:
Kubernetes only allows having one om-podsecurtiypolicy per cluster. The CI will fail if we have multiple builds running at the same time with PSP being enabled.


yfei1:
> Can you add some description as to why we skip this in CI runs?

Done

PR #725, opened 2019-08-03
Fix proto file's go package.

### LINKED ISSUE #724: Fix proto's go package
Currently the proto package omits the open-match piece (just `pkg/pb` instead of `open-match.dev/open-match/pkg/pb`), and then it uses sed to fix the reference in the internal proto package.  This will be broken for other users too.  Find a way to have it generated to the correct place.  The documentation seems to not take go modules into account.

Laremere:
Fixes #724

Also removed redundant instruction to build messages.pb.go, and unused instruction to build message.pb.gw.go.


Makefile
```diff
@@ -642,24 +642,33 @@ gcp-apply-binauthz-policy: build/policies/binauthz.yaml
 
 all-protos: $(ALL_PROTOS)
 
+# The proto generator really wants to be run from the $GOPATH root, and doesn't
+# support methods for directing it to the correct location that's not the proto
+# file's location.  So instead put it in a tempororary directory, then move it
+# out.
 pkg/pb/%.pb.go: api/%.proto third_party/ build/toolchain/bin/protoc$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-grpc-gateway$(EXE_EXTENSION)
-	mkdir -p $(REPOSITORY_ROOT)/pkg/pb
+	mkdir -p $(REPOSITORY_ROOT)/build/prototmp
```
yfei1: mkdir -p $(REPOSITORY_ROOT)/build/prototmp $(REPOSITORY_ROOT)/pkg/pb

PR #732, opened 2019-08-08
Use constants for configuration aliases

yfei1:
This commit uses constants to replace string literals when referring fields in the config file. Open Match has a bunch of customizable settings in its config. Having these string literals in the code base make it quite hard to modify the `matchmaker_config.yaml` file. Considering we have a lot of config-related TODOs in v0.7, I use constants instead so that we could save some time for the future changes.


internal/app/mmlogic/mmlogic_service.go
```diff
@@ -88,11 +88,11 @@ func getPageSize(cfg config.View) int {
 		maxPageSize int = 10000
 	)
 
-	if !cfg.IsSet(name) {
+	if !cfg.IsSet(internal.StatestorePageSize) {
```
sawagh: One idea is to abbreviate this to just store? I ve seen this in previous teams - it communicates the intent - and is short. Thoughts?

internal/app/swaggerui/swaggerui.go
```diff
@@ -55,7 +56,7 @@ func serve(cfg config.View) {
 	mux := &http.ServeMux{}
 	closer := telemetry.Setup(mux, cfg)
 	defer closer()
-	port := cfg.GetInt("api.swaggerui.httpport")
+	port := cfg.GetInt(internal.SwaggerUIPrefix + internal.HTTPPortSuffix)
```
sawagh: Since we are using consts anyways, trying to append these seems odd. We might as well have each of these as first class constants? (SwaggerUIHTTPPort.. ). I understand there ll be many then - but they are scoped to a consts file and the rest of the usage is clean.

internal/app/synchronizer/evaluator_client.go
```diff
@@ -75,8 +76,12 @@ func (ec *evaluatorClient) initialize() error {
 		return nil
 	}
 
+	evaluatorHostName := ec.cfg.GetString(internal.EvaluatorPrefix + internal.HostNameSuffix)
```
sawagh: Similar comment on first class constants over string append. Elsewhere in the file too.

internal/consts.go
```diff
@@ -0,0 +1,50 @@
+package internal
```
Laremere: horrible package name.

Issue #739, opened 2019-08-15
API change: Unary to Stream


yfei1:
## Overview
Psyonix hit the 4MB maximum message size limit for gRPC when stress testing `backend.FetchMatches` call in release v0.6. We investigated this issue and decided to change the following APIs from unary calls to streaming.

## Motivation
Improves Open Match's scalibility 

## Impact
This change will affect most of the APIs that may hit the 4MB gRPC message size limit in the production environment.

__Users who have their director implemented with the v0.6 API__ may need to refactor the way of receiving messages.

__Users who choose to write their own mmf and evaluator harness__ may need to refactor their code in release v0.7.

__Users who are currently using Open Match's native mmf and evaluator harness__ won't be affected.

The APIs that need to be changed from unary into streaming are: 

1. `backend.FetchMatches`
2. `evaluator.Evaluate`
3. `matchfunction.Run`
4. `synchronizer.EvaluateProposals` - this is an internal API change

## Change Proto

- [x] `backend.FetchMatches`: FetchMatches will still have a 4MB request size limit yet will stream back its response instead after this change.
```diff
-rpc FetchMatches(FetchMatchesRequest) returns (FetchMatchesResponse) {
+rpc FetchMatches(FetchMatchesRequest) returns (stream FetchMatchesResponse) {
```
- [x] `evaluator.Evaluate` - This will be the bottleneck if the synchronizer is enabled and do the evaluation in batch.
```diff
-rpc Evaluate(EvaluateRequest) returns (EvaluateResponse) {
+rpc Evaluate(stream EvaluateRequest) returns (stream EvaluateResponse) {
```
- [x] `matchfunction.Run` - This might be a bottleneck if we have too many proposals being returned in one call
```diff
-rpc Run(RunRequest) returns (RunResponse) {
+rpc Run(RunRequest) returns (stream RunResponse) {
```
- [x] `synchronizer.EvaluateProposals` - This might be a bottleneck if we have too many proposals being evaluated in one call
```diff
-rpc EvaluateProposals(EvaluateProposalsRequest) returns (stream EvaluateProposalsResponse) {
+rpc EvaluateProposals(stream EvaluateProposalsRequest) returns (stream EvaluateProposalsResponse)
```

## How to reproduce the error
```
func TestGrpcLimit(t *testing.T) {
	om, closer := e2e.New(t)
	defer closer()
	fe := om.MustFrontendGRPC()
	be := om.MustBackendGRPC()
[...]

PR #740, opened 2019-08-15
Change mmf API from unary call to streaming

yfei1:
This commit lets `matchfunction.Run` stream back the response to bypass the grpc response limit.


internal/app/backend/backend_service.go
```diff
@@ -204,18 +204,32 @@ func matchesFromHTTPMMF(ctx context.Context, profile *pb.MatchProfile, client *h
...
-	pbResp := &pb.RunResponse{}
-	err = json.Unmarshal(body, pbResp)
-	if err != nil {
-		return nil, status.Errorf(codes.FailedPrecondition, "failed to unmarshal response body to response pb for profile %s: %s", profile.Name, err.Error())
+		err := dec.Decode(&item)
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, status.Errorf(codes.FailedPrecondition, "failed to decode RunResponse from JSON: %s", err.Error())
```
Laremere: FailedPrecondition doesn't seem like the correct error code here.  A failure here isn't likely to be a problem with system state which needs to be corrected. (eg, if this error is a timeout, that's not at issue here.)
I think these should all be Unavailable?
See https://github.com/grpc/grpc-go/blob/master/codes/codes.go

yfei1: I agree that this one should be Unavailable as it reflects the service is currently unavailable. For the second one, I changed it to directly return the error since it is from a service that we own. 

I think `FailedPrecondition` applies to the third case since the error most likely means the client wrote a bad mmf and should not retry until the system state has been explicitly fixed

internal/app/backend/backend_service.go
```diff
@@ -204,18 +204,32 @@ func matchesFromHTTPMMF(ctx context.Context, profile *pb.MatchProfile, client *h
...
-		return nil, status.Errorf(codes.FailedPrecondition, "failed to unmarshal response body to response pb for profile %s: %s", profile.Name, err.Error())
+		err := dec.Decode(&item)
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, status.Errorf(codes.Unavailable, "failed to read response from HTTP JSON stream: %s", err.Error())
+		}
+		if len(item.Error) != 0 {
+			return nil, item.Error
```
golangcibot: cannot use item.Error (variable of type map[string]interface{}) as error value in return statement: missing method Error (from `typecheck`)

yfei1:
See proposal #739 for the details.

PR #746, opened 2019-08-16
Suppress golangci error

yfei1:
Encountered a bug in golangci. I tried to revert changes file by file yet it is still unclear what is the reason. This commit suppresses the error from golangci in the Makefile to keep it from failing the CI. Golangci cli check is not crital as long as we have its log and the golangci bot so I prefer to suppress its error.

https://pantheon.corp.google.com/cloud-build/builds/628447a1-0601-4e4e-abbd-51fadfff70ec?project=open-match-build

https://github.com/golangci/golangci-lint/issues/484

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -71,7 +71,7 @@ linters-settings:
 
   govet:
     # report about shadowed variables
-    # check-shadowing: true
+    check-shadowing: true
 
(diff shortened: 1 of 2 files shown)


sawagh:
Please create an issue to track investigating / re-enabling this in future.

PR #750, opened 2019-08-16
Change synchronizer API from unary to bidirectional streaming

yfei1:
This commit changed synchronizer.EvaluateProposals from using unary gRPC call to bidirectional streaming. See proposal #739 for the details.


internal/app/backend/synchronizer_client.go
```diff
@@ -21,6 +22,11 @@ type synchronizerClient struct {
 	m            sync.Mutex
 }
 
+type evaluateProposalsResult struct {
```
golangcibot: `evaluateProposalsResult` is unused (from `deadcode`)

internal/app/backend/synchronizer_client.go
```diff
@@ -21,6 +22,11 @@ type synchronizerClient struct {
...
+	results []*pb.Match
```
golangcibot: `results` is unused (from `structcheck`)

internal/app/backend/synchronizer_client.go
```diff
@@ -21,6 +22,11 @@ type synchronizerClient struct {
...
+	err     error
```
golangcibot: `err` is unused (from `structcheck`)

internal/app/backend/synchronizer_client.go
```diff
@@ -58,15 +59,35 @@ func (sc *synchronizerClient) evaluate(ctx context.Context, id string, proposals
...
-		Matches: proposals})
-	if err != nil {
-		return nil, err
+	if len(proposals) == 0 {
+		return proposals, nil
+	}
+
+	stream, err := sc.synchronizer.EvaluateProposals(ctx)
+	for _, proposal := range proposals {
+		if err = stream.Send(&ipb.EvaluateProposalsRequest{Id: id, Match: proposal}); err != nil {
```
sawagh: need to close stream? or just error out?

internal/app/backend/synchronizer_client.go
```diff
@@ -58,15 +59,35 @@ func (sc *synchronizerClient) evaluate(ctx context.Context, id string, proposals
...
+			return nil, err
+		}
+	}
+	if err = stream.CloseSend(); err != nil {
+		logger.Errorf("failed to close the send stream: %s", err.Error())
+	}
+
+	results := []*pb.Match{}
```
sawagh: I see you following a separate pattern on the Evaluate interface - of sending over a separate goroutine and receiver blocking - any reason for these different patterns? I guess in both places, we dont expect partial responses so serializing the stream send / receive better?

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -82,17 +86,51 @@ func (s *synchronizerService) Register(ctx context.Context, req *ipb.RegisterReq
...
+	var proposals = []*pb.Match{}
+	for {
+		req, err := stream.Recv()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return err
+		}
+		id = req.GetId()
```
sawagh: Could we just fetch the ID for the first proposal - and if the id ever changes, then we throw an error? Right now you ll pick the last one. It was not a failure scenario for non streaming solution - but for this one, we do expect all IDs to be the exact same and if they are not, rather fail than use the last one

internal/app/backend/synchronizer_client.go
yfei1: See the comment in the evaluator PR.
> I think the code is correct since it performed the required no-leakage actions.
> https://github.com/grpc/grpc-go/blob/6fb38bdee17208f9e2f933614eb6d9052095be22/stream.go#L129.

internal/app/backend/synchronizer_client.go
yfei1: Edited

internal/app/backend/synchronizer_client.go
sawagh: Ah so just handling the error ensures the resources are not leaked? ok.

internal/app/synchronizer/synchronizer_service.go
sawagh: I meant more like:
if req.GetId() == "" --> fail // Request cannot have empty id
if id == ""  --> id == req.GetId() // Fetching the id from first object as the id you care about
if id != req.GetId() --> error // making sure every other object has the same id

PR #759, opened 2019-08-20
Add generated protos for CSharp

yfei1:
the `$<` trick doesn't work for CSharp target since the proto name is in lowercase and target name is in Camel case.

A workaround to this is [Makefile#L658](https://github.com/yfei1/open-match/blob/118291685b9a819377f5261b2443c4321a41c863/Makefile#L658).


Makefile
```diff
@@ -651,13 +653,27 @@ pkg/pb/%.pb.go: api/%.proto third_party/ build/toolchain/bin/protoc$(EXE_EXTENSI
...
+	mv $(REPOSITORY_ROOT)/build/prototmp/open-match.dev/open-match/$@ $@
+
 internal/ipb/%.pb.go: internal/api/%.proto third_party/ build/toolchain/bin/protoc$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-go$(EXE_EXTENSION) build/toolchain/bin/protoc-gen-grpc-gateway$(EXE_EXTENSION)
 	mkdir -p $(REPOSITORY_ROOT)/build/prototmp $(REPOSITORY_ROOT)/internal/ipb
 	$(PROTOC) $< \
 		-I $(REPOSITORY_ROOT) -I $(PROTOC_INCLUDES) \
 		--go_out=plugins=grpc:$(REPOSITORY_ROOT)/build/prototmp
 	mv $(REPOSITORY_ROOT)/build/prototmp/open-match.dev/open-match/$@ $@
 
+internal/ipb/%.cs: third_party/ build/toolchain/bin/protoc$(EXE_EXTENSION)
```
Laremere: We shouldn't be building internal c# protos.  All of the core OM (the only thing that should be talking to internal containers) are written in Go.

internal/ipb/Synchronizer.cs
```diff
@@ -0,0 +1,579 @@
+// <auto-generated>
```
calebatwd: The synchronizer is also internal :) looks like leftover

pkg/pb/Backend.cs
```diff
@@ -0,0 +1,833 @@
...
+//     source: api/backend.proto
+// </auto-generated>
+#pragma warning disable 1591, 0612, 3021
+#region Designer generated code
+
+using pb = global::Google.Protobuf;
+using pbc = global::Google.Protobuf.Collections;
+using pbr = global::Google.Protobuf.Reflection;
+using scg = global::System.Collections.Generic;
+namespace Api {
```
calebatwd: `Api` isn't a particularly friendly namespace. If we're going to check in generation (under the pretense of publishing a client package) consider adding `option csharp_namespace = "OpenMatch";`

jeremyje:
Would it make more sense to have  `openmatch.csproj` that has a source include for `*.cs` files that are the autogenerated proto files instead? That way someone can download the repository and include the cs project. We could also compile a .net core library and maybe even nuget it later on?

PR #760, opened 2019-08-20
Support ticket metadata in Open Match

yfei1:
There is no way to use int64 type directly if we wanna enable the HTTP support (See [grpc-gateway#438](https://github.com/grpc-ecosystem/grpc-gateway/issues/438)) so this commit use `google.protobuf.Timestamp` to represent the creationTimestamp instead as a workaround - although complicates the stuffs a bit :(


api/messages.proto
```diff
@@ -28,15 +29,25 @@ import "google/protobuf/struct.proto";
 // of properties. Open Match stores the Ticket in state storage and enables an
 // Assignment to be associated with this Ticket.
 message Ticket {
+  // Metadata contains useful stats about the ticket.
+  message Metadata {
```
sawagh: Could be another indexable property than a separate object on the Ticket? If we do provide this functionality - we would need to provide the ability to filter tickets which have a wait time longer than x - may be?

Issue #681, opened 2019-07-24
Index configuring / Filtering


sawagh:
Improve the Index support / filtering support for Open Match. Key aspects for this task are:
- What data types should we support as values for properties? 
- What filtering operations should we support for each data type?
- Dynamic index support?
- Limit the data read when processing a Pool (set of filters)

Laremere:
For purposes of discussion, this is the current API:
Indexes are configured as a specific list in the configuration.

From mmlogic.proto:
```protobuf
message QueryTicketsRequest {
  // The Pool representing the set of Filters to be queried.
  Pool pool = 1;
}

message QueryTicketsResponse {
  // The Tickets that meet the Filter criteria requested by the Pool.
  repeated Ticket tickets = 1;
}

// The MMLogic API provides utility functions for common MMF functionality such
// as retreiving Tickets from state storage.
service MmLogic {
  // QueryTickets gets the list of Tickets that match every Filter in the
  // specified Pool.
  rpc QueryTickets(QueryTicketsRequest) returns (stream QueryTicketsResponse) {
    option (google.api.http) = {
      post: "/v1/mmlogic/tickets:query"
      body: "*"
    };
  }
}

```

From messages.proto:
```protobuf
// A hard filter used to query a subset of Tickets meeting the filtering
// criteria.
message Filter {
  // Name of the ticket attribute this Filter operates on.
  string attribute = 1;

  // Maximum value. Defaults to positive infinity (any value above minv).
  double max = 2;

  // Minimum value. Defaults to 0.
  double min = 3;
}

message Pool {
  // A developer-chosen human-readable name for this Pool.
  string name = 1;

  // Set of Filters indicating the filtering criteria. Selected players must
  // match every Filter.
  repeated Filter filters = 2;
}
```

Currently only indexes are used for filtering.  Additionally, the time characteristics of querying multiple indexes isn't well described.  It looks up all indexes and does a set union of the values, not very efficient.

Laremere:
Here's an API which supports a few more filter types:

```protobuf
message Pool {
  // A developer-chosen human-readable name for this Pool.
  string name = 1;

  // Set of Filters indicating the filtering criteria. Selected players must
  // match every Filter.
  repeated RangeFilter range_filters = 2;

  repeated StringEquals string_equals = 3;

  repeated BoolEquals bool_equals = 4;
}

// Filters numerical values to only those within a range.
//   attribute: "foo"
//   max: 10
//   min: 5
// matches:
[...]

PR #773, opened 2019-09-03
Have filter tickets take pool instead of list of filters

Laremere:
Part of the work for https://github.com/googleforgames/open-match/issues/681

This changes the API of FilterTickets to take a pool instead of filter list.  Following the API
in the proposal, different filter types will be different fields on the Pool message.


sawagh:
_This changes the API of FilterTickets to take a pool instead of filter list._ 

This PR does not change any external API in this change right? this just changes the store API interface which is internal only - correct?

Laremere:
> _This changes the API of FilterTickets to take a pool instead of filter list._
> 
> This PR does not change any external API in this change right? this just changes the store API interface which is internal only - correct?

Correct.

PR #781, opened 2019-09-06
Seperate redis and OM index concepts

Laremere:
Currently OM filters directly match filtering fields in redis, and OM ticket properties directly map to values in redis.  This change breaks that direct connection.  In followup changes, I will be adding other index and filter types.  They will be translated into redis sorted set values, so that conversion will take place within these methods.  Eg, bool values will be turned into 0 and 1, and bool equal filters will do a range to capture those values.

This does a couple other minor things:
- Removes a test case that indexed fields have to be numbers, which is going to be wrong after other filter types are added anyways.
- Adds a prefix to the redis key for the index. This will be important as other index types are added to avoid collisions.


internal/statestore/redis_indices.go
```diff
@@ -0,0 +1,99 @@
...
+
+	for _, index := range indices {
+		result = append(result, rangeIndexName(index))
+	}
+
+	return result
+}
+
+func rangeIndexName(attribute string) string {
+	// ri = range index
```
yfei1: Do you mind using “ri stands for range index” or some other complete sentences instead? This line looks like a debugging leftover at the first glance

internal/statestore/redis_indices.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"net/url"
+
+	structpb "github.com/golang/protobuf/ptypes/struct"
+	"github.com/sirupsen/logrus"
+	"open-match.dev/open-match/internal/config"
+	"open-match.dev/open-match/pkg/pb"
+)
+
+type indexedFields struct {
```
yfei1: Would it be more clear if we define an alias ‘type indexedFields map[string]float64’ here instead of defining a new struct?

internal/statestore/redis_indices.go
Laremere: ah, good point.  Done.

internal/statestore/redis_indices.go
Laremere: In a future PR, I also want to return a list of indexes that were used (so delete indexes will work properly.)  That's why I'm not just returning the map.

internal/statestore/redis_indices.go
```diff
@@ -0,0 +1,99 @@
...
+	var result []string
+
+	for _, index := range indices {
+		result = append(result, rangeIndexName(index))
+	}
+
+	return result
+}
+
+func rangeIndexName(attribute string) string {
```
sawagh: Can you please add a comment on why is "ri$" being appended to the attribute name? I am unsure of how this works? a comment somewhere on rangeindexname would be helpful?

Laremere: Added a comment in https://github.com/googleforgames/open-match/pull/785

Issue #767, opened 2019-08-23
API change: Return all tickets when no filters are specified


Laremere:
## Overview

When mmlogic is queried, but the pool contains no filters, return all currently indexed tickets in open-match.

## Motivation

Along with #766, this allows systems which need to correct the entirety of open-match's system state to see all tickets.

It reduces the minimal demo complexity, allowing mmf writers which have small enough player populations to simply work with the entire pool by simply not specifying any filters. (Unlike today, where a dummy index is required.)

This also aligns better with other query formats, eg SQL, where no "where" clause returns all values.

## Impact

This is no a breaking change.

## Change Proto

No changes to protos.

calebatwd:
This is pretty neat. Does it return every ticket as the full hash read as well?
This gets around the dummy issue, but with a few built-in indexes, the dummy one goes away as well... namely `created`. I suppose the created field could also just be used by this api though under the hood (get me all things where created is greater than -forever).
Sweet, sounds cool.

calebatwd:
I think for the proposal, it'd be nice to hear how this will be achieved and if there's any performance impact based on how the scan or lack-of would work

Laremere:
Proposed implementation (roughly):

- When indexing a ticket, additionally add it to a redis set of all tickets.
- When de-indexing a ticket, additionally remove it from that set of all tickets.
- When performing a query: If there are filters present, use existing logic (with those characteristics) to get the list of ticket ids.  If there are no filters present, instead get all ticket ids using the all-tickets set.  As is the current implementation, once the tickets ids are identified, query the tickets in batches to return to the caller.

PR #785, opened 2019-09-06
Add all tickets indexed, used on pools with no filters

Laremere:
Resolves https://github.com/googleforgames/open-match/issues/767


internal/statestore/redis_indices.go
```diff
@@ -75,6 +79,14 @@ func extractIndexFilters(p *pb.Pool) []indexFilter {
 		})
 	}
 
+	if len(filters) == 0 {
+		filters = []indexFilter{{
+			name: allTickets,
+			min:  math.Inf(-1),
+			min:  math.Inf(1),
```
golangcibot: duplicate field name min in struct literal (from `typecheck`)

internal/statestore/redis_indices.go
```diff
@@ -56,6 +58,8 @@ func extractIndexedFields(cfg config.View, t *pb.Ticket) indexedFields {
 		}
 	}
 
+	result[allTickets] = float64(time.Now().UnixNano())
```
golangcibot: invalid operation: cannot index result (variable of type indexedFields) (from `typecheck`)

sawagh: This will have a side-effect (or may be a feature?) that "allTickets" is now a queriable index that can return tickets in a time range :) So some could specify that one filter "allTickets" for range t1 to t2? Should we document it? (not in this PR - but in general documentation for 0.7)

sawagh: Also, can this then just be renamed to "createTime" meta index? always stores time created - and can query based on "createTime" filter too?

Laremere: I just changed it to always be 0, because unit testing with time.Now() doesn't work, and I didn't want to do more plumbing to have real vs stubbed values in the function.

Also, you won't be able to query the allTickets index directly.  That is the point of the index prefixes: It avoids such side-effect behavior.

Laremere: Also, index time doesn't equal create time.  Especially if we make changes to properly support index values changing, it is possible for tickets to get reindexed.

sawagh: Ok. I think as a filter on a ticket conceptually - "alltickets" with value 0 - seems more a workaround for UTs - than "createTime" with value time.Now() at creation time. Also, this could be a cheap way to leverage when we get to the meta index feature that we can expose. Wont block that on this PR - we can always modify it when we get to it.

PR #786, opened 2019-09-06
Exclude Copyright headers and gke-metadata-server when rendering static yaml files

### LINKED ISSUE #778: gke_metadata_server workaround breaks Open Match yaml installation
the `gke_metadata_server` workaround in `install/helm/open-match/templates/open-match-metadata-server.yaml` is defined in the `kube-system` namespace. However, [install Open Match via the yaml files](https://open-match.dev/site/docs/installation/install_om/) requires having Open Match being installed in the `open-match` namespace. Kubernetes will complain about the namespace conflict if we are still doing `kubectl apply -f instal.yaml -n open-match`.

There are three options to fix this issue.

__Option 1__:
The `gke_metadata_server` workaround is required if we want to have a cluster with both psp and workload identity feature enabled, yet the workload identity feature is used for granting internal CI permissions only. We could consider having a new chart for Open Match users and keep the current chart for developer usage.

__Option 2__:
Have yet another workaround to process the yaml file and remove the gke-metadata-server definition.

__Option 3__:
Exclude the gke-metadata-server when generating the static yaml files.

yfei1:
Resolves #778


sawagh:
Motivation?

yfei1:
> Motivation?

The Copyright headers are included in the static yaml files even when we choose to not install a specific component. For example, if we look into the demo installation file 
https://storage.cloud.google.com/open-match-build-artifacts/output/02-open-match-demo.yaml?folder=true&organizationId=433637338589 we can see the later half of the file is just the copyright headers

jeremyje:
This seems like a lot of complexity to address what appears to be a cosmetic issue. Perhaps we can ask Helm team to add an option to exclude empty files when generating YAML files instead.

PR #788, opened 2019-09-06
Move pkg/pb to pkg/gopb

yfei1:
We'll be adding non-Go languages support in the future and this commit moves `pkg/pb` to `pkg/gopb` so that we can have generated code of non-Go languages reside in different directory.


jeremyje:
`pkg/pb` is a well understood directory for Go protobufs. Other languages should not be in the `pkg` directory at all. `pkg` is a Go directory. We should use well understood directory structures for other languages and place them in directories outside of the main Go heavy directories.

PR #791, opened 2019-09-07
Add bool filtering and indexing

Laremere:
Part of implementing https://github.com/googleforgames/open-match/issues/681


internal/statestore/redis_indices.go
```diff
@@ -72,6 +78,20 @@ func extractIndexFilters(p *pb.Pool) []indexFilter {
 		})
 	}
 
+	for _, f := range p.BoolEqualsFilters {
```
yfei1: Can you add some comments to explain the mappings between range filters and bool values?

internal/statestore/redis_indices.go
```diff
@@ -72,6 +78,20 @@ func extractIndexFilters(p *pb.Pool) []indexFilter {
...
+		min := -0.5
+		max := 0.5
+		if f.Value {
```
yfei1: Redis returns all the elements in the sorted set at key with a score between max and min (including elements with score equal to max or min), would it be more clear if we use filter {0,0} for false and {1,1} for true?

internal/statestore/redis_indices.go
```diff
@@ -44,6 +44,12 @@ func extractIndexedFields(cfg config.View, t *pb.Ticket) map[string]float64 {
 		switch v.Kind.(type) {
 		case *structpb.Value_NumberValue:
 			result[rangeIndexName(attribute)] = v.GetNumberValue()
+		case *structpb.Value_BoolValue:
```
sawagh: Is the expectation to set this to false explicitly if false is needed? Or does absence equate automatically to false? 
Just to explicitly clarify:
set foo --> false on ticket:
filter:foo value: false returns the ticket in results. 

Do not set foo at all in Ticket. Do we expect filter:foo, value:false to return the ticket or not?
We should document the expected behavior.

Laremere: All filters will not include any tickets where the field being filtered on is absent.  This is a good question though.  I updated the messages.proto example to be false, so it's more obvious.

internal/statestore/redis_indices.go
Laremere: This feels more obviously correct to me.  As I didn't have to look up that behavior to know it would work.  How strong is this concern?

internal/statestore/redis_indices.go
Laremere: I've added a comment at the top of the file explaining the general strategy.  Does this explain everything that wasn't obvious?

internal/statestore/redis_indices.go
sawagh: Correctness wise, both are correct :). Readability wise, I think setting to 0, 1 and checking for {0,0} {1,1} seems more obvious to me. When i read this, i assumed that may be same value for bounds didnt work for redis filtering and so you ve used these other ranges. If they do - then I do think thats clearer.

sawagh: I ll merge this in interest of getting the API change sooner. You can address this in the related next PR?

PR #792, opened 2019-09-09
Stop using environment variables for Redis connection

yfei1:
This change does the following things:

1. Among the list of environment variables we were using, `OM_MMLOGICAPI` and `REDIS_POOL` related configurations are now served through the `matchmaker_config` file and are safe to be removed.
2. Read Redis password from a file mounted by redis-secret instead.


internal/statestore/redis.go
```diff
@@ -56,9 +57,11 @@ func newRedis(cfg config.View) Service {
 	// Add redis user and password to connection url if they exist
 	redisURL := "redis://"
 	maskedURL := redisURL
-	if cfg.IsSet("redis.password") && cfg.GetString("redis.password") != "" {
-		redisURL += cfg.GetString("redis.user") + ":" + cfg.GetString("redis.password") + "@"
-		maskedURL += cfg.GetString("redis.user") + ":******@"
+
+	// REDIS_PASSWORD is injected by the Redis docker image
```
jeremyje: This comment isn't clear. Does the Redis image really hold a password and send it to us.
Also why read directly from environment variables now when we used to have an environment variable mapping.

The change implies that we are removing environment variable reads but this is just the same thing.

jeremyje: I think this change would make more sense if the secret was accessed via file. We already do this for gRPC TLS certificates.
https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets

PR #793, opened 2019-09-13
Move gke-metadata-server workaround out from install/helm directory

### LINKED ISSUE #778: gke_metadata_server workaround breaks Open Match yaml installation

yfei1:
Resolves #778. This commit moved the gke-metadataserver workaround from `install/helm` to `install` and made changes in the Makefile accordingly.


Makefile
```diff
@@ -323,6 +323,7 @@ build/chart/index.yaml.$(YEAR_MONTH_DAY): build/chart/index.yaml
 build/chart/: build/chart/index.yaml build/chart/index.yaml.$(YEAR_MONTH_DAY)
 
 install-large-chart: build/toolchain/bin/helm$(EXE_EXTENSION) install/helm/open-match/secrets/
+	$(KUBECTL) apply -f install/open-match-metadata-server.yaml
```
jeremyje: This is repetitive. Make this a build dependency. You also are missing the `kubectl$(EXE_EXTENSION)` dependency to call this anyway.

Makefile
```diff
@@ -322,7 +322,10 @@ build/chart/index.yaml.$(YEAR_MONTH_DAY): build/chart/index.yaml
 
 build/chart/: build/chart/index.yaml build/chart/index.yaml.$(YEAR_MONTH_DAY)
 
-install-large-chart: build/toolchain/bin/helm$(EXE_EXTENSION) install/helm/open-match/secrets/
+install-chart-prerequisite: build/toolchain/bin/kubectl$(EXE_EXTENSION)
+	$(KUBECTL) apply -f install/open-match-metadata-server.yaml
```
jeremyje: Rename to `gke-metadata-server-workaround.yaml` so it's a bit clearer that this will eventually go away.

Makefile
```diff
@@ -364,6 +367,7 @@ dry-chart: build/toolchain/bin/helm$(EXE_EXTENSION)
 
 delete-chart: build/toolchain/bin/helm$(EXE_EXTENSION) build/toolchain/bin/kubectl$(EXE_EXTENSION)
 	-$(HELM) delete --purge $(OPEN_MATCH_RELEASE_NAME)
+	-$(KUBECTL) delete rolebinding -n kube-system gce:podsecuritypolicy:gke-metadata-server-workaround
```
jeremyje: Remove this. This will break your GKE cluster.

PR #794, opened 2019-09-13
Add csproj baseline

yfei1:
This commit 

1. Added a Makefile command to enable re-building the csproj with .NET SDK binaries.
2. Added a baseline dll file for `using OpenMatch` in csharp.
3. Added a nuget package file to `dotnet nuget push OpenMatch.0.0.0-dev.nupkg` publish the OpenMatch package.

The next step is to add the generated files to `csharp/OpenＭatch` directory and regenerate the dll file


Makefile
```diff
@@ -979,6 +979,10 @@ proxy:
 update-deps:
 	$(GO) mod tidy
 
+build-csharp:
```
jeremyje: You need to add docker dependency here.

Makefile
```diff
@@ -979,6 +979,10 @@ proxy:
...
+	docker build -t csharp $(REPOSITORY_ROOT)/csharp/OpenMatch
+	(cd $(REPOSITORY_ROOT)/csharp/OpenMatch; docker cp $(shell docker create csharp):/app/. ./.)
```
jeremyje: This is somewhat unusual. I think we should build locally. The rest of the makefile builds locally so you have easy access to artifacts.

yfei1:
> Add the binaries to .gitignore and friends.

It seems like they are happy with https://github.com/googleforgames/open-match/blob/master/.gitignore#L74

jeremyje:
> > Add the binaries to .gitignore and friends.
> 
> It seems like they are happy with https://github.com/googleforgames/open-match/blob/master/.gitignore#L74

the nupkg and dll file are still in the change. They should not be. Add them to .gitignore and friends.

I see *.dll is ignored so no idea why that's included. The cases match. nupkg is not ignored.

PR #798, opened 2019-09-16
 Add string equals filtering and indexing

Laremere:
Part of implementing #681


internal/statestore/redis_indices.go
```diff
@@ -58,6 +58,8 @@ func extractIndexedFields(cfg config.View, t *pb.Ticket) map[string]float64 {
 				value = 1
 			}
 			result[boolIndexName(attribute)] = value
+		case *structpb.Value_StringValue:
```
sawagh: Could you as a separate task document the usage of filters / expectations of what needs to be specified etc.? Some easy reference we can point folks to for the current indexing / filtering mechanism?

api/messages.proto
```diff
@@ -97,6 +97,23 @@ message BoolEqualsFilter {
   bool value = 2;
 }
 
+// Filters strings exactly equaling a value.
+//   attribute: "foo"
```
sawagh: Can I specify value of my attribute to be null? Should we fail setting it?

Laremere: Added null to does not match list.

internal/statestore/redis_indices.go
Laremere: Forgot to add strings to the documentation above.  Are you expecting more?

sawagh: I meant website documentation on usage - not so much the code documentation. Again, not for this PR - but something the user guide should have.

Issue #723, opened 2019-08-02
Should the proto package be open-match?


Laremere:
Currently the proto package is "api", or "api.internal".  This means if another project includes anything (such as assignment) in it's protos, then it will need to reference it as "api.assignment", which feels off.  Shouldn't it be "openmatch.assignment"?  Why was the name api chosen in the first place?

yfei1:
We are following the conventions at here https://github.com/golang-standards/project-layout/tree/master/api

Laremere:
Yes, that's where the proto files are.  However I'm talking about the proto package (https://developers.google.com/protocol-buffers/docs/proto3#packages) declaration.

PR #799, opened 2019-09-16
Change protobuf namespace to openmatch

Laremere:
This resolves https://github.com/googleforgames/open-match/issues/723

It would be very weird for other protobuf packages to be importing "api" for Open Match.  This changes to a more reasonable unique name.

Eg, tensorflow uses the package "tensorflow" https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/protobuf/config.proto


api/backend.swagger.json
```diff
@@ -129,7 +129,11 @@
       },
       "description": "An Assignment object represents the assignment associated with a Ticket. Open\nmatch does not require or inspect any fields on assignment."
     },
+<<<<<<< HEAD
```
yfei1: bad merge?

Laremere: fixed.

PR #802, opened 2019-09-17
Implement the profiles package that will be used by the backend to generate profiles for different scenarios for scale testing

sawagh:


examples/scale/profiles/util.go
```diff
@@ -0,0 +1,72 @@
...
+			name: fmt.Sprintf("%s_%dto%d", config.name, r, max),
+			min:  r,
+			max:  max,
+		})
+	
+		r = r + 1 + (config.rangeSize - config.rangeOverlap)
+	}
+
+	return filters
+}
```
yfei1: \n

PR #803, opened 2019-09-17
Implement test frontend that creates tickets in Open Match continuously

sawagh:


examples/scale/frontend/frontend.go
```diff
@@ -39,10 +45,48 @@ func Run() {
...
+				defer wg.Done()
+				req := &pb.CreateTicketRequest{
+					Ticket: tickets.Ticket(cfg),
+				}
+
+				if _, err := fe.CreateTicket(context.Background(), req); err != nil {
+					logger.WithFields(logrus.Fields{
+						"error": err.Error(),
+					}).Error("failed to create a ticket.")
+					failed++
```
yfei1: This pattern may lead to race condition

sawagh: modified to using atomic.AddUint64

PR #804, opened 2019-09-17
Implement test backend that will fetch matches, assign tickets and delete tickets at scale

sawagh:


examples/scale/backend/backend.go
```diff
@@ -40,7 +49,149 @@ func Run() {
...
+			p := p
+			go func(wg *sync.WaitGroup) {
+				defer wg.Done()
+				results, err := doFetch(ctx, be, p)
+				if err != nil {
+					logger.Errorf("FetchMatches failed, got %v", err)
+					return
+				}
+
+				matches = append(matches, results...)
```
yfei1: I think the vanilla slice in Go is not concurrency safe

Laremere: Some patterns are safe concurrently (reads and editing indices known to not overlap), but append is not.  I recommend using channels.  This simplest one here is probably results := make(chan []*pb.Match, len(mprofiles), buffering the results so you don't need another go routine to pull them out.  Using this you can actually get rid the the waitgroup:

for i := 0; i < len(mprofiles); i++ {
  matches = append(matches, results...)
}

Laremere: actually, I was mixing up things in my head.  Buffered is good if you still have the wait group, no need to make it buffered if you replace the waitgroup.

PR #805, opened 2019-09-17
Add end-to-end test for new filter types

yfei1:


test/e2e/query_tickets_test.go
```diff
@@ -155,7 +282,7 @@ func TestQueryTickets(t *testing.T) {
 				pageCounts := 0
 				ctx := om.Context()
 
-				test.preAction(ctx, fe, t)
+				doCreateTicket(ctx, fe, test.gotTickets, t)
```
Laremere: Why are you declaring doCreateTickets and using it once instead of just having the code it contains here?

PR #809, opened 2019-09-18
Enable Synchronizer by default

sawagh:


sawagh:
> Can you do `make update-chart-deps` and upload the charts?

Done! Thanks!

PR #813, opened 2019-09-19
Add csharp generated code and update .csproj dependency

yfei1:


Makefile
```diff
@@ -687,6 +689,24 @@ pkg/pb/%.pb.go: api/%.proto third_party/ build/toolchain/bin/protoc$(EXE_EXTENSI
 		--go_out=plugins=grpc:$(REPOSITORY_ROOT)/build/prototmp
 	mv $(REPOSITORY_ROOT)/build/prototmp/open-match.dev/open-match/$@ $@
 
+csharp/OpenMatch/Annotations.cs:
+	$(PROTOC) third_party/protoc-gen-swagger/options/annotations.proto \
+			-I $(REPOSITORY_ROOT) -I $(PROTOC_INCLUDES) \
```
jeremyje: That indentation tho.

Makefile
```diff
@@ -687,6 +689,24 @@ pkg/pb/%.pb.go: api/%.proto third_party/ build/toolchain/bin/protoc$(EXE_EXTENSI
...
+			--csharp_out=$(REPOSITORY_ROOT)/csharp/OpenMatch
+
+csharp/OpenMatch/Openapiv2.cs:
+	$(PROTOC) third_party/protoc-gen-swagger/options/openapiv2.proto \
+			-I $(REPOSITORY_ROOT) -I $(PROTOC_INCLUDES) \
+			--plugin=protoc-gen-grpc=grpc_csharp_plugin \
+			--csharp_out=$(REPOSITORY_ROOT)/csharp/OpenMatch/
+
+csharp/OpenMatch/%.cs: third_party/ build/toolchain/bin/protoc$(EXE_EXTENSION)
+	$(PROTOC) api/$(shell echo $(*F)| tr A-Z a-z).proto \
```
jeremyje: Don't do this. Have the proto as the input parameter. See
`internal/ipb/%.pb.go: internal/api/%.proto` just below.

Makefile
```diff
@@ -912,6 +932,8 @@ clean-secrets:
 	rm -rf $(OPEN_MATCH_SECRETS_DIR)
 
 clean-protos:
+	rm -rf $(REPOSITORY_ROOT)/build/prototmp
```
jeremyje: is that a file or a directory? if directory add the ending `/`

Makefile
```diff
@@ -1021,8 +1043,8 @@ proxy:
 update-deps:
 	$(GO) mod tidy
 
-build-csharp: build/toolchain/dotnet/
-	(cd $(REPOSITORY_ROOT)/csharp/OpenMatch && $(DOTNET) build -o . && rm -rf obj/ *.pdb *.deps.json)
+build-csharp: build/toolchain/dotnet/ csharp/OpenMatch/Annotations.cs csharp/OpenMatch/Openapiv2.cs
+	(cd $(REPOSITORY_ROOT)/csharp/OpenMatch && $(DOTNET) build -o . && rm -rf ./obj/ *.pdb *.deps.json *.nupkg *.dll)
```
jeremyje: Why delete debug symbols on the build step?

Makefile
yfei1: I've tried but that will throw an error for csharp. The protos are all lowercase and .cs files are in CamelCase

jeremyje: Can you put the error you saw here?

yfei1: Oh actually it is not an error, if I changed it to the target below, I will get an error like this because the `%.proto` regex maps to `Backend.proto` but we only have `backend.proto` instead.
```
yufanfei@yufanfei:~/go/src/open-match.dev/open-match$ make all-protos 
make: *** No rule to make target 'csharp/OpenMatch/Backend.cs', needed by 'all-protos'.  Stop.
```

```bash
csharp/OpenMatch/%.cs: api/%.proto third_party/ build/toolchain/bin/protoc$(EXE_EXTENSION)
	$(PROTOC) $< \
		--plugin=protoc-gen-grpc=grpc_csharp_plugin \
		--csharp_out=$(REPOSITORY_ROOT)/csharp/OpenMatch
```

PR #817, opened 2019-09-19
Pass synchronizer id on the context of the call to synchronizer

sawagh:


internal/util/synccontextid.go
```diff
@@ -0,0 +1,72 @@
...
+	values := getSynchronizerContextIDFromMetadata(md)
+	if len(values) == 1 {
+		return values[0]
+	}
+
+	return ""
+}
+
+func getSynchronizerContextIDFromMetadata(md metadata.MD) []string {
+	values := md.Get(metadataNameSynchronizerContextID)
```
yfei1: Is the following code the same thing as `return md.Get(metadataNameSynchronizerContextID)`?

internal/util/synccontextid.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+// AppendSynchronizerContextID adds the synchronizer context id to a request context metadata.
+func AppendSynchronizerContextID(ctx context.Context, id string) (context.Context, error) {
+	md, ok := metadata.FromOutgoingContext(ctx)
+	if ok {
+		values := getSynchronizerContextIDFromMetadata(md)
+		if len(values) == 1 {
+			if values[0] != id {
+				return ctx, errors.New("request already has a synchronizer context id")
```
yfei1: maybe `request already has a different synchronizer context id`

sawagh: done.

internal/util/synccontextid.go
sawagh: That would get nil in some cases - might be more readable to return empty slice for len on it to be

PR #819, opened 2019-09-19
Open Match scale testing improvements

sawagh:


go.mod
```diff
@@ -31,6 +31,7 @@ require (
 	github.com/golang/protobuf v1.3.2
 	github.com/gomodule/redigo v1.7.1-0.20190322064113-39e2c31b7ca3
 	github.com/google/gofuzz v1.0.0 // indirect
+	github.com/google/uuid v1.1.1
```
yfei1: Where does this dependency come from? Could you run `go mod tidy` and see if it is still here?

PR #823, opened 2019-09-20
Make CI compatible with go1.13

yfei1:
This commit does version updates to the dependencies we are using in CI to support go1.13


yfei1:
> /gcbrun

Golangci-lint starts to support go1.13 last week but golangci bot is not updated yet. Will hold off this change for a few days and see if there is any progress.

PR #828, opened 2019-09-23
Fix scale tests with the up-to-date indices

yfei1:


yfei1:
> Other attributes are pulled from the config values file - (region / character). Can you update those too :)

Those values will be updated by #825, for now changing the config from `mmr.rating` to `attribute.mmr` is enough.

PR #829, opened 2019-09-24
Fix OpenConcensus backend_matches_fetched metric

yfei1:


internal/app/backend/backend_service.go
```diff
@@ -99,15 +99,15 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
+			err = stream.Send(&pb.FetchMatchesResponse{Match: result})
+			telemetry.IncrementCounter(stream.Context(), mMatchesFetched)
 			if err != nil {
 				logger.WithError(err).Error("failed to stream back the response")
 				return err
 			}
 		}
 	}
 
-	telemetry.IncrementCounterN(stream.Context(), mMatchesFetched, len(results))
```
sawagh: Whats the root cause here? Why does IncrementCounterN here fail and IncrementCounter for each match added pass?

yfei1: Most of the emitted metrics are not recorded because function returns from the `stream.Context().Done()` case

PR #830, opened 2019-09-24
DeleteFromIgnoreList when tickets got assigned or deleted

yfei1:


internal/app/backend/backend_service.go
```diff
@@ -298,5 +298,9 @@ func doAssignTickets(ctx context.Context, req *pb.AssignTicketsRequest, store st
 		}
 	}
 
+	if err = store.DeleteTicketsFromIgnoreList(ctx, req.GetTicketIds()); err != nil {
+		logger.WithError(err).Errorf("failed to delete tickets %v from the ignorelist", req.GetTicketIds())
```
sawagh: use logger.WithFields(...).Error() ?

PR #845, opened 2019-10-03
Update telemetry/metrics to support adding histogram view

yfei1:
This commit updated `telemetry/metrics.go` file to support Prometheus's histogram view. Also added a histogram metric to monitor Redis connection latency.

Will rename`IncrementCounterN` and `IncrementCounter` to a more general name in a later PR as it is applicable to `gauge/counter/histogram` views


internal/telemetry/metrics.go
```diff
@@ -45,19 +45,42 @@ func Counter(name string, description string, tags ...tag.Key) *stats.Int64Measu
...
+	histogramView(s, bounds, tags...)
+	return s
+}
+
 // IncrementCounter +1's the counter metric.
 func IncrementCounter(ctx context.Context, s *stats.Int64Measure, tags ...tag.Mutator) {
 	IncrementCounterN(ctx, s, 1, tags...)
 }
 
 // IncrementCounterN increases a metric by n.
```
Laremere: I don't understand this comment, in its relationship to the histogram.  I would expect, eg, that if I handled 2 tickets, I would increment the tickets handled by 2 using this command.  I'm not seeing how this related to histogram usage.  When I made a redis connection, why am I incrementing this by the time it took, eg, if it took 100ms, I should be updating the relevant bucket as +1, not adding 100 times connected to the total connected count.

yfei1: See the commit description for explanation. I'll do the rename in a later PR. Actually `IncrementCounter` is not an appropriate name for this function. What `RecordWithTags` does is `records one or multiple measurements at once`.

Laremere: Generally, I would prefer the rename be done first.  Keep that in mind in the future.  For now, update the comment to be accurate, at least.

yfei1: Done. Will do the rename in the next PR.

Issue #765, opened 2019-08-23
API change: Document-style structure to compositional contracts


calebatwd:
## Overview
Currently, open match apis rely heavily on loosely-structured `struct` (json-style) contracts, similar to a document database. The implication is that open match will, in some way shape or form, care about some _or all_ of the paths and types of the underlying data stored in the ticket documents (today, that's string value, number value, null value, and boolean value). We do this for custom ticket indexing... Spoken: here's my ticket document, the current indexes are "player.[0].skill" and "party.modeSelection". In this way, open match can use custom indexing policies to read and interpret the tickets schema and built-in types and build indexes without needing to infer the type or search for it in the document.

Pros:
1. The built-in types are easily supported without additional work (they can be inferred, which enables wildcard indexing /*)
2. Follows existing document model index policy styles (less invention).
3. Less duplication of data if the developer commits to using the data format in their match function

Cons:
1. Requires the developer use struct in their match functions... any binary data formats (commonly used in games) needs to be serialized into a string, and deserialized from a string back during match function execution. Struct doesn't support *any*, so we're stuck with Base91 (possibly 128) string encoding. 
2. While the types (number, string, boolean, and null) are initially sufficient types for indexing, any additional types (like vector) would require a type field be introduced to any structuring, or the custom index policy would need to map paths to type. (I also think this will be necessary anyways... imagine: "attributes.spacialSkill": "vector". See config proposal coming...)
3. If the developer chooses another format, they have to duplicate data anyways

## Proposal
Adopt a `multipart/form-data`-style api, where there's both a document, and a blob
- Rename `Properties` to `Attributes`... arguably, this is pedantic, but I kinda think about them in reverse of what we currently have. Open to feedback on that.
- Add a `Properties` as an `Any` that the developer can use to bypass open match serialization for their match function. (Any is better than byte[] in this case because the type field can be used to describe the contract and version the developer needs to interpret during match function execution)
```
google.protobuf.Any properties  // An indexable document provided by the game client
google.protobuf.Struct attributes // Whatever data the developer wants to deserialize in their match function, which open match doesn't care about. This can be put into the database hash as is, which is convenient.
```

Pros:
1. Still get the benefits of `struct` for document style indexing policies
2. I think these are a bit more common sense to use (attributes are things for open match to index, properties are mine to use in match functions (my "developer" contract))
3. Reduce serialization and smaller format on the wire if the developer wants (no in and out of string with binary data)

Cons:
[...]

PR #846, opened 2019-10-03
Adding Any fields to messages.proto

Laremere:
See https://github.com/googleforgames/open-match/issues/765 for detailed discussion.

Followup changes will involve wiring up the search_fields, then transitioning the existing tests and examples, then removing the old struct fields.


api/messages.proto
```diff
@@ -34,10 +35,28 @@ message Ticket {
...
   // used in indexing and filtering to find tickets.
+  // TODO: Deprecate and remove this field.
   google.protobuf.Struct properties = 2;
 
   // Assignment associated with the Ticket.
   Assignment assignment = 3;
+
+  SearchFields search_fields = 4;
+
+  google.protobuf.Any criteria = 5;
```
sawagh: Naming is hard :) - 'criteria' is confusing. I think 'properties' seems to be the closest - we should be ok making braking changes at this stage right?

api/messages.proto
```diff
@@ -34,10 +35,28 @@ message Ticket {
...
+// Search fields are the fields which Open Match is aware of, and can be used
+// when specifying filters.
+message SearchFields {
+  // Float arguments.  Filterable on ranges.
+  map<string, double> fargs = 1;
+ 
+  // String arguments.  Filterable on equality.
+  map<string, string> sargs = 2;
+ 
+  // Filterable on presence or absence of given value.
```
sawagh: examples on each of these would be super useful - definitely for this one. may be just my familiarity with old model but stating this is a bool filter seems to clarify well for me.

api/messages.proto
```diff
@@ -47,10 +66,15 @@ message Assignment {
   string connection = 1;
```
sawagh: do we care about explicitly having a connection / error any more? or just have an any that can have whatever? since we are changing this, lets revisit / finalize this - @calebatwd ?

Laremere: For this change, keeping this around.

api/messages.proto
Laremere: Added examples to the filter specification, where the others examples are as well.

api/messages.proto
Laremere: How about "mmf_input"?  I want to keep it clear that this is that any for data for helping run the mmf, and not as input for other systems (which would need their own fields to preserve system composability.)

Changed some other fields to match, not really happy with "handling_output".

Laremere: Changed "handling_output" fields to "fe_outout" and "be_output", better at least.

api/messages.proto
```diff
@@ -34,10 +35,28 @@ message Ticket {
...
   // used in indexing and filtering to find tickets.
+  // TODO: Deprecate and remove this field.
   google.protobuf.Struct properties = 2;
 
   // Assignment associated with the Ticket.
   Assignment assignment = 3;
+
+  SearchFields search_fields = 4;
+
+  google.protobuf.Any mmf_input = 5;
```
sawagh: input / output suffixes are understood only in context of the original proposal discussion - i think their meaning is not obvious just as an Any field seen on the Ticket. 

If not Properties, can we simply call it data? Ticket-->data MatchProfiles-->data Assignment-->data ?

Lets discuss naming offline. The current input / output naming does not seem intuitive.

Laremere: Added a comment, willing to discuss further.

Laremere: Per discussion, renamed to extension.

PR #849, opened 2019-10-04
Switched to gRPC client load balancer and DNS resolver

yfei1:


sawagh:
Have we considered using a service mesh? Here is an article explaining the advantages over the current approach:
https://kubernetes.io/blog/2018/11/07/grpc-load-balancing-on-kubernetes-without-tears/

The approach we take works for internal clients using our library - but how would this go for externally written clients? Linkerd seems to have a better latency based LB too - rather than just roundrobin. Thoughts?

.golangci.yaml
```diff
@@ -181,6 +181,7 @@ linters:
     - interfacer # deprecated - "A tool that suggests interfaces is prone to bad suggestions"
     - gochecknoglobals
     - funlen
+    - gochecknoinits
```
jeremyje: use `// nolint:gochecknoinits instead.`

install/helm/open-match/subcharts/open-match-customize/values.yaml
```diff
@@ -18,10 +18,12 @@
 
 function:
   replicas: 3
+  portType: ClusterIP
   image: openmatch-mmf-go-soloduel
 
 evaluator:
   replicas: 3
+  portType: ClusterIP
```
jeremyje: This is kinda weird to need here.

yfei1:
> Have we considered using a service mesh? Here is an article explaining the advantages over the current approach:
> https://kubernetes.io/blog/2018/11/07/grpc-load-balancing-on-kubernetes-without-tears/
> 
> The approach we take works for internal clients using our library - but how would this go for externally written clients? Linkerd seems to have a better latency based LB too - rather than just roundrobin. Thoughts?

I think doing client side LB or Linked will both introduce an extra layer of complexity to Open Match and installing Linkerd for Open Match is also a non-trivial task. 
1. Linkerd itself includes Prometheus and Grafana installations, they claimed that their Prometheus is tightly bonded to Linkerd and is not optional - I think people may start to wonder why we need two different Prometheus for running Open Match if we choose this approach.
2. We may need to modify our PodSecurityPolicies to grant Linkered enough permission. I've tried working on installing it with Open Match on Friday but with no luck :(

sawagh:
If the current approach works for now, I propose submitting it as a spot fix - and iterating over the issue still in v0.8 - I think all the concerns around this are valid but if a solution is ready and unblocks further scale testing, can we put it in?

sawagh:
Remove this PR unless you plan to use it as reference as short term fix?

PR #854, opened 2019-10-09
Switch Default Evaluator to use proto any

Laremere:
Added a new proto to messages, DefaultEvaluationCriteria.

Reworked the default evaluator to use it.
Changed the pool mmf to output it.


examples/evaluator/golang/simple/evaluate/evaluator.go
```diff
@@ -15,40 +15,78 @@
...
-			return false
-		}
-	}
-	return true
+type matchExt struct {
+	match *pb.Match
+	ext   *pb.DefaultEvaluationCriteria
+}
+
+type matcheExts []*matchExt
```
sawagh: Is brevity the reason to rename []*matchExt? An extra lookup for me to see what matcheExts was and also that it looks like a typo makes met think []*matcheExts to be more readable.

examples/evaluator/golang/simple/evaluate/evaluator.go
```diff
@@ -15,40 +15,78 @@
...
 
-func isNonCollidingMatch(match *pb.Match, validTickets map[string]bool) bool {
-	for _, ticket := range match.GetTickets() {
-		id := ticket.GetId()
-		if _, ok := validTickets[id]; ok {
-			return false
-		}
-	}
-	return true
+type matchExt struct {
```
sawagh: can we move this to the top of the file please?

examples/evaluator/golang/simple/evaluate/evaluator.go
```diff
@@ -15,40 +15,78 @@
...
 	results := []*pb.Match{}
-	dedup := map[string]bool{}
+	ticketsUsed := map[string]struct{}{}
 
-	for _, match := range p.Matches {
-		if isNonCollidingMatch(match, dedup) {
-			for _, ticket := range match.GetTickets() {
-				dedup[ticket.GetId()] = true
+outer:
+	for _, matchExt := range matches {
```
sawagh: can we avoid reusing the same name as a type for an loop variable?

examples/evaluator/golang/simple/evaluate/evaluator.go
```diff
@@ -15,40 +15,78 @@
...
+		for _, ticket := range matchExt.match.GetTickets() {
+			if _, ok := ticketsUsed[ticket.Id]; ok {
+				continue outer
```
sawagh: Can we use a helper function to check hasDuplicateTickets? I tend to agree with the discussion and solution here for that being a relatively elegant (and readable) solution for nested loops

examples/evaluator/golang/simple/evaluate/evaluator.go
```diff
@@ -15,40 +15,78 @@
...
-	scoreInDescendingOrder := func(a, b *pb.Match) bool {
-		return a.GetProperties().GetFields()[examples.MatchScore].GetNumberValue() > b.GetProperties().GetFields()[examples.MatchScore].GetNumberValue()
+	matches := make(matcheExts, 0, len(p.Matches))
+	for _, match := range p.Matches {
+		// Evaluation criteria is optional, but sort it lower than any matches which
+		// provided criteria.
+		ext := &pb.DefaultEvaluationCriteria{
+			Score: math.Inf(-1),
+		}
+		if match.Extension != nil {
```
sawagh: A warning log if this is nil? Basically that match is completely ignored - so some trace in logs will be useful to debug.

examples/evaluator/golang/simple/evaluate/evaluator.go
Laremere: done.

examples/evaluator/golang/simple/evaluate/evaluator.go
Laremere: What discussion?

I think keeping the test on ticketsUsed and the update on it relatively close improves this code.  I could move it all to another function, but that requires more structure just to have a return statement instead of a continue.

sawagh: Sorry - forgot to link the Stackoverflow thread i was referring to: https://stackoverflow.com/questions/51996175/how-to-break-out-of-nested-loops-in-go

Specifically referring to the use of labels to control flow - I do think helper function that returns / breaks is more readable.

sawagh: for _, matchExt := range matches {
    if containsUsedTickets(match, ticketsUsed) {
       continue;
    }
    // append
}

examples/evaluator/golang/simple/evaluate/evaluator.go
sawagh: _This is desired behavior, as for someone who isn't aware of the evaluator, it just works_

Thanks for clarifying the behavior - I agree the behavior is appropriate. I think adding more to the logs is helpful - leaves a trace of what happens. If somoene does not populate any score in MMF because they just didnt know that was a thing - I think it is reasonable (and a very critical debugging piece) for logs to give a hint.

Laremere: Ok, I don't want to spam the logs too much, and it definitely shouldn't be a warning because nothing is going wrong.  Adding info logs once per evaluate call.

examples/evaluator/golang/simple/evaluate/evaluator.go
Laremere: done.

PR #859, opened 2019-10-10
Remove internal dependencies from the demo

yfei1:
This commit -
1. Removed the internal dependencies from the demo as much as possible and use hard-coded hostnames instead of reading things from the configmap. 
2. Use FQDN to allow demo deployment in the `open-match-demo` namespace.

The next step is to decouple the actual matchmaking logic from the frontend and backend components to make the code easier to understand for Open Match beginners.


examples/demo/components/clients/clients.go
```diff
@@ -81,7 +80,13 @@ func runScenario(ctx context.Context, cfg config.View, name string, update updat
 	s.Status = "Connecting to Open Match frontend"
 	update(s)
 
-	conn, err := rpc.GRPCClientFromConfig(cfg, "api.frontend")
+	/*
```
Laremere: This comment (and the one associated with it in the director) seem unnecessary, or at least misplaced.
These files are for the absolute first introduction into OM.  There's no need to throw k8s networking guides here.  As part of the user journey, this type of content may make sense in a "writing your own function and clients" section.

yfei1: I have a different opinion. As the first introduction into OM, we should at least explain where does these magical string literals and port numbers comes from. I believe the goal of the tutorial is to clarify things instead of leaving things there without any comments

sawagh: I think it may be ok to leave out the guides. The first line explaining the cross namespace is useful though. The guide links can belong to documentation references.

Laremere: If you feel a comment is necessary, I would be ok with 
```
// see https://open-match.dev/site/docs/guides/api/
```
Bits about how to connect securely, info about k8s' domain system should be left out.  If further information about how to connect to open match is required, it should be collected into that one spot.
The first comment line should be omitted altogether.  Comments shouldn't state things that the code states directly.  (these comments have a high probability of rotting, and generally overlap with what is easily discernible)  They should generally be reserved for describing why something is done a certain way (eg, to prevent people from "fixing" it), or when a piece of code has a non-obvious side effect.

Laremere: This piece of code is also heavily tied to the first match documentation, so a link to the API guide from https://open-match.dev/site/docs/getting-started/first_match/ would be sufficient.

PR #860, opened 2019-10-10
Use SearchField's doubles for double indexing

Laremere:
Part of moving to the new API proposal: #765

This replaces the double indexing, and fixes the spots which break because of it.
Future PRs will remove indexing and the passing of configuration that were because of it, and then also remove the properties altogether. (there's still a handful of places which use properties but don't actually use it for indexing.)


examples/functions/golang/pool/mmf/matchfunction.go
```diff
@@ -66,8 +66,8 @@ func MakeMatches(params *mmfHarness.MatchFunctionParams) ([]*pb.Match, error) {
 func scoreCalculator(tickets []*pb.Ticket) float64 {
 	matchScore := 0.0
 	for _, ticket := range tickets {
-		for _, v := range ticket.GetProperties().GetFields() {
-			matchScore += v.GetNumberValue()
+		for _, v := range ticket.GetSearchFields().GetDoubleArgs() {
```
sawagh: Some comment on what this does? :) I know you stumbled on what the previous one was doing - so even if not scientific .. a note stating this is what is does as a placeholder would be great.

internal/statestore/redis_indices.go
```diff
@@ -35,8 +33,14 @@ import (
 // Filters are strings look up that attribute/value pair.
```
sawagh: Update the comment above this function needs to change to reflect the tags instead of boolean.

examples/functions/golang/pool/mmf/matchfunction.go
Laremere: There was already one, expanded a little.

internal/statestore/redis_indices.go
Laremere: done.

internal/statestore/redis_indices.go
```diff
@@ -29,14 +27,20 @@ import (
 // for simplicity, sorted sets.  The following translations are used:
 // Float values to range indicies use the float value directly, with filters
 // doing direct lookups on those ranges.
-// Boolean values with bool equals indicies turn true and false into 1 and 0.
-// Filters on bool equality use ranges limited to only 1s or only 0s.
+// Tags indicate presense in the set.  The value used is 0.  Filters on tags
```
golangcibot: `presense` is a misspelling of `presence` (from `misspell`)
```suggestion

```

PR #862, opened 2019-10-11
 Remove index configuration

Laremere:
Part of moving to the new API proposal: #765


Laremere:
> Unless there is already one - can you please create a v0.8 issue to track the ability to 'configure open match to only index on a subset of search criteria fields'? we can discuss the mechanics of how in that issue.

Done: https://github.com/googleforgames/open-match/issues/869

PR #864, opened 2019-10-11
Change k8s service type into headless service and use DNS resolver to get endpoint

yfei1:
Use k8s endpoint getter for the end-to-end test instead to get pod IPs. Also explicitly use passthrough resolver for the end-to-end test to bypass the DNS restriction for localhost deployment.


install/helm/open-match/templates/synchronizer.yaml
```diff
@@ -28,7 +28,11 @@ spec:
     app: {{ template "openmatch.name" . }}
     component: synchronizer
     release: {{ .Release.Name }}
-  type: {{ coalesce .Values.global.kubernetes.service.portType .Values.synchronizer.portType }}
+{{- $portType := coalesce .Values.global.kubernetes.service.portType .Values.synchronizer.portType -}}
```
sawagh: Synchronizer does not scale up? Only single pod - do we still need this.

internal/rpc/server.go
```diff
@@ -300,7 +302,18 @@ func instrumentHTTPHandler(handler http.Handler, params *ServerParams) http.Hand
 }
 
 func newGRPCServerOptions(params *ServerParams) []grpc.ServerOption {
-	opts := []grpc.ServerOption{}
+	opts := []grpc.ServerOption{
+		grpc.KeepaliveParams(keepalive.ServerParameters{
+			// MaxConnectionAge is a duration for the maximum amount of time a
```
sawagh: So will the roundrobin only happen after a minute? Will we get true roundrobin or only at chunks of this time?

yfei1: The max connection age only affects how we establish new gRPC connections, so basically instead of having a long live connection, we instruct gRPC to initialize a new connection every minute so that they can be mapped/unmapped from pods that were spawned/killed.

install/helm/open-match/templates/synchronizer.yaml
yfei1: Good point, fixed.

internal/testing/e2e/common.go
```diff
@@ -60,6 +62,7 @@ func New(t *testing.T) (OM, func()) {
 
 // RunMain provides the setup and teardown for Open Match e2e tests.
 func RunMain(m *testing.M) {
+	resolver.SetDefaultScheme("passthrough")
```
sawagh: We need to add a comment here stating the reason for this so in future, this is not cryptic to someone who does not understand the underlying issue.

yfei1:
> Discussed offline with Yufan. The one drawback of this change is it does not work so well with pod autoscaling. For scaling up, it may take up to the connection timeout for the changes to reflect. For scale down, if it still keeps connections and sends requests to scaled down pods and fails, that may be a non starter. Please explore these scenarios and their behavior. This may still be an OK stopgap to unblock further scale testing - but wont be a final solution if these scenarios dont work well.

It turns out that HorizontalAutoScaler works ok with this solution in both up/down scaling senarios. Tested.

sawagh:
Given the amount of context for this issue, I think it would be best to do the following:
- Create an issue to explore LinkerD for future (v1.0)
- Create an issue to explore solutions based on Ingress Controller for LoadBalancer port type. (v1.0)

We also want to create some documentation explaining the current solution - May be just a page 'GRPC Loadbalancing' under our reference documentation that we can point folks to in the interim that explains what we currently have and the drawbacks. We can later move it to where it fits but for now - having a pointer to all the context would help. Create an issue for that too - and that one can be targeted to this release.

wmorgan:
Definitely interested in how your Linkerd exploration goes! Feedback more than welcome.

PR #867, opened 2019-10-11
Define minimalistic pod resources to deploy sample mmf and evaluator

yfei1:


yfei1:
> Would it be better if these co-exist with the actual example function / evaluator that will be deployed using these? I am working on some standalone samples - and these can come in handy. We can check them in for now and then I can pull them over to my samples? What say?

I think this is a better approach if we only have one function example, yet given we have three different function examples, having just resource definition for one of them may looks inconsistent with the others. What do you think?

sawagh:
I thi

> > Would it be better if these co-exist with the actual example function / evaluator that will be deployed using these? I am working on some standalone samples - and these can come in handy. We can check them in for now and then I can pull them over to my samples? What say?
> 
> I think this is a better approach if we only have one function example, yet given we have three different function examples, having just resource definition for one of them may looks inconsistent with the others. What do you think?

I think it is OK to submit this for now to have all the content we need checked in - at some point in this release (next 2 weeks), lets look at all our examples / demo / scale examples / standalone - and come up with which of these go to ecosystem repo / stay in OM repo and finalize the structure.

PR #868, opened 2019-10-11
Update release template to install open match without telemetry services

yfei1:


docs/governance/templates/release.md
```diff
@@ -54,8 +54,8 @@ To deploy Open Match in your Kubernetes cluster run the following commands:
 kubectl create clusterrolebinding myname-cluster-admin-binding --clusterrole=cluster-admin --user=$(YOUR_KUBERNETES_USER_NAME)
 # Place all Open Match components in their own namespace.
 kubectl create namespace open-match
-# Install Open Match and monitoring services.
-kubectl apply -f https://github.com/googleforgames/open-match/releases/download/v{version}/install.yaml --namespace open-match
+# Install Open Match services.
```
sawagh: What are the implications of changing this here? Only impacts release notes?

yfei1: Yes, only the release notes and the docs in `open-match-doc` repo. The `01-open-match-core.yaml` file already exists so we can just pull it from the GCS bucket.

sawagh:
We can revisit the doc changes together at the end.

PR #872, opened 2019-10-13
Proposal:  Make extension a repeated field

Laremere:
This is a proposal to consider and discuss in PR form (since it was easy enough to just do.)

Replaces the extension field (and match.evaluator_input) with a repeated extensions field.  The previous concept was that different components would read from specific extension fields.   However nothing was enforcing this behavior.  (fun fact: the very first use of this was incorrect, I used extension instead of evaluator_input when updating the default evaluator, but caught myself in the review.)

Instead have a list of uniquely typed any protos.  This allows any producer to add whatever types
it wants, and the consumers to look for the specific types they want.

# Pros:
Better compos-ability, and less "forced" duplication.  Now various components can simply add information which is required by the other components processing each message.  If there is a unified system, they can use one extension type.  If it's a system composed of various parts, then they simply add and use the protos required by the connected pieces.

This allows data to flow through OM better: if there are systems which are composed together outside of OM core, they don't need custom fields or manipulation. eg, if I pass my match to a well known system with returns assignments from Agones, and requires a specific extension, and I add a layer of processing before hand which requires its own extension, the match doesn't need to be modified to have the required extension when being passed to the Agones allocator.  Or if there's data on tickets which need to flow through OM to the director, they can just be added.

# Cons:
- Very simple use cases for OM are a bit more complex.
- JSON specifically now needs to add a list around the any around the actual data.
- Other languages may need to replicate (the admittedly simple) ext package for reasonable usage.
- Read, Modify, Set operations on extension data are more complicated.


pkg/ext/ext.go
```diff
@@ -0,0 +1,61 @@
...
+package ext
+
+import (
+	"errors"
+	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/any"
+)
+
+var ErrNotFound = errors.New("Message type not found in extensions.")
```
golangcibot: exported var `ErrNotFound` should have comment or be unexported (from `golint`)

pkg/ext/ext.go
```diff
@@ -0,0 +1,61 @@
...
+var ErrMultipleFound = errors.New("Message type found multiple times in extensions.")
```
golangcibot: exported var `ErrMultipleFound` should have comment or be unexported (from `golint`)

pkg/ext/ext.go
```diff
@@ -0,0 +1,61 @@
...
+
+func Unmarshal(exts []*any.Any, pb proto.Message) error {
```
golangcibot: exported function `Unmarshal` should have comment or be unexported (from `golint`)

pkg/ext/ext.go
```diff
@@ -0,0 +1,61 @@
...
+				return ErrMultipleFound
+			}
+			a = ext
+		}
+	}
+
+	return ptypes.UnmarshalAny(a, pb)
+}
+
+func MarshalMany(pbs ...proto.Message) ([]*any.Any, error) {
```
golangcibot: exported function `MarshalMany` should have comment or be unexported (from `golint`)

pkg/ext/ext.go
```diff
@@ -0,0 +1,61 @@
...
+		var err error
+		result[i], err = ptypes.MarshalAny(pb)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return result, nil
+}
+
+func MustMarshalMany(pbs ...proto.Message) []*any.Any {
```
golangcibot: exported function `MustMarshalMany` should have comment or be unexported (from `golint`)

pkg/ext/ext.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+// ErrNotFound indicates that no any values of the given type were found.
+var ErrNotFound = errors.New("Message type not found in extensions.")
+
+// ErrMultipleFound indicates that multiple messages of a the given type were
+// found.  Within Open Match, such extension fields are invalid.
+var ErrMultipleFound = errors.New("Message type found multiple times in extensions.")
+
+// Finds the any value which matches the passed message's type.
```
golangcibot: comment on exported function `Unmarshal` should be of the form `Unmarshal ...` (from `golint`)

Laremere:
Closing in favor of choosing https://github.com/googleforgames/open-match/pull/901

PR #874, opened 2019-10-14
Rename telemetry util function to match what it actually does

yfei1:


internal/telemetry/metrics.go
```diff
@@ -52,13 +52,13 @@ func HistogramWithBounds(name string, description string, unit string, bounds []
...
-func IncrementCounter(ctx context.Context, s *stats.Int64Measure, tags ...tag.Mutator) {
-	IncrementCounterN(ctx, s, 1, tags...)
+// RecordUnitMeasurement records one measurement unit with given tags.
+func RecordUnitMeasurement(ctx context.Context, s *stats.Int64Measure, tags ...tag.Mutator) {
+	RecordNUnitMeasurement(ctx, s, 1, tags...)
 }
 
-// IncrementCounterN record one measurement at once with given tags.
-func IncrementCounterN(ctx context.Context, s *stats.Int64Measure, n int64, tags ...tag.Mutator) {
+// RecordNUnitMeasurement records N measurement units at once with given tags.
```
Laremere: This description is basically a rewording of the function name.  Not knowing what "record", "measurement", "units", or "tags" are in this context, it doesn't help me at all.

PR #884, opened 2019-10-16
Create a privileged service account for Redis and disable THP

yfei1:
THP could lead to potential memory leaks and introduce extra query latency to Redis.

This commit aims to disable Transparent Huge Pages (THP) support on the node that is running Redis. 

This requires having a privileged initContainer to access the `/sys` directory in the node. Our current PodSecurityPolicy doesn't allow creating privileged pods so I created a new PSP along with the role bindings just for Redis.

### DIFF
--- a/install/helm/open-match/templates/podsecuritypolicy.yaml
+++ b/install/helm/open-match/templates/podsecuritypolicy.yaml
@@ -14,10 +14,45 @@
 
 {{- if index .Values "open-match-core" "enabled" }}
 {{- if empty .Values.ci }}
+# om-redis-podsecuritypolicy is the least restricted PSP used to create privileged pods to disable THP in host kernel.
 apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
-  name: om-podsecuritypolicy
+  name: om-redis-podsecuritypolicy
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    {{- include "openmatch.chartmeta" . | nindent 4 }}
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'
+spec:
+  privileged: true
+  allowPrivilegeEscalation: true
+  allowedCapabilities:
+  - '*'
+  volumes:
+  - '*'
+  hostNetwork: true
+  hostPorts:
+  # Redis
+  - min: 6379
+    max: 6379
+  - min: 9121
+    max: 9121
+  hostIPC: true
+  hostPID: true
+  runAsUser:
+    rule: 'RunAsAny'
+  seLinux:
+    rule: 'RunAsAny'
+  supplementalGroups:
+    rule: 'RunAsAny'
+  fsGroup:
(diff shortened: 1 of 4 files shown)


install/helm/open-match/templates/podsecuritypolicy.yaml
```diff
@@ -14,10 +14,45 @@
 
 {{- if index .Values "open-match-core" "enabled" }}
 {{- if empty .Values.ci }}
+# om-redis-podsecuritypolicy is the least restricted PSP used to create privileged pods to disable THP in host kernel.
+apiVersion: policy/v1beta1
```
jeremyje: Doesn't Redis already come with a PSP?

install/helm/open-match/templates/service-account.yaml
```diff
@@ -34,16 +34,27 @@ metadata:
...
-  - om-podsecuritypolicy
+  - om-core-podsecuritypolicy
   verbs:
   - use
+# Grant om-service-role get & list permission for k8s endpoints and pods resources
+# Required for e2e in-cluster testing.
+- apiGroups:
+  - ""
+  resources:
+  - endpoints
```
jeremyje: Do all jobs need this or just a few. Create a new Service Account if the elevation isn't universal. There's things like the demo and swaggerui that probably don't need this.

install/helm/open-match/values.yaml
```diff
@@ -118,14 +118,22 @@ redis:
...
   # https://github.com/helm/charts/tree/master/stable/redis#host-kernel-settings
-  securityContext:
-    sysctls:
-    - name: net.core.somaxconn
-      value: "10000"
+  sysctlImage:
+    enabled: true
+    mountHostSys: true
+    command:
+      - /bin/sh
```
jeremyje: Add a link to https://github.com/helm/charts/tree/master/stable/redis#host-kernel-settings.

install/helm/open-match/templates/podsecuritypolicy.yaml
yfei1: It comes with a service-account along with some configured rolebindings but doesn't have a PSP by default https://github.com/helm/charts/tree/master/stable/redis/templates

install/helm/open-match/templates/podsecuritypolicy.yaml
```diff
@@ -14,10 +14,45 @@
...
+  annotations:
+    {{- include "openmatch.chartmeta" . | nindent 4 }}
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'
+spec:
+  privileged: true
+  allowPrivilegeEscalation: true
+  allowedCapabilities:
+  - '*'
+  volumes:
+  - '*'
```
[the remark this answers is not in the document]
yfei1: redis need a privileged podsecuritypolicy to talk to the host system and therefore requires a bunch of capabilities and volumes permissions, so I explicitly set them to `*` to match them all.

PR #885, opened 2019-10-16
Always enable Prometheus scraping annotations

### LINKED ISSUE #882: Prometheus doesn't work in Open Match two-step installation
`kubectl apply -f 01-open-match-core.yaml && kubectl apply -f open-match-telemetry.yaml` does not activate Prometheus scraping process.

yfei1:
Resolved #882. Always enable Prometheus scraping annotations to fix two steps OM installation. Otherwise Prometheus wouldn't work if we install Promethues and Open Match services separately.


sawagh:
How do you envision setting up the 2nd step prometeus? The current config only sets these values if prometheus is enabled - which makes sense since they are prometheus settings. If we are deploying prometheus after, shouldnt that deployment step take care of enabling these?

yfei1:
I oversimplified the problem :( . Open Match will read from the config to bind its handler with the Prometheus exporter so when we only install open-match-core.yaml, Prometheus binding will be turned off. This issue may require some more changes on the way we are serving Open Match.

PR #886, opened 2019-10-16
Added open-match namespace in yaml installation files 

yfei1:
Auto-create open-match namespace when open-match-core is enabled. This can slightly improve the user experience by `kubectl apply -f 01-open-match-core.yaml -n open-match` directly without having to manually create the namespace.

### DIFF
--- a/install/helm/open-match/templates/service-account.yaml
+++ b/install/helm/open-match/templates/service-account.yaml
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 {{- if index .Values "open-match-core" "enabled" }}
-{{- if index .Values "template" }}
+# Include this namespace only when doing `helm template`.
+# helm 2 use namespace to manage its release so `helm install` with this namespace will be broken.
+{{- if index .Values "usingHelmTemplate" }}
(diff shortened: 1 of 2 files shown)


install/helm/open-match/templates/service-account.yaml
```diff
@@ -13,6 +13,16 @@
 # limitations under the License.
 
 {{- if index .Values "open-match-core" "enabled" }}
+{{- if index .Values "template" }}
```
Laremere: How does the name "template" relate to installing a Namespace?!?

yfei1: We only want to include this namespace when we do `helm template`. Helm2 use namespace to manage its release so `helm install` will be broken if we include this namespace :)

Laremere: Ok, that makes sense, but the name "template" doesn't tell me that.
Maybe name is "usingHelmTemplate"?
Then put a comment explaining this?

yfei1: Done.

PR #891, opened 2019-10-21
Update resource yamls to use k8s v1.16 API

yfei1:
The k8s v1.16 release will stop serving the some deprecated API versions in favor of newer and more stable API versions, this commit: 

Migrate PodSecurityPolicy to use the policy/v1beta1 API, available since v1.10.
Migrate Deployment to use the apps/v1 API, available since v1.9.

### DIFF
--- a/install/helm/open-match/subcharts/open-match-customize/templates/evaluator.yaml
+++ b/install/helm/open-match/subcharts/open-match-customize/templates/evaluator.yaml
@@ -48,12 +48,12 @@ metadata:
   annotations: {{- include "openmatch.chartmeta" . | nindent 4 }}
 spec:
   scaleTargetRef:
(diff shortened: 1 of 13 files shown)


Laremere:
How backwards compatible is this?  Is this something that will only work on k8s 1.6, or will it work for the range of versions it's reasonable to be using?  Please include in PR description.

Laremere:
Have you tested on a 1.16 cluster?  Please confirm before merging.

yfei1:
> Have you tested on a 1.16 cluster? Please confirm before merging.

For now GKE only supports k8s up to v1.14. I think as long as the API is compatible with our current CI version - v1.13.9-gke.v3 we should be good.

Issue #893, opened 2019-10-22
Deprecate Function / Evaluator Harness.


sawagh:
## Harness in Open Match ##

Currently, Open Match provides Harness for Match Function and Evaluator. Here is what these do at a high level:

### Evaluator Harness ###
Creates a GRPC server that implements the Evaluator Service - simplifying the user's component to a simple function call. The harness also handles streaming in and out as needed to be implemented by the Evaluator Service proto simplifying the user function to take a slice of matches and return a slice. 

### Function Harness ###
Same as Evaluator Harness with an added functionality where it connects to MMLogic and hydrates the tickets before calling into the user function.

## Problems with Existing Harness ##

- Harness depend on config maps (mainly because they use internal rpc library to create / connect to grpc server - and that uses Open Match Core values from OM core config map). This makes writing decoupled Evaluator, Match Functions difficult. 
- Harness tries to simplify the method to be written by the user - but creates a tight coupling with itself. Although possible, we today dont have any examples of not using a harness.
- Function Harness implements some core functionality of hydration - which is good but also means that it is now available OOB only for users of golang. 

### Deprecate Evaluator Harness ### 

Deprecate evaluator harness and then provide one or more of the following:

- Default Evaluator creates a grpc Evaluator service - handling the streaming etc. within itself - this would act as a decoupled and complete example implementation for those wanting to customize.
- A package that implements a grpc service - and handles the streaming could still be provided - but this package would then be a 'helper' package that can be implemented as an ecosystem add-on to simplify developer experience. 

### Deprecate Function Harness ###

Same as described for the evaluator, we can deprecate the function harness and provide an indigenous match function and the ecosystem repo can have a package that simplifies starting up a service / grpc streaming etc. 

**Handling Hydration**

However, the existing match function today provides hydration functionality. This is commonly used and we will have to provide a way to do this in Open Match core. One way to do this is to provide some library that solves this - however, that is a language dependent solution. Also, given how commonly we expect this to be used, the existing approach (and hence the library approach) is not efficient since it makes multiple round trips outside of Open Match and back in to hydrate the tickets. Alternatively, we can have Open Match itself optionally hydrate the tickets. This can be specified as an option on the function config on FetchMatches as below:

**Proto Changes**

```
message FunctionConfig {
  string host = 1;
  int32 port = 2;
  Type type = 3;
  bool hydrate_tickets = 4;
  enum Type {
    GRPC = 0;
    REST = 1;
  }
}

```

[...]

PR #894, opened 2019-10-22
Update Protos to implement auto-hydration of tickets

sawagh:


sawagh:
This reflects changes described in #893

sawagh:
We will consider auto hydration change as a part of future changes based on dev experience feedback. closing for now.

PR #901, opened 2019-10-24
Proposal: Make extension a map of string to any

Laremere:
This is a proposal to consider and discuss in PR form (since it was easy enough to just do.)

Replaces the extension field (and match.evaluator_input) with a map of string to any.  The previous concept was that different components would read from specific extension fields.   However nothing was enforcing this behavior.  (fun fact: the very first use of this was incorrect, I used extension instead of evaluator_input when updating the default evaluator, but caught myself in the review.)

Instead have a map of string to proto.  This allows any producer to add whatever values
it wants, and the consumers to look for the specific values they want.

# Pros:
Better compos-ability, and less "forced" duplication.  Now various components can simply add information which is required by the other components processing each message.  If there is a unified system, they can use one extension.  If it's a system composed of various parts, then they simply add and use the protos required by the connected pieces.

This allows data to flow through OM better: if there are systems which are composed together outside of OM core, they don't need custom fields or manipulation. eg, if I pass my match to a well known system with returns assignments from Agones, and requires a specific extension, and I add a layer of processing before hand which requires its own extension, the match doesn't need to be modified to have the required extension when being passed to the Agones allocator.  Or if there's data on tickets which need to flow through OM to the director, they can just be added.

# Cons:
- Very simple use cases for OM are a bit more complex.
- JSON specifically now needs to add a map around the any around the actual data.
- Users who have a single extension type have a bit more work to do.  I think the recommendation should be to use the empty string, "", for such cases.
- Read, Modify, Set operations on extension data are more complicated.


Laremere:
Per discussion in the TSC and offline, going with this option.

PR #906, opened 2019-10-25
Config change first step

yfei1:
This commit is the first step of the Open Match config changes. The goal of the config change is to provide a wrapper for Open Match users such that they don't need to start from a verbose config file to draft their OM settings. The internal config library now requires having a `/app/config/default/matchmaker_config_default.yaml` and `/app/config/override/matchmaker_config_override.yaml` to startup. It will first read from the default yaml, use it's bindings as the default variables and applies the overrides to the config.

### DIFF
--- a/install/helm/open-match/templates/om-configmap-override.yaml
+++ b/install/helm/open-match/templates/om-configmap-override.yaml
@@ -12,10 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+{{- if index .Values "open-match-core" "enabled" }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: customize-configmap
+  name: om-configmap-override
   namespace: {{ .Release.Namespace }}
   annotations: {{- include "openmatch.chartmeta" . | nindent 4 }}
   labels:
@@ -23,18 +24,10 @@ metadata:
     component: config
     release: {{ .Release.Name }}
 data:
+  # TODO: Make this override configmap optional.
+  # Kubernetes doesn't allow having a configmap with empty key.
+  # Using logging setting at here as a placeholder.
   matchmaker_config_override.yaml: |-
-    api:
-      mmlogic:
-        hostname: "{{ .Values.mmlogic.hostName }}"
-        grpcport: "{{ .Values.mmlogic.grpcPort }}"
-      
-      functions:
-        hostname: "{{ .Values.function.hostName }}"
-        grpcport: "{{ .Values.function.grpcPort }}"
-        httpport: "{{ .Values.function.httpPort }}"
-      
-      evaluator:
-        hostname: "{{ .Values.evaluator.hostName }}"
-        grpcport: "{{ .Values.evaluator.grpcPort }}"
-        httpport: "{{ .Values.evaluator.httpPort }}"
+    logging:
+      level: debug
+{{- end }}}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -16,67 +16,54 @@
 package config
 
 import (
+	"fmt"
 	"github.com/fsnotify/fsnotify"
-	"github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
-	"go.opencensus.io/stats"
-	"go.opencensus.io/stats/view"
+	"log"
 )
 
-var (
-	logger = logrus.WithFields(logrus.Fields{
-		"app":       "openmatch",
-		"component": "config",
-	})
-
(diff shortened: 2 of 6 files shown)


internal/config/config.go
```diff
@@ -17,66 +17,52 @@ package config
...
+	// read configs from config/default/matchmaker_config_default.yaml
+	// matchmaker_config_default provides default values for all of the possible tunnable parameters in Open Match
+	dcfg := viper.New()
+	dcfg.SetConfigType("yaml")
+	dcfg.AddConfigPath(".")
+	dcfg.AddConfigPath("config/default")
+	dcfg.SetConfigName("matchmaker_config_default")
+	err = dcfg.ReadInConfig()
+	if err != nil {
+		log.Fatalf("fatal error reading default config file, desc: %s", err.Error())
```
Laremere: Wrap and return the error here, this should fail startup.

internal/config/config.go
```diff
@@ -17,66 +17,52 @@ package config
...
-	// Read in config file using Viper
-	err = cfg.MergeInConfig()
+	cfg.SetConfigType("yaml")
+	cfg.AddConfigPath(".")
+	cfg.AddConfigPath("config/override")
+	cfg.SetConfigName("matchmaker_config_override")
+	err = cfg.ReadInConfig()
 	if err != nil {
-		logger.WithError(err).Fatal("Fatal error reading config file")
+		log.Fatalf("fatal error reading override config file, desc: %s", err.Error())
```
Laremere: same here.

internal/config/config.go
```diff
@@ -17,66 +17,52 @@ package config
...
 	cfg.WatchConfig() // Watch and re-read config file.
 	// Write a log when the configuration changes.
 	cfg.OnConfigChange(func(event fsnotify.Event) {
-		logger.WithFields(logrus.Fields{
-			"filename":  event.Name,
-			"operation": event.Op,
-		}).Info("Server configuration changed.")
+		log.Printf("Server configuration changed, operation: %v, filename: %s", event.Op, event.Name)
 	})
 	return cfg, err
```
Laremere: Returning a config or error should be mutually exclusive.  return cfg, nil, if err is known to be nil.

internal/config/config.go
```diff
@@ -17,66 +17,52 @@ package config
...
 	// More details about Open Match's use of Kubernetes ConfigMaps at:
 	// https://open-match.dev/open-match/issues/42
 	cfg.WatchConfig() // Watch and re-read config file.
 	// Write a log when the configuration changes.
 	cfg.OnConfigChange(func(event fsnotify.Event) {
-		logger.WithFields(logrus.Fields{
-			"filename":  event.Name,
-			"operation": event.Op,
-		}).Info("Server configuration changed.")
+		log.Printf("Server configuration changed, operation: %v, filename: %s", event.Op, event.Name)
```
Laremere: Why are you removing the use of logrus fields?

install/helm/open-match/templates/om-configmap-override.yaml
```diff
@@ -0,0 +1,100 @@
...
+metadata:
+  name: om-configmap-override
+  namespace: {{ .Release.Namespace }}
+  annotations: {{- include "openmatch.chartmeta" . | nindent 4 }}
+  labels:
+    app: {{ template "openmatch.name" . }}
+    component: config
+    release: {{ .Release.Name }}
+data:
+  matchmaker_config_override.yaml: |-
```
Laremere: The override should be nearly empty.  Why are all of the values copied here?

internal/config/config.go
```diff
@@ -17,66 +17,52 @@ package config
...
+	// read configs from config/default/matchmaker_config_default.yaml
+	// matchmaker_config_default provides default values for all of the possible tunnable parameters in Open Match
+	dcfg := viper.New()
+	dcfg.SetConfigType("yaml")
+	dcfg.AddConfigPath(".")
+	dcfg.AddConfigPath("config/default")
+	dcfg.SetConfigName("matchmaker_config_default")
+	err = dcfg.ReadInConfig()
+	if err != nil {
+		return nil, fmt.Errorf("fatal error reading default config file, desc: %s", err.Error())
```
golangcibot: undeclared name: `fmt` (from `typecheck`)

internal/config/config.go
```diff
@@ -16,67 +16,54 @@
...
```
Laremere: Use %w, not %s.  It is the proper way to do error wrapping.

PR #913, opened 2019-10-29
Implement Tutorial for authoring a basic match function

sawagh:


tutorials/matchmaker101/matchmaker.yaml
```diff
@@ -0,0 +1,151 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This file contains the sample minimal pod definitions for all of the components that one may need to use Open Match as a match maker.
+# You can find the same pod definitions within the sub-folders under the /tutorials/ directory
+# Run `kubectl apply -f matchmaker.yaml` to deploy these definitions.
+
+apiVersion: v1
```
yfei1: From the tutorials' perspective, is it possible keep using minimal yaml definitions for the director, frontend, matchfunction, and evaluator? It looks a bit weird by using Pod and Deployment resources interchangeably in this yaml.

tutorials/matchmaker101/matchmaker.yaml
```diff
@@ -0,0 +1,151 @@
...
+    app: mm101-tutorial
+    component: evaluator
+  clusterIP: None
+  type: ClusterIP
+  ports:
+  - name: grpc
+    protocol: TCP
+    port: 50508
+---
+apiVersion: autoscaling/v1
```
yfei1: The PodAutoscaler is one of the production best practices. We could potentially skip this part to keep things easy to follow.

sawagh: Removed! Thanks.

tutorials/matchmaker101/matchmaker.yaml
sawagh: Noted. I ll not block this one on the change given that it works and is transparent to actual tutorial steps but will update it to use service / pod in the next iteration.

tutorials/matchmaker101/director/main.go
```diff
@@ -0,0 +1,150 @@
...
+//
+// The Director can be configured to make a single fetch for all the profiles
+// or can keep polling forever at a configured interval.
+
+var (
+	omBackendEndpoint       = "om-backend.open-match.svc.cluster.local:50505"
+	functionHostName        = "mm101-tutorial-matchfunction.mm101-tutorial.svc.cluster.local"
+	functionPort      int32 = 50502
+	interval                = 5000 // Millisecond duration for the polling interval.
+	fetchForever            = true // Fetch once if false, forever if true.
```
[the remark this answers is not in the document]
Laremere: Comments say the director can be configured to loop forever or not... is the documentation wrong or the implementation?

tutorials/matchmaker101/director/main.go
```diff
@@ -0,0 +1,150 @@
...
+					return
+				}
+			}(&wg, p)
+		}
+
+		wg.Wait()
+		if !fetchForever {
+			break
+		}
+		time.Sleep(time.Millisecond * time.Duration(interval))
```
Laremere: use a time.Ticker, as it will lead to a consistent interval time without varying because of how long the fetch matches and assignments takes.

Laremere: (unless it takes too long, in which case ticker will drop missed intervals.)

tutorials/matchmaker101/frontend/ticket.go
```diff
@@ -0,0 +1,57 @@
...
+	avgRating float64 = 40
+	stdDev    float64 = 20
+)
+
+// Ticket generates a Ticket with data using the package configuration.
+func makeTicket() *pb.Ticket {
+	ticket := &pb.Ticket{
+		SearchFields: &pb.SearchFields{
+			DoubleArgs: map[string]float64{
+				e2e.DoubleArgMMR: normalDist(avgRating, minRating, maxRating, stdDev),
```
Laremere: Probably better to use a constant defined just for the tutorial?  (This is also using something from internal...)

tutorials/matchmaker101/matchfunction/mmf/matchfunction.go
```diff
@@ -0,0 +1,157 @@
...
+		if err := stream.Send(&pb.RunResponse{Proposal: proposal}); err != nil {
+			log.Printf("Failed to stream proposals to Open Match, got %w", err)
+			return err
+		}
+	}
+
+	return nil
+}
+
+func makeMatches(p *pb.MatchProfile, poolTickets map[string][]*pb.Ticket) ([]*pb.Match, error) {
```
Laremere: This match logic seems complicated, the roster use doesn't seem intuitive to me.  Would be interested in discussing alternatives.

tutorials/matchmaker101/matchmaker.yaml
```diff
@@ -0,0 +1,137 @@
...
+    app: mm101-tutorial
+    component: matchfunction
+  clusterIP: None
+  type: ClusterIP
+  ports:
+  - name: grpc
+    protocol: TCP
+    port: 50502
+---
+apiVersion: extensions/v1beta1
```
Laremere: We switched to using non-beta apiVersion for deployments.

tutorials/matchmaker101/director/README.md
```diff
@@ -0,0 +1,25 @@
+This folder provides a sample Director for Open Match Matchmaker 101 Tutorial.
+<TODO - Update the README with details of the Director and the steps that need
+to be run before executing the commamds below>
+
+Run the below steps in this folder to set up the Director.
+
+Step1: Specify your Registry URL.
+```
+REGISTRY=[YOUR_REGISTRY_URL]
```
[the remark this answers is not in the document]
yfei1: I intentionally left the URL as a placeholder since users may use docker hub instead

tutorials/matchmaker101/director/director.yaml
```diff
@@ -15,11 +15,11 @@
 apiVersion: v1
 kind: Pod
```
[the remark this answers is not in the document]
yfei1: Deployment is a higher level concept than Pod. From my perspective, I think using Pod in a tutorial for simplicity purpose may make more sense. wdyt?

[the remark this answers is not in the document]
sawagh: The goal is to keep it as simple as possible. I plan to remove the other deployment and make that a single pod service.

tutorials/matchmaker101/matchmaker.yaml
yfei1: FYI simply use `apps/v1` instead for deployment apiVersion

tutorials/matchmaker101/director/main.go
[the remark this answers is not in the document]
sawagh: Right now - this is just a convenience const - if you wanna modify the behavior without understanding the code, change the const and rebuild the director. The next iteration based we can either make it set via env / config map / json file. Keeping it simple to begin with.

tutorials/matchmaker101/director/main.go
sawagh: Basic tutorial - really easy to understand the logs if you follow a fetch-assign-wait pattern. No requirement to tick at consistent interval time. If fetch / assign blocks for whatever reason, I would rather have this loop block and know there is a problem.

tutorials/matchmaker101/director/README.md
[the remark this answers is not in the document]
sawagh: Yeah. Lets not default the tutorial to GCP.

tutorials/matchmaker101/matchfunction/mmf/matchfunction.go
sawagh: Good feedback. I had ported it from a familiar case - but If it isnt intuitive to you, it aint gonna be intuitive for a 101 tutorial user. I've updated this to just pick a certain number from each pool for starters.

tutorials/matchmaker101/director/main.go
```diff
@@ -0,0 +1,142 @@
...
+				}
+			}(&wg, p)
+		}
+
+		wg.Wait()
+		// Wait for an interval before polling again.
+		time.Sleep(time.Millisecond * 5000)
+	}
+
+	// Block forever to enable inspecting state.
```
Laremere: This is now never invoked.

tutorials/matchmaker101/director/main.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	defer conn.Close()
+	be := pb.NewBackendClient(conn)
+
+	// Generate the profiles to fetch matches for.
+	profiles := generateProfiles()
+	log.Printf("Fetching matches for %v profiles", len(profiles))
+
+	for {
```
Laremere:
```
for range time.Tick(time.Second * 5) {
```
Is simpler than sleeping, leads to more consistent behavior, and is more idiomatic.

tutorials/matchmaker101/director/director.yaml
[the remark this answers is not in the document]
Laremere: This goes against best practices: https://kubernetes.io/docs/concepts/configuration/overview/#naked-pods-vs-replicasets-deployments-and-jobs

All of these should be deployments.

tutorials/matchmaker101/director/README.md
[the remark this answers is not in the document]
Laremere: This seems fine, as long as there is some meta section which specifies how to determine your registry.

tutorials/matchmaker101/matchfunction/mmf/matchfunction.go
Laremere: This seems fine for a first iteration.  Thanks!

tutorials/matchmaker101/director/main.go
sawagh: Fixed.

tutorials/matchmaker101/director/main.go
sawagh: Reflected this.

tutorials/matchmaker101/director/director.yaml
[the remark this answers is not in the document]
sawagh: For a tutorial execution, I prefer to keep the concept count simpler (and I dont expect the tutorial to be used for use cases where node failure resiliency is needed)

sawagh:
@googlebot rescan

PR #916, opened 2019-10-29
Split up override configmap from open-match-core static yaml

### LINKED ISSUE #680: Split Open Match configuration into infrastructure v/s application configuration
Currently, All of Open Match configuration is bundled together and is stored in a config file accessed using Viper. We should consider splitting out the application configuration from the infrastructure configuration. Potentially consider enabling API access to set / update Application configuration?

yfei1:
This commit removed the override-configmap from `01-open-match-core.yaml` and had `make install/yaml` step generate a new yaml template solely for the override configmap.

Resolves #680

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -341,6 +341,7 @@ install-large-chart: install-chart-prerequisite build/toolchain/bin/helm$(EXE_EX
 		--namespace=$(OPEN_MATCH_KUBERNETES_NAMESPACE) \
 		--set global.image.registry=$(REGISTRY) \
 		--set global.image.tag=$(TAG) \
+		--set open-match-override.enabled=true \
 		--set global.telemetry.grafana.enabled=true \
 		--set global.telemetry.jaeger.enabled=true \
 		--set global.telemetry.prometheus.enabled=true \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -401,7 +405,7 @@ delete-chart: build/toolchain/bin/helm$(EXE_EXTENSION) build/toolchain/bin/kubec
 	-$(KUBECTL) --ignore-not-found=true delete crd prometheusrules.monitoring.coreos.com
 	-$(KUBECTL) delete namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE)
 
-install/yaml/: update-chart-deps install/yaml/install.yaml install/yaml/01-open-match-core.yaml install/yaml/02-open-match-demo.yaml install/yaml/03-prometheus-chart.yaml install/yaml/04-grafana-chart.yaml install/yaml/05-jaeger-chart.yaml
```
Laremere: I think we should avoid renaming these yaml files whenever possible.  It's liable to lead to unessisary broken links.

Makefile
```diff
@@ -424,17 +428,12 @@ install/yaml/02-open-match-demo.yaml: build/toolchain/bin/helm$(EXE_EXTENSION)
...
 	$(HELM) template --name $(OPEN_MATCH_RELEASE_NAME) --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) \
-		--set global.image.registry=$(REGISTRY) \
-		--set global.image.tag=$(TAG) \
-		--set open-match-customize.enabled=false \
-		--set open-match-demo.enabled=false \
-		--set open-match-core.enabled=false \
-		--set global.telemetry.prometheus.enabled=true \
-		--set usingHelmTemplate=true \
-		install/helm/open-match > install/yaml/03-prometheus-chart.yaml
+		--set open-match-override.enabled=true \
```
Laremere: Shouldn't core be false too?  Or was that switched to default false?  (If so, this should have been cleaned up separately?)

yfei1: `-x templates/om-configmap-override.yaml` only template this specified file

Makefile
yfei1: Done.

PR #918, opened 2019-10-30
Fix a race where assigned tickets still get returned by Query Tickets

sawagh:
Currently in query tickets, we fetch tickets from redis and then check the ignore list. When a ticket is assigned, it gets removed from index and then from index list. Here is a race:
1. Ticket A gets sent as result - gets added to ignore list before it is returned as result.
2. Consider the concurrent requests come in now:
    Request1: Sets assignment received for Ticket A.
    Requset2: Query Tickets call received
3. Request1 fetches tickets from index - this will include A (Since it is still in index) 
4. Request2 removes player from index and ignore list. 
5. Request1 fetches ignore list - player A is not in there - so is returned as a result.

Fix here is to fetch ignore list BEFORE fetching the index. In the worst case, A will be in ignore list and not in index - which is OK since we only consider tickets from index.


sawagh:
@googlebot rescan

PR #920, opened 2019-10-30
Install override yaml file by default when using helm upgrade

yfei1:
#916 introduces a new behavior to our Makefile. Simply running `make install-xx-chart` will now be blocked because of om core services couldn't find the override yaml file. This PR added the override flag to all `make install-xx-chart` commands to simplify the developer experience.


Laremere:
Please describe what this affect has and how it is solving a problem in the pr description.  Thanks!

yfei1:
> Please describe what this affect has and how it is solving a problem in the pr description. Thanks!

Done, thanks.

PR #923, opened 2019-10-31
Move Evaluator to tutorial namespace and enable tutorial configuration for Open Match

sawagh:


tutorials/matchmaker101/matchmaker.yaml
```diff
@@ -112,3 +112,25 @@ spec:
...
     port: 50508
 ---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: om-configmap-override
+  namespace: open-match
+  annotations:
+    chart: open-match-0.0.0-dev
+    heritage: Helm
```
yfei1: could you remove the `annotations` part and `labels.release` from the configmap? These fields are popped up by helm

PR #927, opened 2019-10-31
Add a template for a basic OM based Matchmaker. The tutorials will use this as starting point.

sawagh:
This provides an empty template for the matchmaker tutorials. This is completely deployable. The Tutorial experience would be to deploy this as a starting point - no matches will be generated. Then each tutorial will get into steps of modifying this for their scenarios. 

Note that for now, the actual tutorial solutions replicate this entire code. Thats OK as a starting point in interest of having a simple isolated solution for each tutorial without having dependencies to any shared code.


sawagh:
This PR is identical to MM101 - but just modifications in:
ticket.go - provide an empty ticket generation helper.
profile.go - provide an empty profile generation helper
matchfunction.go - provide an empty scaffold for makematches()

PR #931, opened 2019-10-31
Add Cacher, which invalidates a cache when config changes

Laremere:
I will use this in the evaluator_client, where we want to re-use a client if possible, but get a new client if the used values change.  This solves the generic problem on a meta level, instead of having to manually remember and compare the config values used.  It also prevents programming errors where a new config value is read, but the code doesn't properly detect if it has changed.

The ForceReset method will be used when the evaluator client has an error, so that the system will recover from a client which is stuck in an error state on the next call.

I anticipate there will be other places to use this inside open match.


internal/config/cacher_test.go
```diff
@@ -0,0 +1,271 @@
...
+	if v != "foo" {
+		t.Errorf("Expected foo, got %v", v)
+	}
+	if err != nil {
+		t.Errorf("Expected nil, got %v", err)
+	}
+
+	// Sanity check: value doesn't change because config hasn't.
+	v, err = c.Get(func(cfg View) (interface{}, error) {
+		return "bar", nil
```
golangcibot: string `bar` has 3 occurrences, make it a constant (from `goconst`)

internal/config/cacher_test.go
```diff
@@ -0,0 +1,271 @@
...
+		})
+	}
+}
+
+func Test_Get_Error(t *testing.T) {
+	cfg := viper.New()
+	c := NewCacher(cfg)
+
+	v, err := c.Get(func(cfg View) (interface{}, error) {
+		return "foo", fmt.Errorf("bad")
```
golangcibot: string `foo` has 6 occurrences, make it a constant (from `goconst`)

internal/config/cacher.go
```diff
@@ -0,0 +1,203 @@
...
+// ForceReset causes Cacher to forget the remembered value.  The next call to
+// Get will again construct a new value.
+func (c *Cacher) ForceReset() {
+	c.m.Lock()
+	defer c.m.Unlock()
+	c.r = nil
+	c.v = nil
+}
+
+type rememberingView struct {
```
sawagh: cachedView? (I assume rememberingView implies the view that was last cached? - adding a comment here would make it easier to understand)

Laremere: It's not really a cache because it's not returning any of the values it remembers.  Changed to viewChangeDetector with a comment.

internal/config/cacher.go
```diff
@@ -0,0 +1,203 @@
...
+		cfg: cfg,
+	}
+}
+
+// Get will call f the first time it is called.  It remembers which values are
+// read from the provided View, and the value returned.  When Get is next
+// called, it will check if any of the used values have changed.  If they
+// haven't, it will return the cache'd value.  Otherwise it will call f again.
+// If f returns an error, the value will not be cached or returned.
+func (c *Cacher) Get(f func(cfg View) (interface{}, error)) (interface{}, error) {
```
sawagh: Can we declare a type for this function and document what you expect this function written by the caller to do?

Laremere: How's this for documentation. (On Cacher and on *cacher.Get)

internal/config/cacher_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+	if v != nil {
+		t.Errorf("Expected value to be nil")
+	}
+	if err.Error() != "bad" {
+		t.Errorf("Expected error \"bad\", got %v", err)
+	}
+
+	// No config values changed, still trying.
+	returnError = false // Emulating the enviornment changing.
```
golangcibot: `enviornment` is a misspelling of `environment` (from `misspell`)
```suggestion

```

PR #933, opened 2019-10-31
Refactor evaluator client in the synchronizer

Laremere:
This fixes a number of bugs:
- If it's a grpc connection, the evaluator client also creates an http connection.
- Even if it's an http connection, it never sets it to use that http connection, always trying to recreate the client and failing.
- The http evaluator code does not use the http client which it created.
- If the config is updated, the old evaluator connection details are still used.
- If the client errors, it may still use a broken client.

Refactors the file into several completely separate components:
- A grpc based client.
- An http based client.
- A differed client which selects between creating a grpc or http client, and will detect changes to the config and recreate the client.

As a result of these changes, the type of connection is explicit based on config.  If a grpc port is present, it will always use that connection, never trying to create an http client.

The actual code to create clients and make requests is mostly unchanged.


internal/app/synchronizer/evaluator_client.go
```diff
@@ -51,73 +42,68 @@ var (
...
+	evaluator pb.EvaluatorClient
+}
+
+func newGrpcEvaluator(cfg config.View) (evaluator, error) {
+	grpcAddr := fmt.Sprintf("%s:%d", cfg.GetString("api.evaluator.hostname"), cfg.GetInt64("api.evaluator.grpcport"))
+	conn, err := rpc.GRPCClientFromEndpoint(cfg, grpcAddr)
 	if err != nil {
-		evaluatorClientLogger.WithError(err).Errorf("failed to get a HTTP client from the endpoint %v", httpAddr)
-		return err
+		return nil, fmt.Errorf("Failed to create grpc evaluator client: %w", conn)
```
golangcibot: printf: Errorf format %w has arg conn of wrong type *google.golang.org/grpc.ClientConn (from `govet`)

internal/app/synchronizer/evaluator_client.go
```diff
@@ -51,73 +42,68 @@ var (
...
+type evaluator interface {
+	evaluate(context.Context, []*pb.Match) ([]*pb.Match, error)
 }
 
-// evaluate method triggers execution of user evaluation function. It initializes the configured
-// GRPC or HTTP client. If the client was already initialized, initialize is a no-op.
-func (ec *evaluatorClient) evaluate(ctx context.Context, proposals []*pb.Match) (result []*pb.Match, err error) {
-	if err := ec.initialize(); err != nil {
-		return nil, err
+var noEvaluatorType = errors.New("Unable to determine evaluator type.  Either api.evaluator.grpcport or api.evaluator.httpport must be specified in the config.")
```
golangcibot: error var noEvaluatorType should have name of the form errFoo (from `golint`)

internal/app/synchronizer/evaluator_client.go
```diff
@@ -149,7 +135,29 @@ func (ec *evaluatorClient) grpcEvaluate(ctx context.Context, proposals []*pb.Mat
 	return results, nil
 }
 
-func (ec *evaluatorClient) httpEvaluate(ctx context.Context, proposals []*pb.Match) ([]*pb.Match, error) {
+type httpEvaluatorClient struct {
+	httpClient *http.Client
+	baseURL    string
+}
+
+func newHttpEvaluator(cfg config.View) (evaluator, error) {
```
golangcibot: func newHttpEvaluator should be newHTTPEvaluator (from `golint`)

internal/app/synchronizer/evaluator_client.go
```diff
@@ -31,93 +32,74 @@ import (
...
+type evaluator interface {
+	evaluate(context.Context, []*pb.Match) ([]*pb.Match, error)
 }
 
-// evaluate method triggers execution of user evaluation function. It initializes the configured
-// GRPC or HTTP client. If the client was already initialized, initialize is a no-op.
-func (ec *evaluatorClient) evaluate(ctx context.Context, proposals []*pb.Match) (result []*pb.Match, err error) {
-	if err := ec.initialize(); err != nil {
-		return nil, err
+var errNoEvaluatorType = errors.New("Unable to determine evaluator type.  Either api.evaluator.grpcport or api.evaluator.httpport must be specified in the config.")
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

internal/app/synchronizer/evaluator_client.go
```diff
@@ -31,93 +32,74 @@ import (
...
-// evaluate method triggers execution of user evaluation function. It initializes the configured
-// GRPC or HTTP client. If the client was already initialized, initialize is a no-op.
-func (ec *evaluatorClient) evaluate(ctx context.Context, proposals []*pb.Match) (result []*pb.Match, err error) {
-	if err := ec.initialize(); err != nil {
-		return nil, err
+var errNoEvaluatorType = errors.New("unable to determine evaluator type, either api.evaluator.grpcport or api.evaluator.httpport must be specified in the config")
+
+func newEvaluator(cfg config.View) evaluator {
+	newInstance := func(cfg config.View) (interface{}, error) {
+		if cfg.IsSet("api.evaluator.grpcport") {
```
sawagh: Add a comment indicating that GRPC is preferred over HTTP? :) so in future, somone does not assume that it is an oversight?

internal/app/synchronizer/evaluator_client.go
```diff
@@ -31,93 +32,74 @@ import (
...
+			return newGrpcEvaluator(cfg)
+		}
+		if cfg.IsSet("api.evaluator.httpport") {
+			return newHTTPEvaluator(cfg)
+		}
+		return nil, errNoEvaluatorType
```
sawagh: Will this surface to the caller as InternalError? We should a grpc error code here? Unavailable? invalid argument?

Laremere: Good catch!  Changed to FailedPrecondition.

internal/app/synchronizer/evaluator_client.go
Laremere: done.

internal/app/synchronizer/evaluator_client.go
```diff
@@ -31,93 +31,75 @@ import (
...
+type evaluator interface {
+	evaluate(context.Context, []*pb.Match) ([]*pb.Match, error)
 }
 
-// evaluate method triggers execution of user evaluation function. It initializes the configured
-// GRPC or HTTP client. If the client was already initialized, initialize is a no-op.
-func (ec *evaluatorClient) evaluate(ctx context.Context, proposals []*pb.Match) (result []*pb.Match, err error) {
-	if err := ec.initialize(); err != nil {
-		return nil, err
+var errNoEvaluatorType = grpc.Errorf.New(codes.FailedPrecondition, "unable to determine evaluator type, either api.evaluator.grpcport or api.evaluator.httpport must be specified in the config")
```
golangcibot: undeclared name: `grpc` (from `typecheck`)

PR #938, opened 2019-11-02
Bumped helm version and dependencies versions for k8s 1.16 support

yfei1:
kubectl 1.14.3	and redis chart 8.0.9 was using old kubernetes API surface and doesn't support k8s 1.16. This commit bumped the dependency versions for k8s 1.16 support


sawagh:
Do we need to port this to the v0.8RC?

yfei1:
> Do we need to port this to the v0.8RC?

@sawagh The Redis update is required if we wanna support k8s 1.16 in v0.8

PR #941, opened 2019-11-04
Disable Redis initContainer by default

yfei1:
The THP related configuration is only needed when the node running Open Match is using Linux based system. This commit disable it by default to make sure Open Match can be initialized correctly under non-Linux enviroment.


sawagh:
What is the impact of enabling / disabling this? 

If disabled by default, how critical is it for a Linux environment user to turn this on - and if it is, how do we communicate that? My only concern is, if these are critical settings, we need to have a discoverability story for them?

yfei1:
This setting only affects Linux users and will keep Open Match from initializing if the users is testing Open Match under Windows/Mac environment (Since the init steps for Redis will fail).

I think this setup doesn't really matter for ones who just wanna understand how Open Match works, but it is critical for ones who wanna use Open Match in production under Linux environment as it affects Redis's query latency and may result in memory leaks.

sawagh:
Lets track the production guide update in an issue for now.

PR #943, opened 2019-11-04
Optimize deployment workflow

yfei1:
When `helm upgrade` failed, helm won't clean up all of the deployed resources properly. The resources mentioned above which are not namespace scoped will be leaked and prevent us from running `install-chart` again. This commit added a `kubectl delete` operation in `make delete-chart` to force delete the PodSecurityPolicy, ClusterRole, and ClusterRolebindings from the cluster.

Also fixed a Makefile target when deploying scale chart.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -346,18 +346,26 @@ install-large-chart: install-chart-prerequisite build/toolchain/bin/helm$(EXE_EX
 install-chart: install-chart-prerequisite build/toolchain/bin/helm$(EXE_EXTENSION) install/helm/open-match/secrets/
 	$(HELM) upgrade $(OPEN_MATCH_RELEASE_NAME) $(HELM_UPGRADE_FLAGS) install/helm/open-match $(HELM_IMAGE_FLAGS)
 
-install-scale-chart: build/toolchain/bin/helm$(EXE_EXTENSION) install/helm/open-match/secrets/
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -346,18 +346,26 @@ install-large-chart: install-chart-prerequisite build/toolchain/bin/helm$(EXE_EX
...
+		--set open-match-scale.enabled=false \
 		--set open-match-customize.enabled=true \
 		--set open-match-customize.function.image=openmatch-mmf-go-rosterbased\
 		--set global.telemetry.grafana.enabled=true \
 		--set global.telemetry.jaeger.enabled=true \
 		--set global.telemetry.prometheus.enabled=true \
+		--set global.logging.rpc.enabled=false \
+		--set global.gcpProjectId=$(GCP_PROJECT_ID)
+	$(HELM) template $(OPEN_MATCH_RELEASE_NAME)-scale  install/helm/open-match $(HELM_TEMPLATE_FLAGS) $(HELM_IMAGE_FLAGS) \
+		--set open-match-core.enabled=false \
```
sawagh: Not in this PR but can we have a simpler model of - all disabled by default so these steps only enable the required ones? more readable?

yfei1: Sure, will do it later this week.

PR #955, opened 2019-11-08
Modify statestore API to improve context propagation

yfei1:
OpenConcensus's tracing story is based on context propagation - it will inject a span into the current context and use this modified context for tracing. However, this method doesn't work quite well with Open Match's current statestore implementation - For example, instead of creating a span with two sub-spans, `frontend.CreateTicket` will create two different spans under different contexts, which doesn't make sense.
```
# A span with two-subspans
frontend.CreateTicket(A)
 |                                                                         
 -------- statetstore.CreateTicket(A)              
 |
 -------- statestore.IndexTicket(A)

# Two spans with different contexts
frontend.CreateTicket(A) ---- statestore.CreateTicket(A)
frontend.CreateTicket(A) ---- statestore.IndexTicket(A)              
```

This commit have all statestore APIs return a context to improve context propagation for tracing purpose.

### DIFF
--- a/internal/statestore/public.go
+++ b/internal/statestore/public.go
@@ -28,34 +28,34 @@ type Service interface {
 	HealthCheck(ctx context.Context) error
 
 	// CreateTicket creates a new Ticket in the state storage. If the id already exists, it will be overwritten.
-	CreateTicket(ctx context.Context, ticket *pb.Ticket) error
+	CreateTicket(ctx context.Context, ticket *pb.Ticket) (context.Context, error)
 
 	// GetTicket gets the Ticket with the specified id from state storage. This method fails if the Ticket does not exist.
-	GetTicket(ctx context.Context, id string) (*pb.Ticket, error)
+	GetTicket(ctx context.Context, id string) (context.Context, *pb.Ticket, error)
 
 	// DeleteTicket removes the Ticket with the specified id from state storage. This method succeeds if the Ticket does not exist.
-	DeleteTicket(ctx context.Context, id string) error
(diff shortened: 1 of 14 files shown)


Laremere:
Thanks for the great PR description.

This doesn't feel like the right way to solve this however.  Returning a context that has been wrapped to use for the consecutive calls is not a pattern I've ever seen with context, and feels like an anti-pattern.

The root cause of the problem is that there isn't, eg, a span for the entire CreateTicket call.  From our discussion before, it seems that we have grpc set up to create these spans, but only when the calling grpc client itself has a span on it.  Have you looked into adding spans from the client code to see if, eg, statestore.CreateTicket and statestore.IndexTicket are properly nested under the frontend.CreateTicket call?

internal/statestore/public.go
```diff
@@ -28,34 +28,34 @@ type Service interface {
 	HealthCheck(ctx context.Context) error
 
 	// CreateTicket creates a new Ticket in the state storage. If the id already exists, it will be overwritten.
-	CreateTicket(ctx context.Context, ticket *pb.Ticket) error
+	CreateTicket(ctx context.Context, ticket *pb.Ticket) (context.Context, error)
```
sawagh: I dont think returning the context from state-store APIs is the right pattern. Is this just so you can have the instrumented context outside of this API? If so, then the instrumentation should be done earlier in the stack than passing the instrumented context back out from this.

yfei1: Yea it seems like grpc ecosystem provides a tracing middleware. I’ll try that out and see if it could improve the context propagation

PR #959, opened 2019-11-08
Have MMF and Evaluator in customize chart use different configs

yfei1:
This commit completely decouple the customize chart from using the default configmap in `open-match` namespace. As a workaround, it defines a `default` and `override` yaml in the same configmap and mount it to two different volumes - which is required by the harness.

`make install-chart` will now install a demo chart with its MMF to the `open-match-demo` namespace and `open-match-core` to `open-match` namespace.


sawagh:
Seems ok for now to simplify the Getting Started demo experience. We should revisit the charts and what their uses are and the coupling (the basic install-chart / install-large-chart needing demo installation etc.) - Out of scope for this PR.

PR #961, opened 2019-11-11
Add a tutorial for using the default evaluator

sawagh:
This tutorial demonstrates the usage of the evaluation_input extension to provide a quality score for a match from the match function that is then used by the default evaluator for decolliding matches with overlapping Tickets.


sawagh:
Apart from the namespace being different, the only key files that have changed in this PR are:
matchfunction.go - Adds score calculation to calculate quality score for a match. 
ticket.go - Picks 2 game modes instead of one - and adds queue ticket time.

PR #964, opened 2019-11-12
Add logging to the default evaluator

sawagh:
This surfaces the rejected matches, with information about what matches did what ticket collide with scores of both matches. This helps debug, correlate rejected matches with match function runs.


examples/evaluator/golang/simple/evaluate/evaluator.go
```diff
@@ -79,20 +86,35 @@ func Evaluate(p *harness.EvaluatorParams) ([]*pb.Match, error) {
...
 	for _, t := range m.match.GetTickets() {
-		if _, ok := d.ticketsUsed[t.Id]; ok {
+		if cm, ok := d.ticketsUsed[t.Id]; ok {
+			logger.WithFields(logrus.Fields{
+				"match_id":              m.match.GetMatchId(),
+				"ticket_id":             t.GetId(),
+				"match_score":           m.inp.GetScore(),
+				"colliding_match_id":    cm.id,
+				"colliding_match_score": cm.score,
+			}).Info("Higher quality match with colliding ticket found. Rejecting match.")
```
yfei1: Maybe `logger.Debug`?

sawagh: lets keep a higher tracing level for now specially for things that evaluator drops. I would prefer the demo executions to show this log statement by default

PR #977, opened 2019-11-14
Generate protos using Dockerfile

### LINKED ISSUE #491: CI should fail if running presubmit changes files
Basically, go.mod and the proto generated files can experience version skew because the CI doesn't detect if the files need to be updated.  It should block submit.  Otherwise running presubmit before committing a different change will result in unrelated files being changed.

yfei1:
This is an attempt to harden our current verbose Makefile by generating the proto files using an Docker image.

Also partially resolves #491 if we added the `make cmp-proto` call to our CI pipeline.


Laremere:
I have done this in the ecosystem repo.  We should unify implementations.

yfei1:
> I have done this in the ecosystem repo. We should unify implementations.

Totally agree. We should discuss this offline after v0.8.

PR #989, opened 2019-11-16
Added a fixed development password for Redis

### LINKED ISSUE #642: install-chart cannot be re-applied with Redis password enabled.
`make install-chart` performs a `helm upgrade`. It appears that after deploying a cluster with a Redis password it's changing because a reapplication makes Open Match unhealthy.

yfei1:
Resolved #642 

Helm will generate random password for Redis without specifying a password. Therefore when we re-deploy Open Match into a password-enabled cluster, the core services won't be able to talk to Redis anymore because of the change of password. This commit added a fixed development password for Redis.

The new development workflow now no longer requires deleting the previous release. `make push-images install-chart` would work.


Laremere:
How does this ensure that the password is only used for development, and won't get use in production?

Does this problem still exist in production (eg, I can just change an autoscaler limit because it would change the redis password?)

yfei1:
> How does this ensure that the password is only used for development, and won't get use in production?
> 
I manually set the password in our Makefile and the default password is a random string provided by helm template if not set.

> Does this problem still exist in production (eg, I can just change an autoscaler limit because it would change the redis password?)

Yes and no, merely changing the autoscaler limit won't redeploy redis and won't affect the password. However, for commands like `helm upgrade open-match`, since helm will generate a new password for Redis, Open Match services will fail to connect to Redis and falls into `CrashLoopBackoff` status.

We should cover corner cases like this in Open Match production or rolling upgrade guides.

PR #992, opened 2019-11-19
Update Scale package to sync with the latest config and API changes

yfei1:
Cherry-picked some changes from my private branch to master to better replicate the scale testing scenarios.


examples/scale/profiles/scaleprofiles.go
```diff
@@ -0,0 +1,107 @@
...
+
+package profiles
+
+import (
+	"fmt"
+	"math"
+
+	"open-match.dev/open-match/pkg/pb"
+)
+
```
sawagh: Add some documentation on how many profiles the current implementation generates, how to tweak these to scale the number up / down.

examples/scale/tickets/playlistticket.go
```diff
@@ -0,0 +1,125 @@
...
+		"mmr.playlist11",
+		"mmr.playlist12",
+		"mmr.playlist13",
+		"mmr.playlist14",
+		"mmr.playlist15",
+	}
+)
+
+// PlaylistTicket generates a ticket for profile scale testing
+func PlaylistTicket() *pb.Ticket {
```
sawagh: Where does this get called? Ideally you want this to be configurable right? Also, considered the ticket to be the same for any scale test? Think like.. Just one ticket generation - but the different tests can use profiles that use different aspects of that Ticket?

examples/scale/tickets/tickets.go
```diff
@@ -16,57 +16,112 @@ package tickets
...
 
-// Ticket generates a ticket based on the config for scale testing
-func Ticket(cfg config.View) *pb.Ticket {
-	characters := cfg.GetStringSlice("testConfig.characters")
-	regions := cfg.GetStringSlice("testConfig.regions")
-	min := cfg.GetFloat64("testConfig.minRating")
-	max := cfg.GetFloat64("testConfig.maxRating")
-	latencyMap := latency(regions)
+// PlaylistTicket generates a ticket for profile scale testing
+func PlaylistTicket() *pb.Ticket {
```
sawagh: Just rename to Ticket() ? since this is the only ticket generation now?

PR #993, opened 2019-11-19
Update gRPC middlewares used in the internal/rpc library

yfei1:
This commit did a bunch of changes to our `internal/rpc` library:
- Removed the gRPC retry middleware - gRPC retry is only valid for unary service and server stream service. Some of the Open Match services are using client side streaming/bidirectional streaming calls and will be auto failed by the middleware if turned on.
- Added `grpc_tracing` middleware to help context propagation. According to the [doc](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/tracing/opentracing/doc.go), for a service that sends out requests and receives requests, you *need* to use both, otherwise downstream requests will not have the appropriate requests propagated.
- Moved the KeepAliveParams from server side to client side - adding server side keepalive logics without proper configuration on the client side will have the server force reset the connection when a connection reached its `MaxConnectionAge`. Moving the logic to the client side could help resolve this issue and migitage the `Unavailable desc = all SubConns are in TransientFailure, latest connection error: <nil>` error

### DIFF
--- a/internal/rpc/clients.go
+++ b/internal/rpc/clients.go
@@ -30,13 +30,15 @@ import (
 	"go.opencensus.io/plugin/ocgrpc"
 
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
-	grpc_retry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
+	grpc_tracing "github.com/grpc-ecosystem/go-grpc-middleware/tracing/opentracing"
+
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"go.opencensus.io/plugin/ochttp"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/balancer/roundrobin"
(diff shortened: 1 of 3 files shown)


internal/rpc/clients.go
```diff
@@ -310,6 +312,11 @@ func newGRPCDialOptions(enableMetrics bool, enableRPCLogging bool, enableRPCPayl
 		grpc.WithStreamInterceptor(grpc_middleware.ChainStreamClient(si...)),
 		grpc.WithUnaryInterceptor(grpc_middleware.ChainUnaryClient(ui...)),
 		grpc.WithBalancerName(roundrobin.Name),
+		grpc.WithKeepaliveParams(keepalive.ClientParameters{
```
Laremere: PR description states "Moved the KeepAliveParams from client side to server side"

This does the opposite.

yfei1: oops that's a typo

PR #994, opened 2019-11-20
Optimized Redis connection pool logic and added more metrics for Redis pool

yfei1:
This commit added two prometheus gauge metrics to track the number of Redis connections per pool when `/healthz` is called. Also added a Grafana chart to display the metrics aggregated by components.

Meanwhile, this commit implemented a simple `TestOnBorrow` handler for Redis so that the Pool could discard a bad connection if one returned an error when returning to the pool.
image


internal/statestore/redis.go
```diff
@@ -114,10 +120,15 @@ func (rb *redisBackend) HealthCheck(ctx context.Context) error {
...
+
+	poolStats := rb.redisPool.Stats()
+	telemetry.SetGauge(ctx, mRedisConnPoolActive, int64(poolStats.ActiveCount))
+	telemetry.SetGauge(ctx, mRedisConnPoolIdle, int64(poolStats.IdleCount))
+
+	_, err = redisConn.Do("PING")
 	// Encountered an issue getting a connection from the pool.
 	if err != nil {
-		return status.Errorf(codes.Unavailable, "%v", err)
+		return status.Errorf(codes.Unavailable, "%w", err)
```
golangcibot: printf: Errorf call has error-wrapping directive %w (from `govet`)

internal/statestore/redis.go
```diff
@@ -80,6 +82,10 @@ func newRedis(cfg config.View) Service {
 		MaxActive:   cfg.GetInt("redis.pool.maxActive"),
 		IdleTimeout: cfg.GetDuration("redis.pool.idleTimeout"),
 		Wait:        true,
+		TestOnBorrow: func(c redis.Conn, _ time.Time) error {
```
sawagh: Is this a pattern recommended by redigo? Where where does this get invoked? In absence of this, would the next request using that redis connection fail?

install/helm/open-match/templates/_helpers.tpl
```diff
@@ -104,8 +104,8 @@ livenessProbe:
     scheme: {{ if (.isHTTPS) }}HTTPS{{ else }}HTTP{{ end }}
     path: /healthz
     port: {{ .port }}
-  initialDelaySeconds: 5
-  periodSeconds: 5
+  initialDelaySeconds: 10
```
sawagh: What behavior does this change? Would the user ever need to tweak these? in what scenario?

internal/statestore/redis.go
yfei1: > Is this a pattern recommended by redigo?

Based on the documentation this is optional. I added this function to make sure the pool is not using bad redis connection when receiving a client call.
```
// TestOnBorrow is an optional application supplied function for checking
// the health of an idle connection before the connection is used again by
// the application. Argument t is the time that the connection was returned
// to the pool. If the function returns an error, then the connection is
// closed.
```
> Where where does this get invoked?

This is invoked when an idle connection was returned to the pool.

> In absence of this, would the next request using that redis connection fail?

Yes, although the pool will automatically remove the unhealthy connection if it hit it, in absence of this, the first call to a bad connection will always fail.

install/helm/open-match/templates/_helpers.tpl
yfei1: It changes the health check frequency to 10 seconds per call. This can somewhat alleviate the spam from `/healthz` when we turn on the rpc logging. Also 10 is the default value for k8s probes check.

internal/statestore/redis.go
sawagh: Does this actually address any issues you are seeing - or you are taking this as a best practice?

yfei1: It resolved some of the Unavailable error I see when an idle connection got reset ( though it’s pretty rare)

PR #999, opened 2019-11-25
Fix Redis Err: Max Clients Reached error

yfei1:
This commit fixed an issue where Open Match may throw out `Err: max clients reached` errors from Redis side under load testing scenarios. At this point, Open Match should be able to scale with 1600 profiles and 5000 tickets in the statestore.

The reason that we got those errors from Redis is by default Redis set its `maxClient` connections limit to 10k. However, Open Match has `maxIdle` number set to 5000 per pod, which exceed Redis's limit and failed the API calls. This commit manually overrides the maxClient number to 100k, reduce the maxIdle number to 200, and configure the file descriptors' limit to 10k by setting `sysctl -w net.core.somaxconn=100000` using the initContainer if enabled.

### DIFF
--- a/install/helm/open-match/templates/om-configmap-override.yaml
+++ b/install/helm/open-match/templates/om-configmap-override.yaml
@@ -33,5 +33,5 @@ data:
     synchronizer:
       enabled: true
       registrationIntervalMs: 3000ms
-      proposalCollectionIntervalMs: 2000ms
+      proposalCollectionIntervalMs: 15000ms
 {{- end }}
--- a/install/helm/open-match/values.yaml
+++ b/install/helm/open-match/values.yaml
@@ -106,11 +106,13 @@ configs:
 redis:
   fullnameOverride: om-redis
   redisPort: 6379
-  usePassword: true
-  usePasswordFile: true
+  usePassword: false
+  usePasswordFile: false
   secretMountPath: /opt/bitnami/redis/secrets
   ignoreLists:
     ttl: 60000ms
+  configmap:
+    maxclients 100000
   master:
     disableCommands: [] # don't disable 'FLUSH-' commands
   metrics:
@@ -132,8 +134,8 @@ redis:
       - -c
       - |-
         install_packages procps
-        sysctl -w net.core.somaxconn=10000
-        echo never > /host-sys/kernel/mm/transparent_hugepage/enabled
+        sysctl -w net.core.somaxconn=100000
+        echo never > /host-sys/kernel/mm/transparent_hugepage/enabled # Disable THP support
 
 ###############################################################################################################################
 #                               Open Match configurations for the subcharts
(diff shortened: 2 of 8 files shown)


sawagh:
Can we add a high level description of the fix? That way it is clear what changed from the PR description for future reviewers?

install/helm/open-match/templates/om-configmap-override.yaml
```diff
@@ -32,6 +32,6 @@ data:
         httpport: "{{ .Values.evaluator.httpPort }}"
     synchronizer:
       enabled: true
-      registrationIntervalMs: 3000ms
-      proposalCollectionIntervalMs: 2000ms
+      registrationIntervalMs: 7500ms
```
sawagh: registrationIntervalMs only covers for how long it takes for all the calls to get into OM. If the director does this in a tight loop - then may be a much shorter registration interval (1s?) might work too?

install/helm/open-match/values.yaml
```diff
@@ -106,11 +106,13 @@ configs:
 redis:
   fullnameOverride: om-redis
   redisPort: 6379
-  usePassword: true
-  usePasswordFile: true
+  usePassword: false
```
sawagh: Reason for this?

install/helm/open-match/values.yaml
```diff
@@ -106,11 +106,13 @@ configs:
...
+  usePasswordFile: false
   secretMountPath: /opt/bitnami/redis/secrets
   ignoreLists:
     ttl: 60000ms
+  configmap:
+    maxclients 100000
```
sawagh: Not in this PR but our best practices / scaling overview doc should list what the tradeoffs here are - for example - based on scenarios, how should these change?

install/helm/open-match/templates/om-configmap-override.yaml
yfei1: Based on the test with 4k profiles I will set it to 3s.

install/helm/open-match/values.yaml
yfei1: This is a minor twist to unify the security settings. It feels a bit off by having the TLS settings for Open Match core disabled and Redis TLS enabled.

Do you mind if I have it checked in along with this fix?

install/helm/open-match/values.yaml
yfei1: Sure, will add this as a note to the issue that tracks the production guide.

test/e2e/game_flow_test.go
```diff
@@ -208,12 +208,12 @@ func validateFetchMatchesResponse(ctx context.Context, t *testing.T, wantTickets
 		if err == io.EOF {
 			break
 		}
-		require.Nil(t, err)
+		assert.Nil(t, err)
```
golangcibot: undeclared name: `assert` (from `typecheck`)

PR #1007, opened 2019-12-06
Have applications read in config from custom input

yfei1:
Per discussion, we will have an evaluator and mmf live under the `internal/testing` or `test/e2e` folder for testing purpose. Therefore, it make sense for these components to reuse the internal library instead of reinventing the wheels.

However, the internal config library requires its consumers read in two different config files from different location which complicated the evaluator and mmf setup.

This commit changed the `app.RunApplication` interface to read configurations from a custom getConfig function, such that the new evaluator and mmf could fake their configs and decouple themselves from the internal config library.

### DIFF
--- a/internal/app/appmain.go
+++ b/internal/app/appmain.go
@@ -30,8 +30,8 @@ var (
 )
 
 // RunApplication creates a server.
-func RunApplication(serverName string, bindService func(*rpc.ServerParams, config.View) error) {
-	cfg, err := config.Read()
+func RunApplication(serverName string, getCfg func() (config.View, error), bindService func(*rpc.ServerParams, config.View) error) {
+	cfg, err := getCfg()
 	if err != nil {
 		logger.WithFields(logrus.Fields{
 			"error": err.Error(),
--- a/cmd/backend/backend.go
+++ b/cmd/backend/backend.go
@@ -18,8 +18,9 @@ package main
 import (
 	"open-match.dev/open-match/internal/app"
 	"open-match.dev/open-match/internal/app/backend"
+	"open-match.dev/open-match/internal/config"
 )
 
 func main() {
-	app.RunApplication("backend", backend.BindService)
+	app.RunApplication("backend", config.Read, backend.BindService)
 }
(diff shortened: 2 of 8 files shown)


Laremere:
What is the complexity of the config that needs to be passed?  Eg, are the log, metric, etc settings going to need to be duplicated?

yfei1:
> What is the complexity of the config that needs to be passed? Eg, are the log, metric, etc settings going to need to be duplicated?

For the evaluator, we need the hostname, grpcport and httpport at mininum.
For the mmf, we need the three fields above and a mmlogic address.

Laremere:
That's not all app.RunApplication uses though:

logging.ConfigureLogging(cfg)
rpc.NewServerParamsFromConfig(cfg,

NewServerParamsFromConfig uses ConfigNameEnableRPCLogging, telemetry.ConfigNameEnableMetrics, etc.

yfei1:
> That's not all app.RunApplication uses though:
> 
> logging.ConfigureLogging(cfg)
> rpc.NewServerParamsFromConfig(cfg,
> 
> NewServerParamsFromConfig uses ConfigNameEnableRPCLogging, telemetry.ConfigNameEnableMetrics, etc.

These configs have their default values set to false. IMO, perhaps we don't necessary need the rpc logging and payload logging for an end-to-end test.

internal/app/appmain.go
```diff
@@ -30,8 +30,8 @@ var (
 )
 
 // RunApplication creates a server.
-func RunApplication(serverName string, bindService func(*rpc.ServerParams, config.View) error) {
-	cfg, err := config.Read()
```
sawagh: I see the issue here  (or even in the existing implementation) of the RunApplication depending on values that may just not be set. 

One approach could be for it to define explicitly, a struct with the values it expects - and the callers to fill in that before calling RunApplication. Atleast from readability perspective we could then have a clear understanding of the values the caller chose to not to populate and document its impact.

Wont block this PR for it but can you create an issue for refactoring this?

yfei1: Done.

PR #1009, opened 2019-12-07
Split up stress test into backend/frontend structure

yfei1:


examples/scale/backend/backend.go
```diff
@@ -74,90 +76,48 @@ func Run() {
...
 	fe := pb.NewFrontendClient(feConn)
 
-	for i := 0; i < 10; i++ {
-		errMap := &sync.Map{}
-
-		doCreate(fe, errMap)
-		doFetch(cfg, be, errMap)
+	go doFetch(cfg, be)
+	go doAssign(be)
+	go doDelete(fe)
```
sawagh: In future, we should consider moving this to Frontend - to be more realistic. For now, i think it is ok to get numbers using this (since these numbers will be without fronted checking any assignments at all)

PR #1011, opened 2019-12-09
Moved original evaluator example to internal package

yfei1:
This commit depends on #1007. It moved the evaluator-golang-simple logic and the evaluator harness to `internal/testing/` and `test/evaluator` package. Also changed the Makefile to build and use the new packages for testing accordingly.


internal/testing/customize/evaluator/evaluator.go
```diff
@@ -0,0 +1,50 @@
+// Copyright 2019 Google LLC
```
sawagh: why are these not just a move from the other location instead of a new file with same content?

Issue #762, opened 2019-08-22
Synchronizer should return when all registered MMFs have submitted proposals


sawagh:
Currently, the synchronizer waits for the entire configured time duration before it triggers evaluation. This is not great - lots of wasted dead time. Synchronizer should be able to have a quick return if all the 'registered' calls have submitted proposals - and we are done with the registration window.

yfei1:
Closed as this no longer applies given we are moving to the new synchronizer.

PR #1024, opened 2019-12-18
Rewrite synchronizer and corresponding backend

Laremere:
This rewrite ensures many properties we desire from the synchronizer, and set up several other changes to come.  I have tried my best to make the resultant code simple.  However the Synchronize and runCycle functions have a large number of concerns, so everything isn't easy to split out.

The PR currently does not have unit tests, and I'm not confident that the e2e tests thoroughly ensure all of the desired behaviors.  I would first like input on the logic in the code.  If things seem settled there, I can start working towards full testing.

Properties (provided no bugs) ensured by this PR (some of these are in the previous implementation, but this is half reminding me what to test):
- If all MMFs complete before the proposal collection period, the cycle proceeds.
- If there are any MMFs still running when proposal collection completes, they are canceled.  (In followup, the matches which it produced before canceling will be accepted.)  This isn't exact, due to network and processing delay, but it should mitigate dead on arrival proposals and stuck MMFs clogging up things.
- Tickets are added to the ignore list on a best effort case.  The whole system continues flowing unless all matches created fail their calls to add tickets to the ignore list.
- Synchronizer not being enable now directly breaks.  We should have some discussion here: If it's disabled, then the tickets are never added to the ignore list.
- Most of the parts are set up for non-blocking streaming of matches in a registration cycle.  Followup PRs will make this true end to end.
- Only one sychronization cycle occurs at one time, and it only starts when the first backend registers.  If a backend misses a window, it starts as soon as the previous cycle has added all of its tickets to the ignore list.  With this, it's actually possible to start the next cycle from the backend while the current cycle is still sending matches back.  Though I don't expect this optimization to be necessary, really.
- If all of the calling backends disconnect and it's past the end of the registration window, the synchronizer cycle is canceled.  This should help ensure that long running requests don't stall the entire system (eg an evaluator which takes forever to respond, and the backends already time out).
- Matches are routed back to their calling fetchMatches call.
- No one straggling backend being slow accepting matches will back up the whole system:  At any point matches are sent to another service, the list of matches waiting to be sent buffers.
- Despite running on different go routines and with a different context, errors from the evaluator or statestore which prevent any matches from being returned are properly sent to all backend callers.

Changes to follow:
- Have the backend start streaming the MMF results to the synchronizer as they arrive, and the synchronizer stream to and from the evaluator in the same way.  This will remove points in the code where systems don't overlap work when it could be.
- Add useful metrics.  I would like to measure at least these values: 
  - Time from first registration to last registration in the registration window.
  - Time spent waiting for registrations - indicates that FetchMatch calls are missing registration windows.
  - Time spent where all MMFs have returned, but the synchronizer's registration period hasn't ended.  This is would indicate to OM users that the could possibly reduce MMF registration window.

Resolves https://github.com/googleforgames/open-match/issues/762


internal/app/synchronizer/evaluator_client.go
```diff
@@ -102,17 +102,17 @@ func newGrpcEvaluator(cfg config.View) (evaluator, error) {
...
 	for _, proposal := range proposals {
 		if err = stream.Send(&pb.EvaluateRequest{Match: proposal}); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("Error sending proposals to evaluator: %w", err)
 		}
 	}
 
 	if err = stream.CloseSend(); err != nil {
-		logger.Errorf("failed to close the send stream: %s", err.Error())
+		logger.Errorf("failed to close the send stream: %w", err.Error())
```
golangcibot: printf: Errorf format %w has arg err.Error() of wrong type string (from `govet`)

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -16,145 +16,541 @@ package synchronizer
...
-// of proposals from Match functions.
+var (
+	logger = logrus.WithFields(logrus.Fields{
+		"app":       "openmatch",
+		"component": "app.synchronizer",
+	})
+)
+
+// Matches flow through channels in the synchronizer.  Channel variable names
+// are unused to be consistant between function calls to help track everything.
```
golangcibot: `consistant` is a misspelling of `consistent` (from `misspell`)
```suggestion

```

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -16,145 +16,541 @@ package synchronizer
...
+}
+
+///////////////////////////////////////
+///////////////////////////////////////
+
+// WithCancelCause returns a copy of parent with a new Done channel. The
+// returned context's Done channel is closed when the returned cancel function
+// is called or when the parent context's Done channel is closed, whichever
+// happens first.  Unlike the conext package's WithCancel, the cancel func takes
+// an error, and will return that error on subsiquent calls to Err().
```
golangcibot: `subsiquent` is a misspelling of `subsequent` (from `misspell`)
```suggestion

```

internal/app/backend/backend_service.go
```diff
@@ -72,52 +73,101 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
+			return
+		}
 
-	telemetry.RecordNUnitMeasurement(stream.Context(), mProposalCollectionPhase, time.Since(startTime).Milliseconds())
-	startTime = time.Now()
+	sendProposals:
+		for _, p := range proposals {
+			select {
+			case <-mmfCtx.Done():
+				logger.Warning("proposals from mmfs recieved too late to be sent to synchronizer.")
```
golangcibot: `recieved` is a misspelling of `received` (from `misspell`)
```suggestion

```

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -16,145 +16,541 @@ package synchronizer
...
 	}
 
-	results, err := s.doEvaluateProposals(stream.Context(), proposals, id)
-	if err != nil {
-		return err
+	go func() {
+		for _, ctx := range callingCtx {
+			<-ctx.Done()
+		}
+		cancel(fmt.Errorf("Canceled because all callers were done."))
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -16,145 +16,541 @@ package synchronizer
...
+
+		if err == nil && parent.Err() == nil {
+			ctx.err = err
+		}
+		cancel()
 	}
+}
 
-	return results, nil
+type CancelErrFunc func(err error)
```
golangcibot: exported type `CancelErrFunc` should have comment or be unexported (from `golint`)

internal/app/backend/backend_service.go
```diff
@@ -72,52 +73,101 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
-		case <-stream.Context().Done():
-			return stream.Context().Err()
-		default:
-			err = stream.Send(&pb.FetchMatchesResponse{Match: result})
-			telemetry.RecordUnitMeasurement(stream.Context(), mMatchesFetched)
+	go func() {
+		var startMmfsOnce sync.Once
+		defer func() {
+			startMmfsOnce.Do(func() {
+				errors <- fmt.Errorf("MMFS were never started.")
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

internal/app/backend/backend_service.go
```diff
@@ -72,52 +73,101 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
+				go startMmfsOnce.Do(startMmfs)
+			}
+
+			if resp.CancelMmfs {
+				cancelMmfs()
+			}
+
+			if resp.Match != nil {
+				err = stream.Send(&pb.FetchMatchesResponse{Match: resp.Match})
+				telemetry.RecordUnitMeasurement(stream.Context(), mMatchesFetched)
```
yfei1: `mmfCtx`

internal/app/backend/synchronizer_client.go
```diff
@@ -43,72 +38,19 @@ func newSynchronizerClient(cfg config.View) *synchronizerClient {
...
-
-	ctx, err = util.AppendSynchronizerContextID(ctx, id)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "%v", err)
-	}
-
-	stream, err := client.(ipb.SynchronizerClient).EvaluateProposals(ctx)
-	for _, proposal := range proposals {
-		if err = stream.Send(&ipb.EvaluateProposalsRequest{Id: id, Match: proposal}); err != nil {
+func (sc *synchronizerClient) synchronize(ctx context.Context) (synchronizerStream, error) {
```
yfei1: It looks a little bit weird to me that we need to define a synchronizerStream interface to use the stream. Is it possible to implement a `func getClient() (ipb.SynchronizerClient, error)` function and use it directly?

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -16,145 +16,541 @@ package synchronizer
...
+	m1c       chan<- mAndM6c
+	m2c       chan<- mAndM6c
+	closed    chan struct{}
+	closeOnce sync.Once
+}
+
+// cutoffSender allows values to be passed on the provided channel until cutoff
+// has been called.  This closed the provided channel.  Calls to send after
+// cutoff work, but values are ignored.
+func newCutoffSender(m2c chan<- mAndM6c) *cutoffSender {
```
yfei1: Ah I see this is an effort to avoid panics from closing a closed channel since we may have multiple go routines writing to the same one.

Does this trick fit our use case at here? It might be able to make the code more straightforward if so.
https://stackoverflow.com/a/39237154

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -16,145 +16,541 @@ package synchronizer
...
+///////////////////////////////////////
+
+type mAndM6c struct {
+	m   *pb.Match
+	m6c chan *pb.Match
+}
+
+// For incoming mathces, remembers the channel the match needs to be returned to
+// if it passes the synchronizer.
+func fanInFanOut(m2c <-chan mAndM6c, m3c chan<- *pb.Match, m5c <-chan *pb.Match) {
```
yfei1: To me this is the hardest part to understand, it might worth to elaborate what this function is doing

internal/app/backend/backend_service.go
Laremere: Nope, this is sending a response from the backend.  mmfCtx is only used for calling the match functions.  It can be canceled without canceling the communication with the synchronizer.

internal/app/backend/synchronizer_client.go
Laremere: We need to have a discussion on whether disabling the synchronizer is still allowed.  I didn't return the stream directly (and in that case there's no reason to have this file really.) because that way we could provide a dummy synchronize service.  It would have to add tickets to the ignore list though.

internal/app/synchronizer/synchronizer_service.go
Laremere: No, because changing the channel's value requires editing the variable on multiple go routines.

internal/app/synchronizer/evaluator_client.go
```diff
@@ -102,17 +102,17 @@ func newGrpcEvaluator(cfg config.View) (evaluator, error) {
...
 	for _, proposal := range proposals {
 		if err = stream.Send(&pb.EvaluateRequest{Match: proposal}); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("Error sending proposals to evaluator: %w", err)
 		}
 	}
 
 	if err = stream.CloseSend(); err != nil {
-		logger.Errorf("failed to close the send stream: %s", err.Error())
+		logger.Errorf("failed to close the send stream: %w", err)
```
golangcibot: printf: Errorf call has error-wrapping directive %w (from `govet`)

internal/api/synchronizer.proto
```diff
@@ -17,87 +17,21 @@ package openmatch.internal;
...
-
-  // Identifier for this request issued during request registration.
-  string id = 2;
-}
-
-message EvaluateProposalsResponse {
-  // Results from evaluating proposals for this request.
-  openmatch.Match match = 1;
+message SynchronizeResponse {
+  bool start_mmfs = 1;
```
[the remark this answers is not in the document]
Laremere: Maybe, but this is internal so I didn't care about understand-ability as much as the code is simpler this way....

internal/app/backend/synchronizer_client.go
Laremere: Per discussion, removed the config option altogether.

internal/api/synchronizer.proto
```diff
@@ -17,87 +17,21 @@ package openmatch.internal;
...
+  bool cancel_mmfs = 2;
+  openmatch.Match match = 3;
 }
```
sawagh: Can we make this even simpler. Consider SynchronizeResponse only returns a match stream. 
- The Synchrnoize() call can block till you are ready to run MMFs - so once synchronize returns, the backend can trigger the MMF
- Once the proposal collection interval is over and evaluation done, we close the stream - and the backend automatically cancels any pending mmf. 

That way, the interaction between backend and synchronizer just stays an input / output stream and no need for explicit start / cancel actions overloaded on the stream response?

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -16,145 +16,544 @@ package synchronizer
...
+	for m5 := range m5c {
+		m6cMap[m5.GetMatchId()] <- m5
+	}
+}
+
+///////////////////////////////////////
+///////////////////////////////////////
+
+type cutoffSender struct {
+	m1c       chan<- mAndM6c
```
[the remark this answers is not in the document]
Laremere: I found it very useful when trying to find names for various variables to use this consistent naming.  Coming up with unique names for each variable which makes sense, and then tracking those through the file wasn't working well.

There's no need to generalize this code when it's only used in one place.

internal/app/backend/backend_service.go
```diff
@@ -72,52 +73,101 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
 	if err != nil {
 		return err
 	}
 
-	telemetry.RecordNUnitMeasurement(stream.Context(), mRegisterPhase, time.Since(startTime).Milliseconds())
-	startTime = time.Now()
+	errors := make(chan error, 2)
+	mmfCtx, cancelMmfs := context.WithCancel(stream.Context())
+	startMmfs := func() {
+		resultChan := make(chan mmfResult, len(req.GetProfiles()))
```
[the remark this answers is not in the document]
Laremere: This logic didn't change in this PR.

internal/app/backend/backend_service.go
```diff
@@ -72,52 +73,101 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
 
-	err = doFetchMatchesReceiveMmfResult(stream.Context(), s.mmfClients, req, resultChan)
-	if err != nil {
-		return err
-	}
+		err := doFetchMatchesReceiveMmfResult(mmfCtx, s.mmfClients, req, resultChan)
+		if err != nil {
+			// TODO: Log but continue case where mmfs were canceled once fully
```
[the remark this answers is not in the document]
Laremere: This is a reminder to myself that once this is fully streaming (as in, not blocking on the MMFs before sending any matches to the syncrhonizer), that I need to log in the case where the context was canceled, but continue with the overall call instead of just erroring out.

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -16,145 +16,544 @@ package synchronizer
...
+		"component": "app.synchronizer",
+	})
+)
+
+// Matches flow through channels in the synchronizer.  Channel variable names
+// are unused to be consistent between function calls to help track everything.
+
+// Streams from multiple GRPC calls of matches are combined on a single channel.
+// These matches are sent to the evaluator, then the tickets are added to the
+// ignore list.  Finally the matches are returned to the calling stream.
```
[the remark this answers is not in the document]
Laremere: Well, that's why I said read it from bottom to top ;)

I've intentionally avoided over commenting this code.  I think I can add a little more comments, but any detailed comments will just rot.  The PR description gives a current state of the world for what this code is doing.

internal/app/backend/backend_service.go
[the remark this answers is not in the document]
sawagh: Yup. I just wanted to make a note for a discussion which we had offline - I ll create an issue for this where we can continue the discussion.

internal/app/synchronizer/synchronizer_service.go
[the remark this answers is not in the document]
sawagh: I do not think setting expectation that a file should be read bottom to top helps from readability point of view. Files will grow and change - and future contributors may not stick to the same assumption? 

The PR description is great - but not a substitute for in context comments. 

Given the complex nature of this component, I would err on the side of generously commenting than be concerned about future comment rot. I have added some requests for comments but i definitely think more comments would have saved a lot of review time.

internal/app/synchronizer/synchronizer_service.go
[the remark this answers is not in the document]
sawagh: I do see how this naming scheme was easier as you wrote the code and i agree this wont be generalized and used elsewhere. I mean - when reviewing / revisiting this in future - the following naming of the same code would be easier to grasp (as I do not really need to know what m1c is to know what cutoffSender implements):

type matchAndChannel struct {
	m   *pb.Match
	c chan *pb.Match
}

type cutoffSender struct {
	in       chan<- matchAndChannel
	out       chan<- matchAndChannel
	closed    chan struct{}
	closeOnce sync.Once
}

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -16,145 +16,544 @@ package synchronizer
...
-	var proposals = []*pb.Match{}
+	cfg       config.View
+	store     statestore.Service
+	evaluator evaluator
+
+	synchronizeRegistration chan *registrationRequest
+
+	// startCycle is a buffered channel for containing a single value.  The value
+	// is present only when a cycle is not running.
+	startCycle chan struct{}
```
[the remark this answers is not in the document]
sawagh: I do think that can be achieved with sync primitives too (the previous code was doing that). Anyways, dont have a strong preference - was just not sure if there was a recommended way - and found a detailed back and forth on an open issue here (https://github.com/golang/go/issues/21165).

internal/api/synchronizer.proto
sawagh: I guess then for the register, this is the best option (short of separating out into a separate blocking register call). I will discuss the cancel condition offline on whether it can be handled without explicitly sending any information to the backend.

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -16,145 +16,544 @@ package synchronizer
...
+			}).Error("error streaming in synchronizer to backend: context is done")
+			return stream.Context().Err()
+		case <-registration.cycleCtx.Done():
+			return registration.cycleCtx.Err()
+		}
+	}
+
+}
+
+///////////////////////////////////////
```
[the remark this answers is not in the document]
Laremere: That good?

internal/app/synchronizer/synchronizer_service.go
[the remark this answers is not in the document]
Laremere: There are other cases, eg faninfanout, where it's more useful to know how things are being transferred.  I think it's reasonable to be consistent with naming.

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -16,145 +16,562 @@ package synchronizer
...
+	go s.wrapEvaluator(ctx, cancel, bufferChannel(m3c), m4c)
+	go func() {
+		s.addMatchesToIgnoreList(ctx, cancel, bufferChannel(m4c), m5c)
+		// Wait for ignore list, but not all matches returned, the next cycle
+		// can start now.
+		close(closedOnCycleEnd)
+	}()
 
-		proposals = append(proposals, req.GetMatch())
+	/////////////////////////////////////// Run Registartion Period
```
golangcibot: `Registartion` is a misspelling of `Registration` (from `misspell`)
```suggestion

```

internal/app/synchronizer/synchronizer_service.go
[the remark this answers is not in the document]
Laremere: I added some more comments to registration, runCycle, and Synchronize.

PR #1028, opened 2019-12-26
Move ignorelists.ttl from Redis section to Open Match core

### LINKED ISSUE #1025: Ignore list ttl shouldn't be under redis.
It's currently `redis.ignoreList.ttl`

This doesn't make much sense, it much more a setting for OM than it is for redis.

yfei1:
Resolves #1025


Laremere:
Looks good, but the tests are unhappy?

PR #1030, opened 2020-01-07
Implement basic tunable parameters logic for benchmarking scenarios

yfei1:
This commit implements the knobs to control `ShouldCreateTicketForever`, `ShouldAssignTicket`, `ShouldDeleteTicket`, `TicketCreatedQPS`, and `CreateTicketNumber`. Also removed the roster-based-mmf from the repo since it is only used for the scale test and there is no need to build its image in every CI run.

After this commit got checked in, users are able to configure the knobs via the new benchmarking framework and run `make install-scale-chart` to it. 

TODO: 
1. Implement the filter number and profile number logic. This requires a rewrite for `examples/scale/tickets` and `examples/scale/profiles` package.


examples/scale/backend/backend.go
```diff
@@ -77,8 +80,14 @@ func Run() {
 	fe := pb.NewFrontendClient(feConn)
 
 	go doFetch(cfg, be)
-	go doAssign(be)
-	go doDelete(fe)
+
+	if activeScenario.ShouldAssignTicket {
```
Laremere: Naming: Should this be "BackendAssignsTickets", and "BackendDeletesTickets"?

There are scenarios where the Frontend deletes tickets, so making the distinction would be useful.

examples/scale/backend/backend.go
```diff
@@ -77,8 +80,14 @@ func Run() {
...
+		go doAssign(be)
```
Laremere: These drain the matches and deleteIds channels.  Without running them the channels will get clogged.

examples/scale/frontend/frontend.go
```diff
@@ -67,31 +70,58 @@ func Run() {
 }
 
 func doCreate(cfg config.View, fe pb.FrontendClient) {
```
Laremere: I'm not a fan of "do" being in function names.  It's like appending "Value" to the end of variable names.

examples/scale/frontend/frontend.go
```diff
@@ -67,31 +70,58 @@ func Run() {
...
-	concurrent := cfg.GetInt("testConfig.concurrentCreates")
-	start := time.Now()
+	cycleCreate := func(wg *sync.WaitGroup, ticketPerRoutine int, start time.Time) {
```
Laremere: Why not just have this be a top level function?

examples/scale/frontend/frontend.go
```diff
@@ -67,31 +70,58 @@ func Run() {
...
+				errMap.Store(errMsg, errRead.(int)+1)
+			}
+			cycleCreated += 1
+		}
+
+		atomic.AddUint32(&totalCreated, cycleCreated)
+	}
+
 	for {
+		if !activeScenario.ShouldCreateTicketForever && int(totalCreated) >= activeScenario.CreateTicketNumber {
```
Laremere: CreateTicketNumber is only used when ShouldCreateTicketForever is false.  This means the domain of possible values for these two variables does not map cleanly to the state space of options.

Also, CreateTicketNumber doesn't have immediate meaning to me.  Perhaps "TotalTicketsToCreate".

Two options I see:
- If TotalTicketsToCreate is zero, that signifies to create infinite tickets.  This reduces it down to a single simple type, but it's not a natural mapping (zero is a lot less than infinity)
- A function such as "ShouldContinueCreatingTickets(totalCreated)" might work.  Though it makes the control surface less obvious, now I need to implement a `func(int) bool` in my scenario, though it could have useful implementations such as "createInfiniteTickets" or "createLimitedTickets"?

Though want is the point of having this variable?  What is the scenario for when it is set?

examples/scale/frontend/frontend.go
```diff
@@ -67,31 +70,58 @@ func Run() {
...
+			break
+		}
+
+		// Each inner loop creates TicketCreatedQPS tickets
+		start := time.Now()
+		ticketPerRoutine := int(activeScenario.TicketCreatedQPS) / numOfRoutineCreate
```
Laremere: This might create more tickets than the CreateTicketNumber.

examples/scale/backend/backend.go
```diff
@@ -76,146 +69,110 @@ func Run() {
...
-			errMap.Store(errMsg, errRead.(int)+1)
-		}
+		statProcessor.IncrementStat("Assigned", len(ids))
 
-		atomic.AddUint64(&assigned, uint64(len(ids)))
-		for _, id := range ids {
-			deleteIds <- id
+		// Delete Tickets
+	Delete:
+		if !activeScenario.BackendDeletesTickets {
```
Laremere: invert the if, and get rid of the gotos.

Issue #991, opened 2019-11-19
Investigate: Reduce to only using one grcp port and one http port


Laremere:
Given that multiple grpc services can be hosted on the same port, and same with http, it seems like we could get away with Open Match only using two ports across all services.  http(s) should use the standard ports, and grpc should use whatever the standard is there.  They might even be able to use the same port.

This would reduce the complexity across the stack: Simply use the same port numbers everywhere.  Different grpc services would translate to different k8s services anyways, so that would all work.  In a mini-match case, all the services should be on the same grpc server anyways.

yfei1:
Some thoughts about this change. I definitely think using one port for all services can greatly reduce the complexity of our configurations, but our current debugging experience actually greatly benefits from using different ports for each service. For example, if we see some errors like `failed to handshake with xxxx:50503` from our director, we will then realize the call failed to fetch data from the mmlogic service. It would be a headache to debug the connection if we directly unify the ports without any improvements on logging.

Laremere:
Good point.  We should fix the errors to have more details.

sawagh:
@Laremere - looks like the attempt to fix this was closed. Can you summarize the challenge you hit and whether we will still address this issue?

Laremere:
I closed it because the tests were failing and I didn't want to spend the time debugging them instead of making progress on scale tests.  I thought there was an easier workaround which is why I closed it (it wasn't easier, but the second one I found was).

We should still do this, it's just a matter of taking the time to do it right.

PR #1035, opened 2020-01-11
Switch Open Match services to use consistent ports.

Laremere:
READY FOR REVIEW BUT NEED TO RUN A FULL CLUSTER BEFORE MERGE TO VERIFY PROXY COMMANDS ALL WORK.

Resolves: https://github.com/googleforgames/open-match/issues/991

This simplifies service start up as it means services don't need to be defined in the config in order to know what ports they should open.  I would like to look into simplifying/removing the configuration around hostnames at some point in the future.  It also generally makes things easier from a standpoint of not having to choose ports to allocate.

While I need to test all of the proxy commands anyways, simplifies them.  (except locust, which we might be able to remove?)

### DIFF
--- a/install/helm/open-match/subcharts/open-match-customize/templates/customize-configmap.yaml
+++ b/install/helm/open-match/subcharts/open-match-customize/templates/customize-configmap.yaml
@@ -24,18 +24,18 @@ metadata:
     release: {{ .Release.Name }}
 data:
   matchmaker_config_default.yaml: |-
+    ports:
+      grpcPort: "{{ .Values.global.kubernetes.service.grpcPort }}"
+      httpPort: "{{ .Values.global.kubernetes.service.httpPort }}"
+
     api:
       functions:
         hostname: "{{ .Values.function.hostName }}"
-        grpcport: "{{ .Values.function.grpcPort }}"
-        httpport: "{{ .Values.function.httpPort }}"
       
       evaluator:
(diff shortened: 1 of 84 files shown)


install/helm/open-match/subcharts/open-match-customize/templates/customize-configmap.yaml
```diff
@@ -24,18 +24,18 @@ metadata:
...
       
       evaluator:
         hostname: "{{ .Values.evaluator.hostName }}"
-        grpcport: "{{ .Values.evaluator.grpcPort }}"
-        httpport: "{{ .Values.evaluator.httpPort }}"    
   matchmaker_config_override.yaml: |-
     api:
       mmlogic:
         hostname: "{{ .Values.mmlogic.hostName }}.{{ .Release.Namespace }}.svc.cluster.local"
-        grpcport: "{{ .Values.mmlogic.grpcPort }}"
```
sawagh: Update just the values to the same than updating to a common variable? (keep .Values.mmlogic.grpcPort - and just set that to the desired value)? That would be inline with the rest of the changes in this file?

Laremere: one of the values from this PR is that new binaries don't need to specify all these config values if they're not hosting services.  EG, the scale-frontend doesn't have a service.  However to export metrics it needs to host the http server.

PR #1038, opened 2020-01-14
Enable and establish Redis connections via Sentinel

yfei1:
This commit:

1. Enabled Redis sentinel for both values.yaml and values-production.yaml installation, which by default have sentinel installation turned on when installing Open Match in a Kubernetes cluster.
2. Rewrote the master connection logic. Use Redis sentinel for master discovery to achieve high-availability and failover.
3. Introduced a new `minisentinel` dependency for testing purpose.


sawagh:
Lets get other stuff in, resolve conflicts and I ll PR the resolved file tomorrow.

internal/statestore/redis.go
```diff
@@ -56,65 +56,126 @@ func (rb *redisBackend) Close() error {
...
 // newRedis creates a statestore.Service backed by Redis database.
 func newRedis(cfg config.View) Service {
-	// As per https://www.iana.org/assignments/uri-schemes/prov/redis
-	// redis://user:secret@localhost:6379/0?foo=bar&qux=baz
-
-	// Add redis user and password to connection url if they exist
-	redisURL := "redis://"
-	maskedURL := redisURL
+	return &redisBackend{
+		healthCheckPool: getRedisPool("healthcheck", cfg),
```
sawagh: the getRedisPool with a type "healthcheck" and "master" seems like a hack. furthermore, in their imlementation, you also skip to default case if not healthckeck. Are there other types of pools you envision using? If not, simply having 2 helpers getHealthCheckPool and getMasterPool might do? May be have a getSentinelPool for the common helper they both use? outside of that, most logic for these two seem separated into cases anyways.

internal/statestore/redis.go
```diff
@@ -56,65 +56,126 @@ func (rb *redisBackend) Close() error {
...
-			return err
-		},
-		DialContext: func(ctx context.Context) (redis.Conn, error) {
-			if ctx.Err() != nil {
-				return nil, ctx.Err()
+func getDialFunc(poolType string, cfg config.View, sentinelPool *redis.Pool) func(ctx context.Context) (redis.Conn, error) {
+	idleTimeout := cfg.GetDuration("redis.pool.idleTimeout")
+	healthCheckTimeout := cfg.GetDuration("redis.pool.healthCheckTimeout")
+
+	if cfg.GetBool("redis.sentinelEnabled") {
```
sawagh: if sentinelPool == nil? rather than looking up enabled? else for any reason if it is nil, you panic accessing it further below?

internal/statestore/redis.go
```diff
@@ -56,65 +56,126 @@ func (rb *redisBackend) Close() error {
...
-
-	return &redisBackend{
-		healthCheckPool: healthCheckPool,
-		redisPool:       pool,
-		cfg:             cfg,
+		default:
+			redisLogger.Fatalf("dial type: %s not supported, sentinelEnabled: %v", poolType, cfg.GetBool("redis.sentinelEnabled"))
+			return nil
+		}
+	} else {
```
sawagh: The more i read this, the more i think it might be readable to split out healthcheck / master pool fetch logic up top - and each can then have if sentinel enabled blocks within.

internal/statestore/redis.go
```diff
@@ -139,6 +200,36 @@ func (rb *redisBackend) HealthCheck(ctx context.Context) error {
...
+	// Assume the connection is valid if it was used in 30 sec.
+	if time.Since(lastUsed) < 15*time.Second {
+		return nil
+	}
+
+	_, err := c.Do("PING")
+	return err
+}
+
+func redisURLFromAddr(addr string, cfg config.View) string {
```
sawagh: nice to have - similar one line helper to get sentinel url - wont be long - but good to abstract into helper.

internal/statestore/redis.go
```diff
@@ -56,66 +56,116 @@ func (rb *redisBackend) Close() error {
...
-		}
-		redisURL += fmt.Sprintf("%s:%s@", cfg.GetString("redis.user"), string(passwordData))
-		maskedURL += fmt.Sprintf("%s:%s@", cfg.GetString("redis.user"), "**********")
+func getHealthCheckPool(cfg config.View) *redis.Pool {
+	var healthCheckURL string
+	var maxIdle = 3
+	var maxActive = 0
+	var healthCheckTimeout = cfg.GetDuration("redis.pool.healthCheckTimeout")
+
+	if cfg.IsSet("redis.sentinelHostname") {
```
sawagh: why decide based on this vs enabled or not? Also, if enabled and these fields not provided, do we have a config validation anywhere? Either we should - or there should some helper here which checks validates and falls back to master directly if any sentinel config is missing?

sawagh:
Where is the sentinel deployment happening? Dont see that in this PR. I mean, I see we setting the configuration here - but shouldnt our default deployment steps also deploy the sentinel now?

internal/statestore/redis.go
```diff
@@ -56,66 +56,116 @@ func (rb *redisBackend) Close() error {
...
+			if ctx.Err() != nil {
+				return nil, ctx.Err()
 			}
 
-			_, err := c.Do("PING")
-			return err
-		},
-		DialContext: func(ctx context.Context) (redis.Conn, error) {
+			sentinelConn, err := sentinelPool.GetContext(ctx)
+			if err != nil {
```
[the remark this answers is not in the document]
yfei1: Redis sentinel is a sidecar container for each of the redis pod that we installed. The sentinel installation will also install a Kubernetes service that match to these redis pods, so as long as we have at least one redis instance working, the sentinel should works fine.

install/helm/open-match/templates/om-configmap-default.yaml
```diff
@@ -87,10 +87,18 @@ data:
 
     redis:
 {{- if index .Values "open-match-core" "redis" "enabled" }}
+{{- if index .Values "redis" "sentinel" "enabled"}}
```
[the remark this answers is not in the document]
yfei1: If sentinel is enabled, we only expect the sentinel information to be set, all of the information about the current master needs to be fetched from the sentinel instead - the helm chart that we are using will expose a service that map to all of the redis pods (master + slaves). If not, the helm chart that we are using right now will automatically deploy a redis-master service (without the slaves) and expose the redis-master hostname within the cluster.

internal/statestore/redis.go
yfei1: There is not much difference between using this vs enabled or not. 

If the sentinel config is missing, then it means based on the helm config, the redis sentinel is not enabled - which means redis is configured to talk to the master directly. 

When the sentinel config exists, the sentinel service has HA guarantees from both the master and slave deployments so there is no need to fall back to master directly.

PR #1040, opened 2020-01-14
Remove zipkin references from Open Match

yfei1:
Per discussion, remove Zipkin code and documentation references from Open Match since its functionality is overlapped with Jaeger and its installation steps are hard to automate using helm.


go.sum
```diff
@@ -19,8 +19,6 @@ contrib.go.opencensus.io/exporter/prometheus v0.1.0 h1:SByaIoWwNgMdPSgl5sMqM2KDE
 contrib.go.opencensus.io/exporter/prometheus v0.1.0/go.mod h1:cGFniUXGZlKRjzOyuZJ6mgB+PgBcCIa79kEKR8YCW+A=
 contrib.go.opencensus.io/exporter/stackdriver v0.12.8 h1:iXI5hr7pUwMx0IwMphpKz5Q3If/G5JiWFVZ5MPPxP9E=
 contrib.go.opencensus.io/exporter/stackdriver v0.12.8/go.mod h1:XyyafDnFOsqoxHJgTFycKZMrRUrPThLh2iYTJF6uoO0=
-contrib.go.opencensus.io/exporter/zipkin v0.1.1 h1:PR+1zWqY8ceXs1qDQQIlgXe+sdiwCf0n32bH4+Epk8g=
```
Laremere: Did you manually remove these?  It's my understanding that lines shouldn't get removed from go.sum?

yfei1: I did a ‘make update-does’ to regenerate the go.sum file. Should I leave them untouched?

Laremere: ah, ok, that looks fine then.

PR #1042, opened 2020-01-15
Add metric support to the scale tests

Laremere:
This makes scale-frontend and scale-backend emit metrics which we can use to judge how well the scale it working.

I haven't touched the start of the other metric option for the scale test.  Can clean it up later.

Here's a snapshot of the updated dashboard: https://snapshot.raintank.io/dashboard/snapshot/Xm5S6nVGwHA4iRsnv3hcVTdUYlB1mjId?orgId=2


examples/scale/backend/backend.go
```diff
@@ -40,19 +40,27 @@ var (
 
 	activeScenario = scenarios.ActiveScenario
 	statProcessor  = scenarios.NewStatProcessor()
+
+	mIterations          = telemetry.Counter("scale_backend_iterations", "fetch match iterations")
```
yfei1: we can actually use prometheus query to calculate these rates without adding these metrics. But for the sake of fast iterations I'll approve this PR. Can you add a TODO here so that we could clean it up later?

e.g. If you want to query FetchMatches error rate
`rate(grpc_io_server_completed_rpcs{grpc_server_method="openmatch.Backend/FetchMatches", grpc_server_status!~"OK"}[5m])`

install/helm/open-match/subcharts/open-match-telemetry/dashboards/scale-test.json
```diff
@@ -0,0 +1,840 @@
+{
```
yfei1: I don't mind adding a grafana chart to track the emitted scale metrics but is it possible to place this file elsewhere in the helm chart - perhaps `subcharts/open-match-scale` is a better place for this schema. Otherwise users who don't wanna run the scale test will always see this empty graph when they follow the Grafana installation instructions

Laremere: To do that we need to add something which specifically selects which directories to look for dashboards in.  I'm up for that being done, but I think we should do it as a cleanup so we can make progress here.  This is a general problem, as other future extensions are going to want to add their own dashboards.

Laremere: https://github.com/googleforgames/open-match/issues/1043

yfei1: Sure. I think this is a v0.9 issue - having users seeing unexpected grafana charts is a regression comparing to what offered in v0.8. I'll look into this issue later this week.

examples/scale/backend/backend.go
Laremere: I specifically wanted to only use metrics from the point of view from the caller.  This way, eg, errors for calls which never even make it to the service are accounted for.

I just seemed easier for me conceptually to add these.

PR #1050, opened 2020-01-16
Add AUTH tests to Redis implementation

yfei1:
This commit added basic AUTH tests for Redis connections w/ and w/o sentinel. 

Depends on #1038.

The minisentinel library that we used for testing does not support AUTH from sentinel yet, so I manually set the value of `redis.sentinelUsePassword` to false when doing the end-to-end test. Added a TODO to change the test and enable the sentinel AUTH when minisentinel starts to supports it.


sawagh:
Please merge and resolve so the PR is unblocked.

internal/statestore/redis.go
```diff
@@ -71,10 +71,10 @@ func getHealthCheckPool(cfg config.View) *redis.Pool {
 
 	if cfg.IsSet("redis.sentinelHostname") {
 		sentinelAddr := getSentinelAddr(cfg)
-		healthCheckURL = redisURLFromAddr(sentinelAddr, cfg)
+		healthCheckURL = redisURLFromAddr(sentinelAddr, cfg, false)
```
sawagh: Why no password for the healthcheckurl? can you add a note about this to the PR description. The PR description just seems to state this being a test only change but looks like there is more.

yfei1: The redis sentinel library available in golang does not support authentication for now, so I only added the test case for sentinel w/o auth.

yfei1: working on an update

PR #1051, opened 2020-01-16
Add gRPC CSharp bindings

### LINKED ISSUE #1047: Generated C# files do not include GRPC bindings
**What happened**: Generating the C# files - or using the ones included in the repo - does not include the GRPC service bindings. In C#, the `--grpc_out` parameter is required to build these; it is missing in the `Makefile`.

**What you expected to happen**: I'd expect the GRPC bindings to be built.

I attempted to PR this myself by adding the necessary line. However, the `grpc_csharp_plugin` doesn't seem to be included in the toolchain, and so this makes the build fail. My understanding of how the toolchain is built is limited, but it does seem like `dotnet` is supposed to be installed - which could be used to fetch the necessary NuGet package containing the plugin binary - but right now it isn't.

yfei1:
Resolved #1047. 

This commit:
1. Added the `grpc_out` flag to the corresponding Makefile target.
2. Fixed the Makefile dependencies of the `build-csharp` target.
3. Added the required Grpc dependencies to the `.csproj` file


Makefile
```diff
@@ -644,19 +644,22 @@ csharp/OpenMatch/Annotations.cs: third_party/
...
+		--csharp_out=$(REPOSITORY_ROOT)/csharp/OpenMatch \
+		--grpc_out=$(REPOSITORY_ROOT)/csharp/OpenMatch/
 
 csharp/OpenMatch/Openapiv2.cs: third_party/
 	$(PROTOC) third_party/protoc-gen-swagger/options/openapiv2.proto \
 		-I $(REPOSITORY_ROOT) -I $(PROTOC_INCLUDES) \
 		--plugin=protoc-gen-grpc=grpc_csharp_plugin \
-		--csharp_out=$(REPOSITORY_ROOT)/csharp/OpenMatch/
+		--csharp_out=$(REPOSITORY_ROOT)/csharp/OpenMatch/ \
+		--grpc_out=$(REPOSITORY_ROOT)/csharp/OpenMatch/
```
JoelOtter: I don't _think_ GRPC bindings are necessary for these first two.

csharp/OpenMatch/OpenMatch.csproj
```diff
@@ -11,6 +11,8 @@
 
   <ItemGroup>
     <PackageReference Include="Google.Api.CommonProtos" Version="1.7.0" />
+    <PackageReference Include="Google.Protobuf" Version="3.8.0" />
+    <PackageReference Include="Grpc" Version="2.26.0" />
```
JoelOtter: When I did this locally I only needed the `Grpc` package for the build to pass; ymmv. :)

Makefile
yfei1: Good catch, thanks.

csharp/OpenMatch/OpenMatch.csproj
yfei1: I see. I'm following the guide in this [link](https://docs.microsoft.com/en-us/aspnet/core/grpc/basics?view=aspnetcore-3.1#c-tooling-support-for-proto-files).

PR #1054, opened 2020-01-17
Add a first match scenario to the scale tests

Laremere:
This first match scenario runs one pool with all tickets, pairing tickets into 1v1 matches with no logic.

Metrics example: https://snapshot.raintank.io/dashboard/snapshot/JZQvjGLgZlezuZfNxPAh8n098JQuCyPW


install/helm/open-match/templates/om-configmap-override.yaml
```diff
@@ -31,6 +31,6 @@ data:
         grpcport: "{{ .Values.evaluator.grpcPort }}"
         httpport: "{{ .Values.evaluator.httpPort }}"
     synchronizer:
-      registrationIntervalMs: 3000ms
-      proposalCollectionIntervalMs: 15000ms
+      registrationIntervalMs: 250ms
```
yfei1: can you make the following change to https://github.com/googleforgames/open-match/blob/master/Makefile#L376 or give the CI environment a larger registration interval to make our CI happy? The ignoreListTTL should be in sync with the one in https://github.com/googleforgames/open-match/blob/master/internal/statestore/testing/consts.go#L29. Our previous registrationInterval didn't catch this discrepancy but the new one does.

```diff
- --set open-match-core.ignoreListTTL=1000ms \
+ --set open-match-core.ignoreListTTL=500ms \
```

examples/scale/scenarios/first_match_scenario.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+)
+
+func firstMatchProfile() []*pb.MatchProfile {
+	return []*pb.MatchProfile{
+		{
+			Name: "entirePool",
+			Pools: []*pb.Pool{
+				{
+					Name: "all",
```
yfei1: can you use a variable for `all` instead? you also refer to this literal at line 41 and this correlation is not obvious at first glance.

PR #1058, opened 2020-01-17
Run scale test assignment and deletes in parallel

Laremere:
Start 30 go routines for each at the beginning of the test, and pass them from fetch matches with a buffer.

Gets the redis state store first match to handle >500 tickets per second: 
https://snapshot.raintank.io/dashboard/snapshot/yO88xrIUe1bFR29iNZt4YuM0xuBb8PX9

### DIFF
--- a/examples/scale/backend/backend.go
+++ b/examples/scale/backend/backend.go
@@ -81,7 +81,17 @@ func run(cfg config.View) {
 	w := logger.Writer()
 	defer w.Close()
 
-	for {
+	matchesForAssignment := make(chan *pb.Match, 1000)
+	ticketsForDeletion := make(chan string, 1000)
+
+	for i := 0; i < 30; i++ {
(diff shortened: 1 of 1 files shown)


examples/scale/backend/backend.go
```diff
@@ -81,7 +81,17 @@ func run(cfg config.View) {
 	w := logger.Writer()
 	defer w.Close()
 
-	for {
+	matchesForAssignment := make(chan *pb.Match, 1000)
+	ticketsForDeletion := make(chan string, 1000)
```
yfei1: can we write it in a way without these two channels? it is possible to have these two channels become the bottleneck of the scale test in some scenarios.

Laremere: Bumped up to meet our maximum values for what we'll expect out of the scale tests.  I do want the director to put the breaks on if it can't keep up bough.

Issue #1037, opened 2020-01-13
Allow directors to return matched tickets to the match-making pool for transient failures


rib:
**Is your feature request related to a problem? Please describe.**

In my director it's possible for there to be an error allocating a server that may simply be a race condition as a result of clients trying to join a named private group before someone else has actually created that group.

In general errors are possible in the director after retreiving matches and it seems important that the director should be able to determine what happens to any tickets caught up in this condition.

There doesn't seem to be a way for the director to gracefully put those tickets back into the pool for matching and instead they will (as far as I understand it) remain in limbo for some time before being re-considered for matching.

Another example where I encountered this was while reviewing this example director:
https://github.com/Laremere/space-agon/blob/master/director/director.go
which currently has this commented code:
```go
// TODO: This drops matches, instead of properly allocating them.  Tickets will only return to
// the general pool after (iirc) one minute.  We should either tell OM that an assignment isn't
// coming, or retry for a little while.
if gsa.Status.State != allocationv1.GameServerAllocationAllocated {
	log.Printf("failed to allocate game server.\n")
	continue
}
```

Related to this, there also doesn't seem to be a clear way for Directors to propagate errors back to a clients if the director ultimately determines that a ticket in unallocatable.

My director is handling allocating servers for private groups where the ticket contains a unique group name. Allocation can fail when interacting with Agones to allocate a server because someone else has created a group with the same name. In this case it's not a race condition and I wouldn't want to simply put the ticket back into the pool for matching - rather I want to trigger an error state for the ticket that will be propagated back via the frontend to the client letting them know why their ticket will never get an assignment.

Maybe for this case the idea is that the ticket should be given a dummy 'assignment' which can contain custom metadata to communicate error conditions?

**Describe the solution you'd like**
For the first issue then maybe the Backend could have an `UnmatchTickets()` or anything along those lines that can be given ticket IDs that were previously got through `FetchMatches` to request that those tickets be re-considered for matching as soon as possible.

For the second issue if the recommendation is to use a dummy Assignment in this case - maybe it could be good if the documentation covered this. It would be good to know if anyone can see a problem with communicating errors back via Assignments.

**Describe alternatives you've considered**
For now I just leave the tickets in limbo for join failures and I think they may eventually be re-considered for matching but not quickly.

For the group creation failures I'm currently also just leaving the tickets in limbo. I had been considering deleting the ticket before realising that a fake assignment would probably be better for communicating error state back to clients.

calebatwd:
[...]

PR #1059, opened 2020-01-18
feature: release tickets api

calebatwd:
A proposal pull request for adding a "releaseTickets" api to the backend service. This api is to answer the issue propose here (https://github.com/googleforgames/open-match/issues/1037) .

It essentially just issues the `DeleteTicketsFromIgnoreList` for the requested ids. This allows a director which is no longer able to handle an assignment for a match to gracefully put the tickets back into the query pool.


calebatwd:
@googlebot I fixed it.

calebatwd:
So strange. I definitely have several CLA's on file for our collaborations?

calebatwd:
@Laremere, @yfei1. Not sure what I've done wrong but my backend_service_test failures look like I didn't set something up correctly. Will probably need some help getting it sorted :)

internal/app/backend/backend_service_test.go
```diff
@@ -168,6 +168,132 @@ func TestDoFetchMatchesFilterChannel(t *testing.T) {
...
+			},
+			req: &pb.ReleaseTicketsRequest{
+				TicketIds: []string{"1"},
+			},
+			wantCode: codes.Unavailable,
+		},
+		{
+			description: "expect ok code when submitted list is empty",
+			preAction: func(_ context.Context, cancel context.CancelFunc, _ statestore.Service, pool *pb.Pool) {
+				cancel()
```
yfei1: My gut tells me this test is failing because the context is cancelled before it is used ;)

calebatwd: Ah! So silly. I'm a bit new to reading some of these lambda-setup test patterns. Thanks! I'll fix it up

calebatwd:
I looked through some of the tutorials to try and figure out where this would make sense to add some examples, but they're all pretty optimistic. I would need to break apart the assign(matches) into something more akin to 
createGame(match)
if err -> release(match) continue
assign(match)

and maybe add some random failures to createGame?
The alternative could be to make a new sample director snippet that shows the above instead of baking it into a tutorial. Where could I put that?

yfei1:
> I looked through some of the tutorials to try and figure out where this would make sense to add some examples, but they're all pretty optimistic. I would need to break apart the assign(matches) into something more akin to
> createGame(match)
> if err -> release(match) continue
> assign(match)
> 
> and maybe add some random failures to createGame?
> The alternative could be to make a new sample director snippet that shows the above instead of baking it into a tutorial. Where could I put that?

I think the tricky thing is the current code under the `tutorial` folder is tightly coupled with what we have in [open match tutorials](https://open-match.dev/site/docs/tutorials/matchmaker101/), so any changes to the code require a rewrite of the tutorial logics. Given that we'll be releasing 0.9 by the end of this month and we still have a reasonable amount of API changes that we wanna check into the release, I propose adding a `How to handle errors` or `FAQ` tab under the current tutorials. WDYT?

calebatwd:
> > I looked through some of the tutorials to try and figure out where this would make sense to add some examples, but they're all pretty optimistic. I would need to break apart the assign(matches) into something more akin to
> > createGame(match)
> > if err -> release(match) continue
> > assign(match)
> > and maybe add some random failures to createGame?
> > The alternative could be to make a new sample director snippet that shows the above instead of baking it into a tutorial. Where could I put that?
> 
> I think the tricky thing is the current code under the `tutorial` folder is tightly coupled with what we have in [open match tutorials](https://open-match.dev/site/docs/tutorials/matchmaker101/), so any changes to the code require a rewrite of the tutorial logics. Given that we'll be releasing 0.9 by the end of this month and we still have a reasonable amount of API changes that we wanna check into the release, I propose adding a `How to handle errors` or `FAQ` tab under the current tutorials. WDYT?

I like it. What's the process here. Do I need more tests or do I need to complete documentation and FAQs for the new api? Or do we checkin and do subsequent pr with the description? It's going to be a week before I can get to doing any tutorial work for this api.

api/backend.proto
```diff
@@ -110,4 +118,14 @@ service Backend {
       body: "*"
     };
   }
+
+  // ReleaseTickets removes the submitted tickets from the list that prevents tickets 
+  // that are awaiting assignment from appearing in MMF queries, effectively putting them back into
+  // the matchmaking pool
```
Laremere: Please add the following warning:
```
BETA FEATURE WARNING:  This call and the associated Request and Response messages are
not finalized and still subject to possible change or removal.
```

PR #1060, opened 2020-01-18
Revamp gRPC grafana dashboard

yfei1:
This is the first commit to revamp the dashboard and metrics. This commit added a few panels under the gRPC tab to track the *gRPC error rate*, client request rate, server response rate as an overview. 
Also added the gRPC call latency graph and network IO graph for each of our gRPC calls.

This commit removed the need to instrument and monitor the latency of the internal statestore calls.

https://snapshot.raintank.io/dashboard/snapshot/iw0CZ6XpkwOWBztjjE3ZcpotZ33XdrKc


Laremere:
Please provide raintank published dashboard.

sawagh:
Are we removing these because they are not useful? Or because we think they do not give us much more than what gRPC latency charts already give us?

yfei1:
> Are we removing these because they are not useful? Or because we think they do not give us much more than what gRPC latency charts already give us?

This new dashboard is based on the metrics that I want to monitor when running the scale test and it fits our current needs better. The original chart is useful, but lack of some detailed numbers, e.g. gRPC error rate and network I/O per method.

PR #1075, opened 2020-01-29
Starts streaming when sending requests from synchronizer to evaluator

yfei1:
This commit started to stream when calling the evaluator.Evaluate method such that the synchronizer is able to process the data more efficiently.

### DIFF
--- a/internal/app/synchronizer/evaluator_client.go
+++ b/internal/app/synchronizer/evaluator_client.go
@@ -106,7 +106,6 @@ func (ec *grcpEvaluatorClient) evaluate(ctx context.Context, pc <-chan []*pb.Mat
 	}
 
 	sc := make(chan error, 1)
-	defer close(sc)
 	go func() {
 		defer close(sc)
 		for proposals := range pc {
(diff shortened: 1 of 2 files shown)


internal/app/synchronizer/evaluator_client.go
```diff
@@ -99,23 +99,29 @@ func newGrpcEvaluator(cfg config.View) (evaluator, error) {
 	}, nil
 }
 
-func (ec *grcpEvaluatorClient) evaluate(ctx context.Context, proposals []*pb.Match) ([]*pb.Match, error) {
+func (ec *grcpEvaluatorClient) evaluate(ctx context.Context, pc <-chan []*pb.Match) ([]*pb.Match, error) {
```
Laremere: // at start
sendError := make(chan error, 1)

// before return in the go routine:
sendError <- err

// before end
err = <-sendError
if err != nil {
  return nil, err
}
return results, nil

Laremere: oh also send nil to sendError at end of go routine

yfei1: I made a slight twist to it to have this change align with the pattern that i used to handle the errors in http calls at line 168, could review this PR again?

PR #1077, opened 2020-01-29
Stream proposals from mmf to synchronizer

Laremere:
This improves efficiency for overall system latency, and sets up for better mmf error handling.

The overall structure of the fetch matches call has been reworked.  The different go routines now set an explicit err variable.  So once we have FetchSummary, we can just set the mmf err variable on it.  Synchronizer calls which err will always result in an error here (as it's relatively fatal), while mmf and evaluator errors will be passed gently to the client.

One thing this code isn't doing anymore is checking if an mmf returns a match with no tickets.  This seems fine to me, but willing to discuss if anyone disagrees.

Deleted the tests for the following reasons:
- TestDoFetchMatchesInChannel didn't actually test fetching matches, it only tested creating a client.  Since callMmf now both creates the client and makes the call, this code now blocks actually trying to make a connection.  I'm not worried about having full branch test coverage on err statements...
- TestDoFetchMatchesFilterChannel tested merging of mmf runs.  Since there's only one mmf run now, it's no longer necessary.


internal/app/backend/backend_service.go
```diff
@@ -75,198 +70,194 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
+
+	// TODO: Send mmf error in FetchSummary instead of erroring call.
+	if synchronizeSendErr != nil || synchronizeRecvErr != nil || mmfErr != nil {
+		logger.WithFields(logrus.Fields{
+			"synchronizeSendErr": synchronizeSendErr,
+			"synchronizeRecvErr": synchronizeRecvErr,
+			"mmfErr":             mmfErr,
+		}).Error("error(s) in FetchMatches call.")
+
+		return fmt.Errorf(
```
yfei1: Would it make more sense to use `status.Errorf` at here?

internal/app/backend/backend_service.go
```diff
@@ -75,198 +70,194 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
...
-	for i := 0; i < 2; i++ {
-		err := <-errors
-		if err != nil {
-			logger.WithFields(logrus.Fields{
-				"error": err.Error(),
-			}).Error("error in FetchMatches call.")
-			return err
+		if resp.Match != nil {
+			telemetry.RecordUnitMeasurement(stream.Context(), mMatchesFetched)
+			err = stream.Send(&pb.FetchMatchesResponse{Match: resp.Match})
```
yfei1: nit: resp.GetMatch()

internal/app/backend/backend_service.go
Laremere: The way I'm combining errors, not really.  It's not clear, eg, if there are different status codes what the combined one should be.  FetchSummary will alleviate this and this call shouldn't ever end with an error being returned.

internal/app/backend/backend_service.go
Laremere: err is nil, so resp is not nil.  There's also the above line resp.Match != nil, which guarantees there to be a value here.

yfei1: I understand the code logic, but I am more concerned about using the proto getters as a standard in our repo.

PR #1078, opened 2020-01-30
Added sync.Map to backend and synchronizer

### LINKED ISSUE #939: API change: Have evaluator return only match ids
## Overview

Instead of the evaluator returning a stream of matches, only return the ids of the matches it wants to form.

## Motivation and  Impact

Currently, the evaluator returns the entire protobuf of the matches it wishes to form.  This means that the entire protobuf is re-serialized and sent back to the caller which sent the protobuf in the first place.  (The same thing happens in the backend to synchronizer call, but that's internal, so can be changed without an externally visible breaking change.)  It would be much more efficient to only return the id.

One downside is that the evaluator can't change the contents of any matches (eg, dropping a player seen in another match, but still making the match as a whole.)  It's not clear that it's supported anyway (I don't know what the code does today.)  Generally, I think it's better if things are treated as immutable.

## Change Proto

From:
```protobuf
message EvaluateResponse {
  // A Match shortlisted by the evaluator representing one of the final results.
  Match match = 1;
}
```

To:
```protobuf
message EvaluateResponse {
  // A Match shortlisted by the evaluator representing one of the final results.
  string match_id = 1;
}
```

yfei1:
This is an intermediate step to resolve #939. Leaving a bunch of TODOs in this PR and will fix them after the proto change.

### DIFF
--- a/internal/app/synchronizer/synchronizer_service.go
+++ b/internal/app/synchronizer/synchronizer_service.go
@@ -436,7 +436,6 @@ func (s *synchronizerService) addMatchesToIgnoreList(ctx context.Context, m *syn
 		// For now assume m4c is a channel of MatchIds, will fix it in a later PR.
 		ids := []string{}
 		for _, match := range matches {
-
 			tids, ok := m.Load(match.GetMatchId())
 			if ok {
 				ids = append(ids, tids.([]string)...)
--- a/internal/app/synchronizer/synchronizer_service.go
+++ b/internal/app/synchronizer/synchronizer_service.go
@@ -225,11 +225,11 @@ func (s *synchronizerService) runCycle() {
 		}
 	}()
 
-	mIDToTIDs := &sync.Map{}
-	go s.cacheMatchIDToTicketIDs(mIDToTIDs, bufferChannel(m3c), m3point5c)
+	matchTickets := &sync.Map{}
+	go s.cacheMatchIDToTicketIDs(matchTickets, m3c, m3point5c)
 	go s.wrapEvaluator(ctx, cancel, bufferChannel(m3point5c), m4c)
 	go func() {
-		s.addMatchesToIgnoreList(ctx, mIDToTIDs, cancel, bufferChannel(m4c), m5c)
+		s.addMatchesToIgnoreList(ctx, matchTickets, cancel, bufferChannel(m4c), m5c)
 		// Wait for ignore list, but not all matches returned, the next cycle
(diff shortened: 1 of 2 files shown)


internal/app/synchronizer/synchronizer_service.go
```diff
@@ -221,9 +224,12 @@ func (s *synchronizerService) runCycle() {
 			close(r.m6c)
 		}
 	}()
-	go s.wrapEvaluator(ctx, cancel, bufferChannel(m3c), m4c)
+
+	mIDToTIDs := &sync.Map{}
```
Laremere: matchTickets ?

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -221,9 +224,12 @@ func (s *synchronizerService) runCycle() {
...
+	go s.cacheMatchIDToTicketIDs(mIDToTIDs, bufferChannel(m3c), m3point5c)
```
Laremere: this shouldn't be buffered

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -396,23 +402,44 @@ func (s *synchronizerService) wrapEvaluator(ctx context.Context, cancel cancelEr
 	close(m4c)
 }
 
+func (s *synchronizerService) cacheMatchIDToTicketIDs(m *sync.Map, m3c <-chan []*pb.Match, m3point5c chan<- *pb.Match) {
```
Laremere: should this be in a comment section (/////////////////////////////////////////////)

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -396,23 +402,44 @@ func (s *synchronizerService) wrapEvaluator(ctx context.Context, cancel cancelEr
...
+		// For now assume m4c is a channel of MatchIds, will fix it in a later PR.
 		ids := []string{}
 		for _, match := range matches {
-			for _, ticket := range match.GetTickets() {
-				ids = append(ids, ticket.GetId())
+			tids, ok := m.Load(match.GetMatchId())
+			if !ok {
+				logger.Errorf("failed to get MatchId %s with its corresponding tickets from the cache", match.GetMatchId())
 			}
+			ids = append(ids, tids.([]string)...)
```
Laremere: if ok {
 ids = append
} else {
 log
}

PR #1080, opened 2020-01-30
Changed Synchronizer proto to return matchIDs instead

yfei1:
This commit changed Synchronizer proto to return matchIDs instead. Also bumped up the numbers of the unnamed channels in synchronizer starting from m3c and changed the channel type starting from `m4c` to `chan string` as the next step of the API change is to have the evaluator returns the match ids instead.

```diff
message SynchronizeResponse {
  // Instructs the backend call that it can start running the mmfs.
  bool start_mmfs = 1;

  // Instructs the backend call that it should cancel any RPC calls to the mmfs,
  // not send any more matches, and close the send stream.
  bool cancel_mmfs = 2;

-  // A match which passes evaluation and should be returned to the FetchMatches
-  // caller.
-  openmatch.Match match = 3;
+  // A match ID returned by the evaluator and should be returned to the FetchMatches
+  // caller.
+  string match_id = 4;
+
+  // Deprecated fields.
+  reserved 3;
}
```


internal/app/synchronizer/synchronizer_service.go
```diff
@@ -46,14 +46,14 @@ var (
 //  -> m1c ->
 // close incoming when done or timed out | newCutoffSender
 //   -> m2c ->
-// remember return channel m6c for match | fanInFanOut
+// remember return channel m7c for match | fanInFanOut
 //   -> m3c -> (buffered)
```
Laremere: Add a comment for your new step.

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -542,6 +539,44 @@ func bufferChannel(in chan *pb.Match) chan []*pb.Match {
 	return out
 }
 
+// bufferStringChannel collects matches from the input, and sends
+// slice of matches on the output.  It never (for long) blocks
```
Laremere: nit: strings

PR #1084, opened 2020-01-30
Deprecate Rosters, remove from Match, MatchProfiles

sawagh:


api/messages.proto
```diff
@@ -164,16 +152,8 @@ message MatchProfile {
...
   // Set of pools to be queried when generating a match for this MatchProfile.
-  // The pool names can be used in empty Rosters to specify composition of a
-  // match.
   repeated Pool pools = 3;
 
-  // Set of Rosters for this match request. Could be empty Rosters used to
-  // indicate the composition of the generated Match or they could be partially
-  // pre-populated Ticket list to be used in scenarios such as backfill / join
-  // in progress.
-  repeated Roster rosters = 4;
```
Laremere: Add to reserved.

api/messages.proto
```diff
@@ -201,9 +181,6 @@ message Match {
   // Tickets belonging to this match.
   repeated Ticket tickets = 4;
 
-  // Set of Rosters that comprise this Match
-  repeated Roster rosters = 5;
```
Laremere: Add to reserved.

api/messages.proto
sawagh: Done.

api/messages.proto
sawagh: Done

PR #1086, opened 2020-01-31
Deprecated mmf harness

yfei1:
This is the last bit of effort to deprecate the previous MMF harness. The MMF harness was still in the pkg package since the demo soloduel match function was still using it. This commit rewrote the glue code of the soloduel mmf and removed its dependency on the MMF harness, plus deleted the `pkg/harness directory`.


yfei1:
> Where does soloduel get used now? Is it still used by the demo?

So the soloduel image is solely get used by the demo. Yet the concept of soloduel is still being used under the [first_match_scenario](https://github.com/googleforgames/open-match/blob/master/examples/scale/scenarios/first_match_scenario.go#L45) file.

PR #1083, opened 2020-01-30
Update proto descriptions to reflect API changes

yfei1:
This commit updated the proto descriptions to have it better align with our API and behavior changes.

e.g.
```diff
message FetchMatchesResponse {
  // A Match generated by the user-defined MMF with the specified MatchProfiles.
-  // A valid Match response will contain at least one ticket.	
  Match match = 1;
}
```

and 

```diff
service BackendService {
- // FetchMatches triggers a MatchFunction with the specified MatchProfiles, while each MatchProfile
- // returns a set of match proposals.
+ // FetchMatches triggers a MatchFunction with the specified MatchProfile and returns a set of match proposals that 
+ // match the description of that MatchProfile.
  rpc FetchMatches(FetchMatchesRequest) returns (stream FetchMatchesResponse) {
}
```

PR #1090, opened 2020-02-03
Update proto descriptions to reflect API changes

yfei1:
Same as #1083.

### DIFF
--- a/api/query.proto
+++ b/api/query.proto
@@ -61,7 +61,7 @@ message QueryTicketsRequest {
 }
 
 message QueryTicketsResponse {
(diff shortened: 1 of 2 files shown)


api/query.proto
```diff
@@ -61,7 +61,7 @@ message QueryTicketsRequest {
 }
 
 message QueryTicketsResponse {
-  // Tickets is a list of Ticket representing one or more Tickets which meet all Filter criterias.
+  // Tickets is a list of Ticket that satisfies all the filtering criterias.
```
sawagh: Simply shorten to - 'Tickets that satisfy all the filtering criteria'

PR #1092, opened 2020-02-03
Drafted a short README for the benchmarking framework

yfei1:
Added a short README for users to get started with the framework.


sawagh:
Given I see that the structure for the scenarios is changing, lets wait for that change to get in and then update this PR.

sawagh:
Check this in so long as it is updated to the latest instructions. If not, please update and check in. Also, at some point we will need documentation for this on the website - but we could handle that post v0.10

PR #1107, opened 2020-02-05
Remove stats processor from scale tests

Laremere:
Two reasons:
Using Prometheus and Grafana for the stats processing have proved to be fruitful.  No reason to reinvent the wheel for these tests.

For the error logging, just going to standard method of error logging.  When looking at logs for errors previously, it could be hard to figure out which errors were happening when, because it would always log all the errors which have occurred.  Additionally, when an error message was unique because it contained a ticket ID, the number of errors logged each iteration would grow unbounded.


yfei1:
We need to figure out a way to share the stackdriver dashboard as a template with errors aggregated by the error message. Can you open an issue to track this?

PR #1109, opened 2020-02-06
Reduced Redis PING check frequency on Redis pool

yfei1:
As https://snapshot.raintank.io/dashboard/snapshot/IDKoZqYacXMYcECikPgK9GYJPTn10kEu pointed out, 20% of the Redis commands are spent on doing Ping checks when being returned to the pool. This commit checks if this connection have been returned for more than 10 seconds, and only do a PING check if so.

Performance improvement after this change comparing to our current benchmark result at 3k CreateTicketQPS.
https://snapshot.raintank.io/dashboard/snapshot/Wr3lY7TlRN6NIQbF1eOdd1EBNiijKA5f

### DIFF
--- a/internal/statestore/redis.go
+++ b/internal/statestore/redis.go
@@ -83,8 +83,8 @@ func newRedis(cfg config.View) Service {
 		IdleTimeout: cfg.GetDuration("redis.pool.idleTimeout"),
 		Wait:        true,
 		TestOnBorrow: func(c redis.Conn, t time.Time) error {
-			// Check if this connection have been returned for more than 10 seconds, if so, do a PING check.
-			if time.Since(t) > 10*time.Second {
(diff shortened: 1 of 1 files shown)


internal/statestore/redis.go
```diff
@@ -82,7 +82,12 @@ func newRedis(cfg config.View) Service {
 		MaxActive:   cfg.GetInt("redis.pool.maxActive"),
 		IdleTimeout: cfg.GetDuration("redis.pool.idleTimeout"),
 		Wait:        true,
-		TestOnBorrow: func(c redis.Conn, _ time.Time) error {
+		TestOnBorrow: func(c redis.Conn, t time.Time) error {
+			// Check if this connection have been returned for more than 15 seconds, if so, do a PING check.
```
Laremere: This comment isn't clear without context.  What does "this connection have been returned for more than 15 seconds" mean precisely?

PR #1110, opened 2020-02-06
Move scale scenarios into unique packages

Laremere:
Per an earlier discussion, this makes it easier to define things such as "poolName" without conflicting other scenarios.  It also goes to define the logical scenario, which will be separate from the infra scenario being tested.

Future pieces built on top of this:
One more logical scenario for a session based game with skill, modes, and regions.
Different infra scenarios, for different usage patterns. (eg, frontend waiting for assignments.)


examples/scale/scenarios/scenarios.go
```diff
@@ -14,10 +14,50 @@
...
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	"open-match.dev/open-match/examples/scale/scenarios/battleroyal"
+	"open-match.dev/open-match/examples/scale/scenarios/firstmatch"
+	"open-match.dev/open-match/internal/util/testing"
+	"open-match.dev/open-match/pkg/matchfunction"
+	"open-match.dev/open-match/pkg/pb"
+)
+
+var _ = firstmatch.Scenario()
```
sawagh: What does this achieve? A comment might help.

examples/scale/scenarios/scenarios.go
```diff
@@ -14,10 +14,50 @@
...
+		Ticket:   ls.Ticket,
+		Profiles: ls.Profiles,
+
+		MMF:       queryPoolsWrapper(ls.MatchFunction),
+		Evaluator: ls.Evaluate,
+	}
+}()
 
 // Scenario defines the controllable fields for Open Match benchmark scenarios
 type Scenario struct {
```
sawagh: Now that I understand these, here are my 2 cents on naming :) Just reading without any pre-conceived notions.

I think BattleRoyale, FirstMatch ... fit well as a different packages that follow the Scenario interface. This structure (existing scenario) seems more appropriately as - Parameters / ScenarioParameters / ScenarioConfiguration / ScenarioOptions ... ?

examples/scale/scenarios/scenarios.go
```diff
@@ -48,6 +88,13 @@ type Scenario struct {
 	Evaluator evaluatorFunction
 }
 
+type LogicalScenario interface {
```
sawagh: Naming suggestion above - but can you please add more comments over these interfaces to add clarity?  Also moving the interface declaration above its first usage in the file will help.

examples/scale/scenarios/scenarios.go
Laremere: Forgot to remove.

examples/scale/scenarios/scenarios.go
Laremere: updated per discussion

PR #1116, opened 2020-02-09
have scale-frontend use individual go routines for each ticket

Laremere:
This creates tickets from individual go routines instead of using a worker pool.  The logical change here is not large, but it sets up for followup PRs which add waiting for and deleting tickets from the frontend.  The worker model doesn't work when a large number of tickets are calling waitForAssignments.

The random sleep at the start of the worker evens calls out over the second period, and makes timing between ticket creation calls a more realistic poisson distribution.

Added metrics and graphs for the current number of tickets in each phase.

### DIFF
--- a/examples/scale/frontend/frontend.go
+++ b/examples/scale/frontend/frontend.go
@@ -17,9 +17,11 @@ package frontend
 import (
 	"context"
 	"math/rand"
+	"sync"
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"go.opencensus.io/stats"
 	"go.opencensus.io/trace"
 	"open-match.dev/open-match/examples/scale/scenarios"
 	"open-match.dev/open-match/internal/config"
@@ -37,7 +39,8 @@ var (
 
 	mTicketsCreated        = telemetry.Counter("scale_frontend_tickets_created", "tickets created")
 	mTicketCreationsFailed = telemetry.Counter("scale_frontend_ticket_creations_failed", "tickets created")
-	mOutstandingRunners    = telemetry.Gauge("scale_frontend_outstanding_runners", "outstanding runners")
+	mRunnersWaiting        = concurrentGauge(telemetry.Gauge("scale_frontend_runners_waiting", "runners waiting"))
+	mRunnersCreating       = concurrentGauge(telemetry.Gauge("scale_frontend_runners_creating", "runners creating"))
 )
 
 // Run triggers execution of the scale frontend component that creates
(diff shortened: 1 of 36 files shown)


examples/scale/frontend/frontend.go
```diff
@@ -60,73 +60,89 @@ func run(cfg config.View) {
...
-		var wg sync.WaitGroup
-		for i := 0; i < numOfRoutineCreate; i++ {
-			wg.Add(1)
-			if i < ticketModRoutine {
-				go createPerCycle(&wg, fe, ticketPerRoutine+1, start)
-			} else {
-				go createPerCycle(&wg, fe, ticketPerRoutine, start)
-			}
-		}
+	g.start(mRunnersWaiting)
```
sawagh: Why is runners waiting an interesting metric? Isnt creating the only interesting one? Given that there is no failure condition between running / creating, wont every running get to creating? If there was only creating, could the usage of state gauge be avoided and we simply have mRunnersCreating(1) with a defer to mRunnersCreating(-1)?

examples/scale/frontend/frontend.go
```diff
@@ -60,73 +60,89 @@ func run(cfg config.View) {
...
+	time.Sleep(time.Duration(rand.Int63n(int64(time.Second))))
```
sawagh: Please add a comment (either on this line or function or just header on the file) stating why the sleep exists and what it does (PR description covers that really well - but i think it is important enough to have it in context)

examples/scale/frontend/frontend.go
Laremere: It's a little interesting, as it helps the conceptual model of what each runner is doing: waiting a little bit, then creating a ticket.  In the near future there will be other stages, waiting/polling for assignment, and deleting the ticket.

Laremere:
> Is the recommendation going to be to ever run only one frontend pod - given it ensures the right QPS? Given we expect this to run for scale scenarios, would we be testing the limits of the framework itself?
> 
> (not specific to this PR - just wanting to know the eventual desired setup)

Currently what I've been doing is using the fact that each instance creates a fixed number of tickets per second, and then scaled up the number of instances to simulate higher load.  In the near future might have better control over how many tickets each instance is creating, in that case I'd flip it:  The number of instances of scale_frontned would be large, but fixed.

There's no way to avoid the fact that the scale tests also are potentially limited by the test's scaling characteristics.  We've already hit that a few times.  The frontend is really easy to scale to many instances though.

examples/scale/frontend/frontend.go
sawagh: IMO given there are no ways you fail, a measure of how long you wait before create may not tell us much (what would we infer between multiple runs - or about the system based on this metric?). I am ok it being in there for now - but as you revisit for other metrics, based on whether it is useful, please consider removing it. (sometimes having more clutter on the dashboard can be distracting :)

PR #1127, opened 2020-02-20
Do some randomization on return order of tickets

Laremere:
This prevents clients from unintentionally depending on the return order of tickets.

This is currently happening with our tests, so this includes several fixes for the tests.

### DIFF
--- a/internal/statestore/redis.go
+++ b/internal/statestore/redis.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"time"
 
 	"github.com/cenkalti/backoff"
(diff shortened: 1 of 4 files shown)


internal/statestore/redis.go
```diff
@@ -477,6 +478,15 @@ func (rb *redisBackend) FilterTickets(ctx context.Context, pool *pb.Pool, pageSi
 
 	idSet = set.Difference(idSet, idsInIgnoreLists)
 
+	// Do a little randomization to prevent dependency on return order.
```
yfei1: Could you update the PR description and the comment at here to describe the intention of this changes? e.g. `The goal of this randomization is to discourage users from taking the return order into account when implementing the MMF/Evaluator`.

Laremere: Per discussion... it already says that?

Issue #1121, opened 2020-02-13
Optimistic bulk ticket assign


calebatwd:
**What happened**:
Bulk api assignments fail if the assignment doesn't exist or if a write fails for an individual ticket

**What you expected to happen**:
The bulk ticket assign api should create tickets that no longer exist with an assignment field AND not fail the bulk api call based on whether the ticket exists or if the write of and individual succeeds or fails.

sawagh:
Would like to get more context on the expectation here. I see the value in not failing the bulk assignment. However, the auto Ticket creation concerns me. 

If a Ticket has explicitly been deleted, creating it automatically in assignment seems strange. Who is the owner of this created Ticket? Who is aware (from the frontend) that this exists then - assuming it was explicitly deleted automatically? Who will eventually cleanup  / delete this Ticket? 

Also, today a Ticket creation in Open Match assigns a generated ID to the Ticket (not taken from the user).  Should in this case, we make an exception and take the ID from the given Ticket? Can you elaborate on the scenario requiring auto ticket creation on assign?

sawagh:
Based on TSC discussion, 

It is not a key requirement that the missing Ticket be created. The primary requirement is that a failure to create a single Ticket should NOT fail the entire assignment call. 

The reasonable approach here would be to return a status object in the response that indicates the list of TicketIds that failed creation. This can be used by the caller to debug / know that some assignments failed, while the other assignments on the call still succeed.

sawagh:
I will write up a proposal for this one.

sawagh:
Proposal: Assignment will succeed even if individual assignments fail due to ticket not being present. The information about the failed assignments will be populated in the response object. The caller may use this for debugging / alerting purposes. Here is a proposed change to the proto to achieve this:

Option1: Concise but ugly! 
```
message AssignmentFailure {
  string ticket_id = 1;
  google.rpc.Status error = 2;
}

message AssignTicketsResponse {
  repeated AssignmentFailure failed = 1;
}
```

Option2 (Cleaner but redundant)
```
message AssignmentStatus {
  string ticket_id = 1;
  google.rpc.Status status = 2;
}

message AssignTicketsResponse {
  repeated AssignmentStatus status = 1;
}
```

sawagh:
Proposal: #1155 covers this and other assignment api changes.

PR #1131, opened 2020-02-25
Optimized `MULTI` `EXEC` querys to reduce Redis CPU consumption

yfei1:
This commit rewrote some Redis queries to reduce the CPU consumption. The overall idea is to eliminate the usage of unnecessary `MULTI`-`EXEC` pattern to reduce the number of commands Redis needs to process per second.

The `MULTI`-`EXEC` pattern was used to stream the queries to Redis to avoid possible network bottleneck. However, after the evaluator API change in v0.9, the overall network I/O number is not as large as it was.

Tested and can confirm the following changes can improve the system throughput by 10% under 1v1 scenario.

1. Removed the `MULTI`, `EXEC` pattern in CreateTicket since it is an overkill for two subcommands.
2. Removed the `MULTI`, `EXEC` pattern in UpdateAssignments and use a `MGET` and `MSET` instead.
3. Removed the `MULTI`, `EXEC` pattern in AddTicketsToIgnoreList and DeleteTicketsFromIgnoreList and use `ZREM` and `ZADD` with multi args instead.


Laremere:
Please fix build.

third_party/google/api/backend.proto
```diff
@@ -96,6 +96,25 @@ message BackendRule {
   string selector = 1;
 
   // The address of the API backend.
+  //
+  // The scheme is used to determine the backend protocol and security.
```
sawagh: Auto update to proto files unrelated to this? Is this just stale by now and wont be needed when you check this in?

internal/statestore/redis.go
```diff
@@ -192,7 +183,7 @@ func (rb *redisBackend) CreateTicket(ctx context.Context, ticket *pb.Ticket) err
 	if rb.cfg.IsSet("redis.expiration") {
 		redisTTL := rb.cfg.GetInt("redis.expiration")
 		if redisTTL > 0 {
-			err = redisConn.Send("EXPIRE", ticket.GetId(), redisTTL)
+			_, err = redisConn.Do("EXPIRE", ticket.GetId(), redisTTL)
```
sawagh: revert any change here as this is redundant with https://github.com/googleforgames/open-match/pull/1146?

internal/statestore/redis.go
```diff
@@ -672,24 +659,15 @@ func (rb *redisBackend) DeleteTicketsFromIgnoreList(ctx context.Context, ids []s
...
 	defer handleConnectionClose(&redisConn)
 
-	err = redisConn.Send("MULTI")
-	if err != nil {
-		redisLogger.WithError(err).Error("failed to pipeline commands for DeleteTicketsFromIgnoreList")
-		return status.Error(codes.Internal, err.Error())
+	cmds := make([]interface{}, len(ids)+1)
+	cmds[0] = "proposed_ticket_ids"
+	for i, id := range ids {
+		cmds[i+1] = id
```
sawagh: minor nit - append() over index manipulation? less bug prone for future changes?

internal/statestore/redis.go
```diff
@@ -628,32 +620,27 @@ func (rb *redisBackend) GetAssignments(ctx context.Context, id string, callback
...
 	currentTime := time.Now().UnixNano()
-	for _, id := range ids {
-		// Index the DoubleArg by value.
-		err = redisConn.Send("ZADD", "proposed_ticket_ids", currentTime, id)
-		if err != nil {
-			redisLogger.WithError(err).Error("failed to append proposed tickets to redis")
-			return status.Error(codes.Internal, err.Error())
-		}
+	cmds := make([]interface{}, 2*len(ids)+1)
+	cmds[0] = "proposed_ticket_ids"
```
sawagh: can we use append() over index manipulation? (other places too - easier to read and possibly less bug prone)

internal/statestore/redis.go
```diff
@@ -628,32 +620,27 @@ func (rb *redisBackend) GetAssignments(ctx context.Context, id string, callback
...
+		cmds[2*i+1] = currentTime
+		cmds[2*i+2] = id
 	}
 
-	// Run pipelined Redis commands.
-	_, err = redisConn.Do("EXEC")
+	_, err = redisConn.Do("ZADD", cmds...)
 	if err != nil {
-		redisLogger.WithError(err).Error("failed to execute pipelined commands for AddTicketsToIgnoreList")
+		redisLogger.WithError(err).Error("failed to append proposed tickets to redis")
```
sawagh: append proposed tickets to ignore list? 

Easier to understand when you see that in debug logs?

internal/statestore/redis.go
```diff
@@ -535,36 +516,45 @@ func (rb *redisBackend) UpdateAssignments(ctx context.Context, ids []string, ass
...
 	}
 	defer handleConnectionClose(&redisConn)
 
-	err = redisConn.Send("MULTI")
+	b := make([]interface{}, len(ids))
+	for i := range ids {
+		b[i] = ids[i]
+	}
+
+	ticketBytes, err := redis.ByteSlices(redisConn.Do("MGET", b...))
```
sawagh: How does this surface error with missing Tickets? Can we isolate that error from any other redis error? I ask because we have issue https://github.com/googleforgames/open-match/issues/1121 which most certainly need us to ignore 'Not Found' failures (not just ignore - but idenitfy the IDs for these and pass them back in some status object) so that makes it critical to identify and ignore the not found errors from the rest. Possible to do that here?

internal/statestore/redis.go
```diff
@@ -535,36 +516,45 @@ func (rb *redisBackend) UpdateAssignments(ctx context.Context, ids []string, ass
...
-			if err != nil {
-				redisLogger.WithError(err).Errorf("failed to get ticket %s from redis when updating assignments", id)
-				return err
-			}
-			tickets = append(tickets, ticket)
+	tickets := make([]*pb.Ticket, 0, len(ticketBytes))
+	for i, ticketByte := range ticketBytes {
+		// Tickets may be deleted by the time we read it from redis.
+		if ticketByte == nil {
+			return status.Errorf(codes.NotFound, "failed to get ticket %s from statestore", ids[i])
```
sawagh: Aha. This seems to suggest that MGET will return nil object for missing ticket id? if that is so - ignore and resolve previous comment.

yfei1: yup, the missing objects are nil

PR #1134, opened 2020-02-26
Fixed a ticket format that may result in serialization errors

yfei1:
We were using `proto.Marshal` and `proto.Unmarshal` to serial and de-serialize the tickets object. This flow works when we are using `SET` and `GET` against tickets but may result in serialization errors when we process tickets using `MSET` and `MGET`.

According to Redis command doc, this commit changed the serialization method from `proto.Marshal`/`proto.Unmarshal` to `proto.MarshalTextString`/`proto.UnmarshalText` in order to fix the potential issue.
> GET
 Get the value of key. If the key does not exist the special value nil is returned. An error is returned if the value stored at key is not a string, because GET only handles string values.

### DIFF
--- a/internal/statestore/redis.go
+++ b/internal/statestore/redis.go
@@ -170,16 +170,7 @@ func (rb *redisBackend) CreateTicket(ctx context.Context, ticket *pb.Ticket) err
 		return status.Errorf(codes.Internal, "%v", err)
 	}
 
-	value, err := proto.Marshal(ticket)
-	if err != nil {
-		redisLogger.WithFields(logrus.Fields{
-			"key":   ticket.GetId(),
-			"error": err.Error(),
-		}).Error("failed to marshal the ticket proto")
-		return status.Errorf(codes.Internal, "%v", err)
-	}
-
-	err = redisConn.Send("SET", ticket.GetId(), value)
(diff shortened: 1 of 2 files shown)


internal/statestore/redis.go
```diff
@@ -170,16 +170,7 @@ func (rb *redisBackend) CreateTicket(ctx context.Context, ticket *pb.Ticket) err
...
-	if err != nil {
-		redisLogger.WithFields(logrus.Fields{
-			"key":   ticket.GetId(),
-			"error": err.Error(),
-		}).Error("failed to marshal the ticket proto")
-		return status.Errorf(codes.Internal, "%v", err)
-	}
-
-	err = redisConn.Send("SET", ticket.GetId(), value)
+	err = redisConn.Send("SET", ticket.GetId(), proto.MarshalTextString(ticket))
```
Laremere: This will "serialize" using protobuf's text format, which has a number of large disadvantages:
- It's slower.
- It's much larger.
- I think unrecognized fields are dropped.
- Field renames aren't forward/backwards compatible.

You should be instead be serializing the proto to bytes, then serializing those bytes into a string which works on redis, eg https://golang.org/pkg/encoding/base64/

Issue #1123, opened 2020-02-15
Proposal: Use query for caching and filtering


Laremere:
This is a proposal to switch from doing any indexing and filtering with redis to doing most of that work in memory with query.

Query's memory contains a map from ticket id to ticket pb.

Query would operate in a loop:

1. Wait for a first request.
2. Collect all waiting requests.
3. Query for list of current ids.
4. Delete all tickets in the query map that are in the active list.
5. Query for all tickets bytes for missing ticket ids.
6. Run the request(s) on the pool in parallel.
7. Loop back to 1.

In the event that a request is received after 1or2, the request will wait for the next iteration.  The expectation is that under load, many requests may be waiting for the iteration.

Few notes:

- Getting all ids uss the same methodology as is used today, so no change in correctness.
- Reading the whole state sounds scary, but it's really what OM is doing already:  Tickets are very quickly created, read many times, then deleted.  If a ticket isn't in any pools, then it won't get matches.  As such in the current implementation tends to read the tickets at least once (possibly many times) from redis every iteration.  This leads to a constant read load, once for every query instance.  Unless you're draining the entire pool every iteration, this will lead to fewer requests on redis.
- Filtering in memory is very fast.  As a bonus, it's also flexible, which means that we can very easily add new filter types in the future.
- Ticket creation no longer requires indexing each field.  So that will reduce load.

A working prototype is available here: https://github.com/googleforgames/open-match/compare/master...Laremere:queryCache

Here's the scale test showing that it was able to handle 960 profiles with ease: https://snapshot.raintank.io/dashboard/snapshot/evdz1XK1E701owZr2u47wUVMrQYg5k5K?orgId=2

Laremere:
Oh, and if a ticket is in the all-tickets index, but not present when the whole bytes is queried for, it will not be added to the cache.  This is equivalent to what we do today.

sawagh:
Quick Question before I leave my detailed notes here. 

In the Synchronizer, when we cancel a pending function that over-runs past the synchronization window, do we wait for that cancelled function to terminate? or is the cancellation best effort and the function could possibly continue to do work but will fail to stream results?

Laremere:
The cancellation notification from the sycnchronizer to backend is purely an fyi.
Once the backend get it, it cancels the context used to call the mmf.  This will cancel the call from the backend's point of view.
If the mmf does proper cancel propagation (go's context does this, unsure about other languages), the call to query will be canceled.

This all is already in.

[...]

PR #1135, opened 2020-02-26
Replace redis indexing with in memory cache

Laremere:
As detailed in https://github.com/googleforgames/open-match/issues/1123

TestDoQueryTickets has three test cases:
Nothing returned because pool is empty.
Some tickets returned with filtering.
Error sending tickets back to caller.
The first two are covered by e2e tests, and the third is no longer relevant, because of the changed statestore interface.

TestTicketIndexing is removed from redis_test, as indexing is now a lot simpler, and so is tested by the other two tests (and all the e2e tests on query....)

backend_service_test
TestDoReleaseTickets has three tests:
- Code “unavailable” when context is already canceled - why is this a test???
- release tickets with no tickets specified is ok
- release tickets works (few tickets added to ignore list, then are released)

TestDoAssignTickets
- Code “unavailable” when context is already canceled - why is this a test???
- invalid argument if assignment is nill
- not found if ticket does not exist
- ok code on successful assignment

Out of these, the context canceled don’t make any sense as a test.  The other ReleaseTicket tests seem reasonable.  The ticket assignment cases are already covered in e2e tests.


yfei1:
```
--- FAIL: TestTicketFound (0.52s)
    --- FAIL: TestTicketFound/TagPresent_multiple_all_present (0.10s)
        query_tickets_test.go:154: 
            	Error Trace:	query_tickets_test.go:154
            	            				query_tickets_test.go:114
            	Error:      	Expected nil, but got: &status.statusError{Code:13, Message:"sync: negative WaitGroup counter", Details:[]*any.Any(nil), XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}
```

internal/statestore/redis.go
```diff
@@ -447,84 +348,75 @@ func (rb *redisBackend) FilterTickets(ctx context.Context, pool *pb.Pool, pageSi
...
+}
 
-		if i == 0 {
-			idSet = idsInFilter
-		} else {
-			idSet = set.Intersection(idSet, idsInFilter)
-		}
+// GetTickets returns multiple tickets from storage.  Missing tickets are
+// silently ignored.
+func (rb *redisBackend) GetTickets(ctx context.Context, ids []string) ([]*pb.Ticket, error) {
```
sawagh: What instrumentation do we have for GetTickets? The performance for this method may be interesting as size / count of tickets goes up. I see this as the only real bottleneck in the new approach?

internal/statestore/public.go
```diff
@@ -36,14 +36,18 @@ type Service interface {
...
 	IndexTicket(ctx context.Context, ticket *pb.Ticket) error
 
-	// DeindexTicket removes the indexing for the specified Ticket. Only the indexes are removed but the Ticket continues to exist.
+	// DeindexTicket removes specified ticket from the index. The Ticket continues to exist.
 	DeindexTicket(ctx context.Context, id string) error
 
-	// FilterTickets returns the Ticket ids for the Tickets meeting the specified filtering criteria.
-	FilterTickets(ctx context.Context, pool *pb.Pool, pageSize int, callback func([]*pb.Ticket) error) error
+	// GetIndexedIds returns the ids of all tickets currently indexed.
+	GetIndexedIds(ctx context.Context) (map[string]struct{}, error)
```
sawagh: I see we need a map on the receiving end for our use case - but simply looking at this API surface, the below seems more natural:
	GetIndexedIds(ctx context.Context) ([]string, error)
We can always convert it into a map at the receiving end.

internal/app/query/query_service.go
```diff
@@ -107,3 +150,108 @@ func getPageSize(cfg config.View) int {
 
 	return pSize
 }
+
+////////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+
+type ticketCache struct {
```
yfei1: +1 for Saurabh's point. Maybe change the signature to `func (s *queryService) update()` instead?

internal/statestore/public.go
```diff
@@ -36,14 +36,18 @@ type Service interface {
...
+
+	// GetTickets returns multiple tickets from storage.  Missing tickets are
+	// silently ignored.
+	GetTickets(ctx context.Context, ids []string) ([]*pb.Ticket, error)
```
yfei1: Can we combine the interface of `GetTickets` and `GetTicket` to something like `GetTickets(ctx context.Context, ids ...string)`? Having two functions solving the same things seems redundant to me. @sawagh what do you think?

internal/app/query/query_service.go
```diff
@@ -107,3 +150,108 @@ func getPageSize(cfg config.View) int {
...
+	previousCount := len(tc.listed)
+
+	currentAll, err := store.GetIndexedIds(context.Background())
+	if err != nil {
+		tc.err = err
+		return
+	}
+
+	deletedCount := 0
+	for id := range tc.listed {
```
yfei1: Would it be better to parallelize these checks?

internal/statestore/public.go
sawagh: Ah yeah. Unless the single one has any optimizations over the multiple one - we can just do with GetTickets

internal/app/query/query_service.go
Laremere: parallelize what?  This requires modifying the ticket map, which can't be done in parallel.

internal/app/query/query_service.go
```diff
@@ -107,3 +150,108 @@ func getPageSize(cfg config.View) int {
...
+
+	return results, nil
+}
+
+// Does not fail, but may set error which will be returned to clients.
+func (tc *ticketCache) update(store statestore.Service) {
+	previousCount := len(tc.listed)
+
+	currentAll, err := store.GetIndexedIds(context.Background())
+	if err != nil {
```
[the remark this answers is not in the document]
Laremere: Added log statement to queryTickets, which this error is returned to.

internal/app/query/query_service.go
```diff
@@ -107,3 +150,108 @@ func getPageSize(cfg config.View) int {
...
+
+	for id := range currentAll {
+		if _, ok := tc.listed[id]; !ok {
+			toFetch = append(toFetch, id)
+		}
+	}
+
+	newTickets, err := store.GetTickets(context.Background(), toFetch)
+	if err != nil {
+		tc.err = err
```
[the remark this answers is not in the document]
Laremere: Added log statement to queryTickets, which this error is returned to.

internal/app/query/query_service.go
```diff
@@ -107,3 +150,108 @@ func getPageSize(cfg config.View) int {
...
+		return
+	}
+
+	for _, t := range newTickets {
+		tc.listed[t.Id] = t
+	}
+
+	logger.Infof("Ticket Cache update: Previous %d, Deleted %d, Fetched %d, Current %d", previousCount, deletedCount, len(toFetch), len(tc.listed))
```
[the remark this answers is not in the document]
Laremere: Moved to debug.

internal/app/query/query_service.go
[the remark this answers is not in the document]
Laremere: I've reworked the abstraction.

I still want it separate, as there's a lot going on.  What I've arrived upon is that the cache handles concurrency and updating the ticket map, and the query call itself handles verification, filtering from the map, and streaming the results in pages.

internal/statestore/redis.go
Laremere: Are you concerned that this will be introducing a new bottleneck, or that removing other bottlenecks only leaves this as the next remaining possible slowdown?
We already used mget in filter, and ticket query load on redis was far greater for the old system.

sawagh: I am not so much concerned about bottlenecks - as much as this is probably the most interesting data point now that we have eliminated the zranges due to query. If i look at this from what charts might be most interesting if we were to analyze scale for different ticket sizes, this one should be interesting. Need not do it as a part of this PR - but i think should be added as a part of the metrics we are adding anyways.

internal/app/query/query_service.go
yfei1: Ah sorry for the confusion, I'm not referring to this specific for loop. Would it be more performant if you rewrite this function with the following pattern? 

```
var wg *sync.Waitgroup = &sync.WaitGroup{}
wg.Add(2)

go func() {
  defer wg.Done()
  // get toDelete
}

go func() {
  defer wg.Done()
  // get toFetch
}

wg.Wait()
// delete and fetch
```

internal/statestore/redis.go
Laremere: Agreed that it'd be interesting, but is potential future work.

internal/statestore/public.go
Laremere: Both the interface implementation and the caller require operating on it as a map.  Converting it to a slice only to turn it back feels like a waste.  (both cycles, and more code)

Switched the name to GetIndexedIdSet to align more with what it's returning.  Is that acceptable?

internal/statestore/public.go
Laremere: I anticipate there being a difference in the future: whether assignments are included or not.  This is if we only combine assignments on query time. see #1137 for my rambling.  Otherwise these could be combined in the future.

internal/statestore/public.go
sawagh: Designing an API for a single client use-case is a good approach. This approach is completely acceptable from say - helpers in same package where you return a type convenient for the caller as the helper is meant to be internal. 

For Interfaces that may be potentially implemented by other backends in future, focusing on more generic primitives helps. However, I do see renaming this slight improvement over the previous approach and my concern is not big enough to hold the PR.

internal/app/backend/backend_service_test.go
```diff
@@ -14,298 +14,284 @@
 
 package backend
 
-import (
-	"context"
-	"testing"
+// func TestDoReleaseTickets(t *testing.T) {
```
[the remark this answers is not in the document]
Laremere: did an assessment of the tests and moved coverage to e2e tests.

internal/app/backend/backend_service_test.go
```diff
@@ -14,298 +14,284 @@
```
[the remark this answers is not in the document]
Laremere: did an assessment of the tests and moved coverage to e2e tests.

PR #1146, opened 2020-03-05
Removed ticket auto-expiring logic from statestore

### LINKED ISSUE #815: statestore.CreateTicket does not expire ticket indices when redis.expiration is set

yfei1:
Resolved #815.

Per discussion, removed the `EXPIRE` logic from our statestore implementation.

### DIFF
--- a/install/helm/open-match/templates/om-configmap-default.yaml
+++ b/install/helm/open-match/templates/om-configmap-default.yaml
@@ -103,7 +103,6 @@ data:
         maxActive: {{ index .Values "open-match-core" "redis" "pool" "maxActive" }}
         idleTimeout: {{ index .Values "open-match-core" "redis" "pool" "idleTimeout" }}
         healthCheckTimeout: {{ index .Values "open-match-core" "redis" "pool" "healthCheckTimeout" }}
-      expiration: 43200
 
     telemetry:
       zpages:
--- a/internal/statestore/redis.go
+++ b/internal/statestore/redis.go
@@ -161,15 +161,6 @@ func (rb *redisBackend) CreateTicket(ctx context.Context, ticket *pb.Ticket) err
 	}
 	defer handleConnectionClose(&redisConn)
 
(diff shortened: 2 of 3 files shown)


sawagh:
To confirm my understanding, #815 speaks of us not expiring indices with Ticket expiration. With the query cache, we have completely done away with indices creation (except all tickets) - so if the index expiration is added to the all tickets index, wont #815 be addressed? 

My suspicion is, come customer do not manage Ticket deletion at all and just let the tickets expire and assume auto cleanup. Although that was not working as intended anyways, this change mandates customers to cleanup Tickets explicitly - correct? If so, this is a behavior change that we should add documentation for and explicitly shout out to the community. 

I think it is worth pushing this change anyways - and then creating a separate issue to document the Ticket cleanup mechanism for Open Match users - and lets get feedback on that (and re-implement changes the correct way if that turns out to be the conclusion)

yfei1:
> To confirm my understanding, #815 speaks of us not expiring indices with Ticket expiration. With the query cache, we have completely done away with indices creation (except all tickets) - so if the index expiration is added to the all tickets index, wont #815 be addressed?
> 
> My suspicion is, come customer do not manage Ticket deletion at all and just let the tickets expire and assume auto cleanup. Although that was not working as intended anyways, this change mandates customers to cleanup Tickets explicitly - correct? If so, this is a behavior change that we should add documentation for and explicitly shout out to the community.
> 
> I think it is worth pushing this change anyways - and then creating a separate issue to document the Ticket cleanup mechanism for Open Match users - and lets get feedback on that (and re-implement changes the correct way if that turns out to be the conclusion)

Yes, the query cache resolved most of the index expiration problem except for the allTickets part. The problem is, Redis doesn’t have a good way for expiring the non top-level data. E.g, there is no way in Redis to partially expire a set. We’ll need to change the current choice of data structure into something else if we want to keep this behavior untouched. I’ll open up an issue to make sure we document the behavior change and gather the feedbacks.

PR #1151, opened 2020-03-09
Added QueryTicketIds method to QueryService

### LINKED ISSUE #766: API change: Add QueryTicketIds to mmlogic
## Overview

This change would add a new service rpc endpoint which takes a pool (like QueryTicketsRequest), but only returns a list of Ticket ids.

## Motivation

There are some system level operations when working with open-match which require only the ticket ids.  These are also operations which are likely to require a large number of tickets be returned.  Examples:

- Delete all tickets in Open Match.
- Get all of the tickets, and reconcile actual OM state with what I expect to be in OM.

These operations are useful when recovering from system failures.

## Impact

New RPC endpoint, does not break any existing users.  

## Change Proto

```protobuf
message QueryTicketIdsRequest {
  // The Pool representing the set of Filters to be queried.
  Pool pool = 1;
}

message QueryTicketIdsResponse {
  // The ids of tickets that meet the Filter criteria requested by the Pool.
  repeated string ids = 1;
}

service MmLogic {
  // ... existing rpc here ...

  // QueryTicketId gets the list of ticket ids that match all Filters in the
  // specified Pool.
  rpc QueryTicketIds(QueryTicketIdsRequest) returns (stream QueryTicketIdsResponse) {
    option (google.api.http) = {
      post: "/v1/mmlogic/ticketids:query"
      body: "*"
    };
  }
}
```

yfei1:
This commit added a QueryTicketIds method to QueryService to resolve #766 and added (basically copied from QueryTickets) the tests accordingly.

Also fixed a minor dependency error in our Makefile.


api/query.proto
```diff
@@ -56,25 +56,46 @@ option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
...
 
 message QueryTicketsRequest {
-  // A Pool is consists of a set of Filters.
+  // The Pool representing the set of Filters to be queried.
   Pool pool = 1;
 }
 
 message QueryTicketsResponse {
-  // Tickets that satisfy all the filtering criteria.
+  // Tickets that meet all the filtering criteria requested by the pool.
```
Laremere: why satisfy -> meet?

internal/app/query/query_service.go
```diff
@@ -80,6 +80,43 @@ func (s *queryService) QueryTickets(req *pb.QueryTicketsRequest, responseServer
...
 
+func (s *queryService) QueryTicketIds(req *pb.QueryTicketIdsRequest, responseServer pb.QueryService_QueryTicketIdsServer) error {
+	pool := req.GetPool()
+	if pool == nil {
+		return status.Error(codes.InvalidArgument, ".pool is required")
+	}
+
+	var results []string
+	err := s.tc.request(responseServer.Context(), func(tickets map[string]*pb.Ticket) {
+		for _, ticket := range tickets {
```
Laremere: `for id, ticket := range tickets`, then use the id instead of ticket.GetId()

api/query.proto
yfei1: this is what you have in the queryid issue.

Laremere: oh, ok me fighting with the wording I gave.

PR #1162, opened 2020-03-19
Add a built in created time field for Tickets and the ability to filter Tickets by created time.

sawagh:
Here are the highlights of this change:

- Adds a create_time field to Ticket proto and populates this field during Ticket creation
- Adds created_before, created_after to the Pool and filters Tickets based on create time if this field is specified. 

One implementation nuance was that the current filter package directly operates on Pools. In this case (and may be similar in future), we need to convert the proto to time field - which doing for each Ticket is a lot of unnecessary repetitive work. Further, this conversion can fail - at which point, failing it at a Ticket level does not make sense (that validation / failure should be outside of the filter package). 

Hence, this change abstracts the Pool from the filter package and instead introduces a Criteria structure that contains validated filtering criteria. Any validations on Pool specification can happen outside (in query service) before passing the Criteria to filter package. 

One nuance is the test cases. The way the test is currently structured, the 'testcases' package is used by E2E as well as filter_tests which is great - however, with this change, the filter tests need Criteria. Hence, the subtle change is that the testcases package still uses Pool but filter_test converts that to criteria as it intends to test only the criteria whereas the e2e will continue to use more E2E Pool specification.


internal/statestore/redis.go
```diff
@@ -319,6 +320,8 @@ func (rb *redisBackend) GetTicket(ctx context.Context, id string) (*pb.Ticket, e
 		return nil, status.Errorf(codes.Internal, "%v", err)
 	}
 
+	ticket.CreateTime = ptypes.TimestampNow()
```
Laremere: This should be in the frontend, not redis.
Additionally please add validation that create time is not set before setting it here, with appropriate testing (see https://github.com/googleforgames/open-match/pull/1160/files )

internal/filter/filter.go
```diff
@@ -18,18 +18,60 @@
...
+// MeetsCriteria returns whether the ticket meets all the criteria of the pool.
+func MeetsCriteria(ticket *pb.Ticket, fc *Criteria) bool {
 	s := ticket.GetSearchFields()
 	if s == nil {
 		s = emptySearchFields
 	}
-	for _, f := range pool.GetDoubleRangeFilters() {
+
+	// CreateTime is only populated by Open Match and hence expected to be valid.
+	if ct, err := ptypes.Timestamp(ticket.CreateTime); err == nil {
```
Laremere: I don't quite like how this needs to be parsed every time.  It will be parsed for every ticket for every query, regardless of if it uses CreatedBefore / CreatedAfter.

We're setting this field, so it **should** be valid always.  Hm.
Maybe we need a type for a parsed ticket too... I don't know.  What are your thoughts?

If we are logging, it should be an error, because the system is definitely not behaving at any point where a field it set is invalid.

internal/filter/filter.go
```diff
@@ -18,18 +18,60 @@
...
+			}
+		}
+
+		if !fc.CreatedBefore.IsZero() {
+			if !ct.Before(fc.CreatedBefore) {
+				return false
+			}
+		}
+	} else {
+		logger.WithFields(logrus.Fields{
```
Laremere: if we keep this logging, the ticket id should be logged.

internal/filter/filter.go
Laremere: Ok, possible course of action (unsure):
Create another type in this package that is created from a ticket proto.  No visible fields, but it's passed to the In/Meets method.  Maybe it can return the ticket that it was created from.

Then in query cache, use this object in the map instead of tickets.

If the ticket create time parsing fails, I think a reasonable thing to do is loudly log (but this will only happen when query cache reads the ticket, instead of every query), and then assume the ticket is one minute old.

internal/statestore/redis.go
sawagh: Not sure what I was thinking there! :) Too late in the night may be  - Did not even add it to create :)! Thanks for pointing out. Fixed and added validation  (new case to the test you wrote)

internal/filter/testcases/testcases.go
```diff
@@ -106,12 +110,58 @@ func IncludedTestCases() []TestCase {
...
+			"No time search criteria positive",
+			&pb.Ticket{
+				CreateTime: timestamp(now.Add(time.Second * -2)),
+			},
+			&pb.Pool{},
+		},
 	}
 }
 
+func timestamp(t time.Time) *tspb.Timestamp {
```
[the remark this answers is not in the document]
sawagh: Moved to bottom.

internal/filter/filter.go
```diff
@@ -18,18 +18,60 @@
...
+	logger = logrus.WithFields(logrus.Fields{
+		"app":       "openmatch",
+		"component": "filter",
+	})
+)
+
+// Criteria specifies all the criteria that the Ticket needs to meet to
+// be selected. This is used instead of the Pool directly to avoid any
+// criteria validation logic in the filter package.
+type Criteria struct {
```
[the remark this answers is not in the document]
sawagh: I like your suggestion. Implemented as suggested. Added unit tests (additional to validate pool behavior, plus pool for the E2E test cases gets validated)

internal/filter/filter.go
sawagh: For your logging comment, I ve changed the log to Error. 

I see your concern around Ticket parsing. I did go down the path of exploring a type for the Ticket and updating query cache with it. Now, instead of parsing time on each query, you would end up re-creating the ticket for each query. Was not great. The alternative was for that custom type to just have (time.time and Ticket), store and use the entire ticket as is and only parse the time - so query cache still uses the custom type - but internally the type just stores Ticket+time. Seems strange again.

I've added a condition now so that the time gets parsed only if using before / after  - so slight improvement over the previous. The time parsing btw, is just another proto conversion - I dont think it would impact performance but again, I ve not validated that. With the condition though, atleast the impact is limited to only using the time range feature.

I would prefer to keep this change as is - and if we want to explore optimizing the time parsing, doing that as a separate change - given that is an internal only optimization.

internal/filter/filter.go
sawagh: Done.

internal/filter/testcases/testcases.go
```diff
@@ -259,6 +309,45 @@ func ExcludedTestCases() []TestCase {
 			},
 		},
 
+		{
```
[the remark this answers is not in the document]
sawagh: I've followed your suggestion. I do think though that we should rethink the structure here. What you call local and e2e - are actually testing two different components. One is a unit test for filter package, the other is an e2e for the query service. The inputs for these (though coincidentally were identical) need not be same (as now with the change). To test the filter package, it is fine (infact required) to create a Ticket with timestamp - whereas for the E2E, you want the auto-timestamp behavior. 

Thus, even with the suggestion, I have to in filter_test set the time to now() before creating the ticket. I do not want to take further refactoring in this PR - however, I may propose these two sharing the cases may be an overkill.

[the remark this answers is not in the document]
sawagh: I've followed your suggestion. I do think though that we should rethink the structure here. What you call local and e2e - are actually testing two different components. One is a unit test for filter package, the other is an e2e for the query service. The inputs for these (though coincidentally were identical) need not be same (as now with the change). To test the filter package, it is fine (infact required) to create a Ticket with timestamp - whereas for the E2E, you want the auto-timestamp behavior. 

Thus, even with the suggestion, I have to in filter_test set the time to now() before creating the ticket. I do not want to take further refactoring in this PR - however, I may propose these two sharing the cases may be an overkill.

internal/filter/filter_test.go
```diff
@@ -33,9 +44,47 @@ func TestInPool(t *testing.T) {
...
+			if err != nil {
+				t.Error("pool should be valid")
+			}
+			tc.Ticket.CreateTime = ptypes.TimestampNow()
+			if pf.In(tc.Ticket) {
 				t.Error("ticket should be excluded from the pool")
 			}
 		})
 	}
+
```
Laremere: This should be a separate test method.

sawagh:
@googlebot rescan

PR #1167, opened 2020-03-24
Have evaluator client and synchronizer return error when observing invalid match IDs

### LINKED ISSUE #1079: Error handling around match ids collisions and missing values
There are two error cases which should be handled better:
- If two match functions (or even one match function) return multiple matches with the same id, it should be an error.
- If the evaluator returns a match id which does not correspond to a match sent to it, it should also be an error.

Currently these are ignored and or logged.

See also https://github.com/googleforgames/open-match/issues/990

yfei1:
Resolve #1079

This commit:
1. Changed the synchronizer service to keep a record of all input match IDs per synchronization cycle, and return error if there are any proposals with duplicate match IDs in it.
2. Changed the evaluator client to keep a record of all input match IDs, and return error if it found any unexist/duplicate IDs from the output.


internal/app/synchronizer/evaluator_client.go
```diff
@@ -145,6 +147,10 @@ func (ec *grcpEvaluatorClient) evaluate(ctx context.Context, pc <-chan []*pb.Mat
 			if err != nil {
 				return fmt.Errorf("failed to get response from evaluator client, desc: %w", err)
 			}
+			if !matchIDs[resp.GetMatchId()] {
```
Laremere: We can tell whether it was returned twice, or is an invalid id, the error messages should be distinct.

internal/app/synchronizer/evaluator_client.go
```diff
@@ -145,6 +147,10 @@ func (ec *grcpEvaluatorClient) evaluate(ctx context.Context, pc <-chan []*pb.Mat
...
+				return fmt.Errorf("evaluator returned unmatched/duplicated matchID %s which does not correspond to its input", resp.GetMatchId())
+			}
+			matchIDs[resp.GetMatchId()] = false
```
Laremere: We should have e2e tests for this.

Laremere: Though we don't have a great way to set up a custom evaluator, hrm.

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -110,6 +111,13 @@ func (s *synchronizerService) Synchronize(stream ipb.Synchronizer_SynchronizeSer
 				registration.allM1cSent.Done()
 				return
 			}
+			if _, ok := matchIDs[req.GetProposal().GetMatchId()]; ok {
```
Laremere: This has a couple problems:

It's not detecting conflicts from multiple FetchMatches calls.  This synchronize method only sees one.

We most definitely needs to return an error here, instead of just logging.

internal/app/synchronizer/evaluator_client.go
yfei1: I'll mark this as a TODO.

Laremere: Use issues, not code todos.

internal/app/backend/backend_service.go
```diff
@@ -117,22 +117,40 @@ func (s *backendService) FetchMatches(req *pb.FetchMatchesRequest, stream pb.Bac
 	return nil
 }
 
-func synchronizeSend(ctx context.Context, syncStream synchronizerStream, m *sync.Map, proposals <-chan *pb.Match) error {
-sendProposals:
+func synchronizeSend(ctx context.Context, syncStream synchronizerStream, m *sync.Map, pc <-chan *pb.Match) error {
```
Laremere: Doing this adds buffering until the mmf is finished, which isn't desired.

You could just use loadandstore, returning an error if the value already exists (everything will clean up once an error is return which cancels the context.

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -91,67 +93,85 @@ func (s *synchronizerService) Synchronize(stream ipb.Synchronizer_SynchronizeSer
...
+				return fmt.Errorf("error streaming in synchronizer from backend, desc: %w", err)
 			}
-			registration.m1c.send(mAndM6c{m: req.Proposal, m7c: registration.m7c})
+			if _, ok := matches[req.GetProposal().GetMatchId()]; ok {
+				duplicateIDs = append(duplicateIDs, req.GetProposal().GetMatchId())
+			}
+			matches[req.GetProposal().GetMatchId()] = req.GetProposal()
+		}
+
+		if len(duplicateIDs) != 0 {
```
Laremere: The synchronize method only has context from one mmf run... this isn't checking different mmf calls....

See comment on evaluator client for alternative.

internal/app/synchronizer/evaluator_client.go
```diff
@@ -119,10 +119,12 @@ func (ec *grcpEvaluatorClient) evaluate(ctx context.Context, pc <-chan []*pb.Mat
 	}
 
 	results := []string{}
+	matchIDs := map[string]bool{}
 
 	eg.Go(func() error {
 		for proposals := range pc {
 			for _, proposal := range proposals {
+				matchIDs[proposal.GetMatchId()] = true
```
Laremere: When checking for duplicate ids within the synchronizer, I think this is a reasonable place to do it:
- You have a set of previously sent ids.
- You're already checking for duplication.

Just check here if it's already been seen, and return an error if it has?

internal/app/backend/backend_service.go
yfei1: But this won't return all of the duplicated IDs though.

Laremere: That doesn't matter.

Duplicate ids indicate a programming error.  We just need to fail loudly (because the evaluator won't work anyways) so that they know they're using OM wrong.  Having 1 duplicate id or a million doesn't change that they need to fix their programming.

internal/app/synchronizer/evaluator_client.go
```diff
@@ -119,10 +119,15 @@ func (ec *grcpEvaluatorClient) evaluate(ctx context.Context, pc <-chan []*pb.Mat
 	}
 
 	results := []string{}
+	matchIDs := map[string]bool{}
 
 	eg.Go(func() error {
 		for proposals := range pc {
 			for _, proposal := range proposals {
+				if matchIDs[proposal.GetMatchId()] {
+					return fmt.Errorf("found duplicat matchID %s", proposal.GetMatchId())
```
Laremere: "duplicate"

internal/app/synchronizer/evaluator_client.go
```diff
@@ -119,10 +119,15 @@ func (ec *grcpEvaluatorClient) evaluate(ctx context.Context, pc <-chan []*pb.Mat
 	}
 
 	results := []string{}
+	matchIDs := map[string]bool{}
 
 	eg.Go(func() error {
 		for proposals := range pc {
 			for _, proposal := range proposals {
+				if matchIDs[proposal.GetMatchId()] {
```
Laremere: `_, ok := matchIDs[proposal.GetMatchId()]; ok`

It's possible that it could have already been returned by the evaluator

internal/app/synchronizer/evaluator_client.go
```diff
@@ -119,10 +119,15 @@ func (ec *grcpEvaluatorClient) evaluate(ctx context.Context, pc <-chan []*pb.Mat
 	}
 
 	results := []string{}
+	matchIDs := map[string]bool{}
```
Laremere: actually, I just noticed this is being used by multiple go routines at the same time.  Needs to be a sync.map

Issue #1155, opened 2020-03-11
Modify AssignTickets API to support Bulk and Optimistic assignments. 


sawagh:
**Overview**

This proposal combines motivations and proposals from the following issues:
https://github.com/googleforgames/open-match/issues/1121 
https://github.com/googleforgames/open-match/issues/1112

Please see those issues for details of motivations for these changes. Here is an overview of the changes needed:

1. Provide an ability to set multiple assignments in the same AssignTickets call
2. Do not fail the AssignTickets call for a single assignment failure, but rather provide this as status information. 

Proposal:

This issue converges proposals in both the linked issues. A minor change though is that in #1112, each assignment is to a single Ticket - which actually will make the originally supported scenario (multiple tickets to one assignment) very verbose. Hence, this proposal enables assigning an assignment to one or more ticket and supports multiple such assignment units in the same TicketAssignment call. 

```
message TicketAssignment{
  // TicketIds is a list of strings representing Open Match generated Ids which apply to an Assignment.
  repeated string ticket_id = 1;

  // An Assignment specifies game connection related information to be associated with the TicketIds.
  Assignment assignment = 2;
}

// AssignmentFailure contains information about the Ticket and the Assignment that failed and the failure status.
message AssignmentFailure {
  string ticket_id = 1;
  Assignment assignment = 2;
  google.rpc.Status error = 2;
}

message AssignTicketsRequest {
  // Assignments is a map of TicketIds to Assignments for updating the associated ticket with the assignment value.
  repeated TicketAssignment assignments = 1;
}

message AssignTicketsResponse {
  // Failed is a list of all the Assignments that failed, along with failure status. 
  repeated AssignmentFailure failed = 1;
}

rpc AssignTickets(AssignTicketsRequest) returns (AssignTicketsResponse) {
  option (google.api.http) = {
    post: "/v1/backendservice/tickets:assign"
    body: "*"
  };
}
```

Laremere:
Mostly looks good.

Naming wise, everything looks good but `TicketAssignment` is a little unintuitive for having multiple tickets.  Consider `TicketsAssignment`, or `TicketAssignmentGroup`, or maybe just leaving what you have, idk.

For functionality, I am wondering about two things:
What is the purpose of `AssignmentFailure.assginment`?  The client sent this assignment, so why should they get it back?  What info does it have?

For `AssignmentFailure.error`, I'm not sure the rpc.Status makes sense?  It's not clear to me how to interpret that.  I would say consider:

[...]

PR #1170, opened 2020-03-26
Implement assignment groups and graceful failure

Laremere:
Resolves https://github.com/googleforgames/open-match/issues/1155

Rewrote the tests to mostly be not parameterized, since they are testing different scenarios, I find straightforward tests to be simple and easier to understand.


internal/statestore/public.go
```diff
@@ -50,7 +50,7 @@ type Service interface {
 	GetTickets(ctx context.Context, ids []string) ([]*pb.Ticket, error)
 
 	// UpdateAssignments update the match assignments for the input ticket ids
-	UpdateAssignments(ctx context.Context, ids []string, assignment *pb.Assignment) error
+	UpdateAssignments(ctx context.Context, req *pb.AssignTicketsRequest) (*pb.AssignTicketsResponse, error)
```
yfei1: can you update the outdated comments?

internal/statestore/redis.go
```diff
@@ -476,78 +476,83 @@ func (rb *redisBackend) GetTickets(ctx context.Context, ids []string) ([]*pb.Tic
...
+				TicketId: ids[i],
+				Cause:    pb.AssignmentFailure_TICKET_NOT_FOUND,
+			})
+		} else {
+			t := &pb.Ticket{}
+			err = proto.Unmarshal(ticketByte, t)
+			if err != nil {
+				redisLogger.WithFields(logrus.Fields{
+					"key": ids[i],
+				}).WithError(err).Error("Failed to unmarshal ticket from redis.")
```
yfei1: lowercase

internal/statestore/redis.go
```diff
@@ -476,78 +476,83 @@ func (rb *redisBackend) GetTickets(ctx context.Context, ids []string) ([]*pb.Tic
...
-		}
-		t := &pb.Ticket{}
-		err = proto.Unmarshal(ticketByte, t)
-		if err != nil {
-			redisLogger.WithFields(logrus.Fields{
-				"key": ids[i],
-			}).WithError(err).Error("Failed to unmarshal ticket from redis.")
-			return status.Errorf(codes.Internal, "%v", err)
+			resp.Failed = append(resp.Failed, &pb.AssignmentFailure{
+				TicketId: ids[i],
```
yfei1: `idsI[i]` works at here. do we really need the `ids` slice?

Laremere: It's also used to create an AssignmentFailure, which needs a string.

yfei1: is type assertion not working?

Laremere: I'd rather not use a type assertion if I don't have to.

PR #1181, opened 2020-04-01
Removed make all commands and pinned dependency versions

yfei1:
Per discussion, this commit:

1. Pinned third_party code versions to avoid having presubmit updating files randomly. The google apis version is set to a most recent commit hash since this repo doesn't do regular release.
2. Removed `make all` commands from the Makefile since it isn't being used in the development workflow for a while.


yfei1:
> Remove from developer guide?

I created a separate PR for the development guide since it needs more updates than removing the `make all` command

PR #1184, opened 2020-04-09
Unified images used in helm chart and release artifacts

yfei1:
Open Match had a long-exist issue that our published helm chart use images in `gcr.io/open-match-public-images` bucket while our release assets use the ones under `gcr.io/open-match-build` bucket. This works since both buckets exist, yet makes validating and debugging releases painful. 

This commit added an environmental variable check to the `make install/yaml/` rule such that cloud build could generate assets using the public-images bucket when running against a release branch.


Laremere:
Can we instead for the release instructions have overrides on the make command?

yfei1:
> Can we instead for the release instructions have overrides on the make command?

I gave it some more thought and seems like simply overrides the make command wouldn't work.

The yaml artifacts are deployed to our gcs bucket through our CI in the `ci-deploy-artifacts` step, so it is unrelated to our release process and it automatic.

Laremere:
Ok, I see the problem.  Will think more about it - my concern is that this PR is adding more (untested) magic to the makefile, when we should be trying to reduce that.  Maybe some other aspects of the system / processes need to align to solve this in a different way, I don't know.

PR #1185, opened 2020-04-10
Use consistent main code for mmf and evaluator

Laremere:
This brings the test mmf and standard evaluator closer to the other commands.  So now they read the config instead of hardcoding it, and they use a bind function (though it's customized with the actual mmf or evaluator code.)

Also a small extra empty line to separate out license because it's not godoc.


yfei1:
lint is unhappy

PR #1191, opened 2020-04-16
Moved customized configmap values to default

yfei1:
This commit moved configs under the customized-configmap to default and let the evaluator and mmf use the default-configmap only. This allows the evaluator and mmf to inherit the same telemetry settings as the other core components.


Laremere:
Per discussion: It seems like we should just get rid of customize chart's configmaps, and use core's for everything there?

Issue #1124, opened 2020-02-19
Add metrics in Open Match concept


yfei1:
This is the last bit of Open Match's monitoring story before v1.0.

Right now, Open Match provides dashboards for monitoring k8s pod usage, grpc-level latency and error rate, and Redis performance, but lack of a story of monitoring Open Match related concepts. I plan to add a new dashboard to track the following metrics:

Ticket related metrics:
- [x] Ticket size and ticket extension size.
- [x] Number of SearchFields per ticket
- [ ] Time that a ticket spent & number of tickets in each match making stage - created, addedtoignorelist, deindexed, deleted.

Match related metrics:
- [x] Match size and Match extension size.
- [x] Number of tickets per match.

QueryService related metrics:
- [x] Number of tickets per QueryTickets call returned.
- [x] (the step after query cache) metrics related to the performance of query cache.

Evaluator related metrics:
- [x] Number of matches sent to the evaluator
- [x] Number of matches returned from the evaluator
- [x] (optional) number of collided matches using the default evaluator - the evaluator is a replaceable component so this metric applies only if one is using the default evaluator.

Synchronizer related metrics (1-3 copied from #1104 )
- [x] A metric for number of iterations started. - @Laremere I'm not sure what is the value-added for this metric, can you explain?
- [x] Time waiting for registration: indicates reg window too small
- [x] Time wasted in reg window with done mmfs: indicates reg window too big
- [x] Number of registered call per synchronization cycle

Laremere:
> A metric for number of iterations started. - @Laremere I'm not sure what is the value-added for this metric, can you explain?

This is just how many synchronizer iterations we have.  Currently is on the scale tests.  It might actually be more useful to have this divided up into buckets for how long each iteration took.

When this goes low, you're cycles are taking too long and you're unhealthy.

sawagh:
What milestone are we targeting this for?

yfei1:
> What milestone are we targeting this for?

This is a must-have for 1.0 and I'm working on it this week when I'm running out of tasks. Aiming to close this in v0.10.

Laremere:
Per comments in #1135, it'd be nice to have metrics on time spent updating the query service.

yfei1:
Moving this issue to v1.0. Almost done with the metrics part yet I'm not sure if there is anything that needs to be further changed when setting up the grafana dashboards.

PR #1192, opened 2020-04-16
Refactor internal telemetry package

yfei1:
This commit refactored the internal telemetry package. The pattern used in `internal/app/xxx/xxx.go` follows the one used in [openconcensus-go](https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ocgrpc/client_metrics.go#L25). Besides adding metrics covered in #1124, this commit also introduced changes to make the telemetry settings more efficient and easier to turn on/off.

1. In this factorization, a metric recorded can be cast into different views through different aggregation methods. Since the metric is the one that consumes most of the resources, this can make the telemetry setups more efficient than before.
2. Also removed some metrics that were meaningful for debugging in v0.8 but are becoming useless for the current stage.


yfei1:
Per discussion, hold off this PR until Scott's binding PR s in

Laremere:
Binding PR is in.

yfei1:
Ready for review.

internal/app/backend/backend.go
```diff
@@ -34,6 +75,14 @@ func BindService(p *appmain.Params, b *appmain.Bindings) error {
 	b.AddHandleFunc(func(s *grpc.Server) {
 		pb.RegisterBackendServiceServer(s, service)
 	}, pb.RegisterBackendServiceHandlerFromEndpoint)
-
+	if err := view.Register(
```
Laremere, line 78: This should be returning the error instead of logging fatal, and addding a closer
for unregistering.  Since this is a common pattern across BindService functions,
it can be generalized in appmain.

The BindService function doesn't really care about the view error either: It won't
add any context to it, and it should fail app startup.  So we can additionally
handle that by just setting on the side, and noticing it later.

```
// Bindings allows applications to bind various functions to the running servers.
type Bindings struct {
	sp *rpc.ServerParams
	a  *App
	firstErr error
}
```


```
func (b *Bindings) RegisterView(v *view.View) {
	err := view.Register(v)
	if err != nil {
	    if b.firstError == nil {
	      b.firstError = err
	    }
	    return
	}
	b.AddCloser(func() {
	  view.Unregister(v)
	})
}
```

in NewApplication
```
	err = bindService(p, b)
	if err != nil {
		surpressedErr := a.Stop() // Don't care about additional errors stopping.
		_ = surpressedErr
		return nil, err
	}

    if b.firstErr != nil {
		surpressedErr := a.Stop() // Don't care about additional errors stopping.
		_ = surpressedErr
		return nil, b.firstErr
    }
```

cmd/default-evaluator/main.go
```diff
@@ -21,5 +21,5 @@ import (
 )
 
 func main() {
-	appmain.RunApplication("evaluator", evaluator.BindServiceFor(defaulteval.Evaluate))
+	appmain.RunApplication("evaluator", evaluator.BindServiceFor(evaluator.Evaluator{EvaluateFunc: defaulteval.Evaluate, Binders: defaulteval.Binders}))
```
Laremere, line 24: This pattern is awkward for a couple reasons:
- Passing the arguments inside a single a struct, instead of just using multiple args.
- The function and "binders" are always used together, so why doesn't that package do it?

Also, the change to evaluator breaks my testing change, I think that it's
reasonable to have a type for the function itself.

I think a cleaner pattern would be to invert how these are being used: instead
of the default evaluator being passed to the BindServiceFor, have the default
evaluator use BindServiceFor.

Within the defaulteval package, rename Evaluate to evaluate.  Add a BindService
function, which calls evaluator.BindServiceFor(evaluate)(p, b).  In addition, it
can bind any metrics. Then have this main function use that BindFunction instead
of BindServiceFor.

internal/appmain/appmain.go
```diff
@@ -89,6 +92,18 @@ func (b *Bindings) AddHealthCheckFunc(f func(context.Context) error) {
 	b.sp.AddHealthCheckFunc(f)
 }
 
+// RegisterViews begins collecting data for the given views.
+func (b *Bindings) RegisterViews(v ...*view.View) {
+	if err := view.Register(v...); err != nil {
+		b.firstErr = err
```
Laremere, line 98: This will set firstErr, even if it's not the first error.

internal/app/evaluator/defaulteval/evaluator.go
```diff
@@ -30,16 +35,35 @@ var (
...
+	}
 )
 
 type matchInp struct {
 	match *pb.Match
 	inp   *pb.DefaultEvaluationCriteria
 }
 
+// BinderServices define the initialization steps for this evaluator
+func BinderServices() appmain.Bind {
```
Laremere, line 54: Like the other apps, no need for a lambda, just declare the function as:
```
func BindService(p *appmain.Params, b *appmain.Bindings) error {
```
Then reference with `evaluator.BindService`, instead of `evaluator.BindServices()`

internal/app/backend/backend_service.go
```diff
@@ -168,9 +165,14 @@ func synchronizeRecv(ctx context.Context, syncStream synchronizerStream, m *sync
 			cancelMmfs()
 		}
 
-		if match, ok := m.Load(resp.GetMatchId()); ok {
-			telemetry.RecordUnitMeasurement(ctx, mMatchesFetched)
-			err = stream.Send(&pb.FetchMatchesResponse{Match: match.(*pb.Match)})
+		if v, ok := m.Load(resp.GetMatchId()); ok {
+			match, ok := v.(*pb.Match)
```
Laremere, line 169: I don't think this type check is necessary.  It will always be a match because that's the only thing put into the map.  It's reasonable to expect tests to cover this kind of stuff.

internal/app/frontend/frontend.go
```diff
@@ -15,12 +15,33 @@
...
 	"open-match.dev/open-match/pkg/pb"
 )
 
+var (
+	totalBytesPerTicket   = stats.Int64("openmatch.dev/total_bytes_per_ticket", "Total bytes per ticket", stats.UnitBytes)
+	searchFieldsPerTicket = stats.Int64("openmatch.dev/searchfields_per_ticket", "Searchfields per ticket", stats.UnitDimensionless)
+
+	totalBytesPerTicketView = &view.View{
+		Measure:     totalBytesPerTicket,
+		Name:        "openmatch.dev/total_bytes_per_ticket",
```
Laremere, line 33: Some of these have the service in the name, some don't.  What's the difference?

Laremere, line 33: Why openmatch.dev, and not open-match.dev?

internal/app/query/query.go
```diff
@@ -15,11 +15,59 @@
...
+		Measure:     cacheTotalItems,
+		Name:        "openmatch.dev/query/total_cache_items",
+		Description: "Total number of tickets query service cached",
+		Aggregation: view.LastValue(),
+	}
+	cacheDeltaItemsView = &view.View{
+		Measure:     cacheDeltaItems,
+		Name:        "openmatch.dev/query/delta_cache_items",
+		Description: "Number of tickets cache delta items",
+		Aggregation: view.LastValue(),
```
Laremere, line 49: Delta can be calculated in prometheus with a query.  It's also not likely to be useful, maybe there's big updates, but most of the time I only do a small update, so this metric would look too low.  However it would be useful to know the rate of change, instead of the last value.  So maybe include that metric instead?

internal/app/query/query.go
```diff
@@ -15,11 +15,59 @@
...
+		Measure:     cacheWaitingQueries,
+		Name:        "openmatch.dev/query/cache_updates",
+		Description: "Number of query cache updates in total",
+		Aggregation: view.Count(),
+	}
+	cacheWaitingQueriesView = &view.View{
+		Measure:     cacheWaitingQueries,
+		Name:        "openmatch.dev/query/waiting_requests",
+		Description: "Number of waiting requests in total",
+		Aggregation: telemetry.DefaultMillisecondsDistribution,
```
Laremere, line 61: Should this be a count distribution, not millisecond?

internal/app/frontend/frontend.go
yfei1: I glanced through a bunch of examples but rarely see using `-` in the view name so I chose not to. I don't have a strong opinion here as long as the naming is consistent.

internal/app/query/query.go
yfei1: Removed this metric.
rate of query can also be calculated via a query and is included in the next PR.

internal/app/query/query.go
yfei1: good catch!

internal/app/backend/backend_service.go
yfei1: lint blocked me from any types of direct casting
```
internal/app/backend/backend_service.go:169:11: Error return value is not checked (errcheck)
                        match, _ := v.(*pb.Match)
```

internal/app/evaluator/defaulteval/evaluator.go
```diff
@@ -51,14 +51,12 @@ type matchInp struct {
...
 
 // BinderServices define the initialization steps for this evaluator
-func BinderServices() appmain.Bind {
-	return func(p *appmain.Params, b *appmain.Bindings) error {
-		if err := evaluator.BindServiceFor(evaluate)(p, b); err != nil {
-			return err
-		}
-		b.RegisterViews(collidedMatchesPerEvaluateView)
-		return nil
+func BinderServices(p *appmain.Params, b *appmain.Bindings) error {
```
Laremere, line 54: BindService not BinderServices

internal/app/backend/backend_service.go
Laremere: just do `match := v.(*pb.match)` ?

internal/appmain/appmain.go
```diff
@@ -89,6 +92,18 @@ func (b *Bindings) AddHealthCheckFunc(f func(context.Context) error) {
 	b.sp.AddHealthCheckFunc(f)
 }
 
+// RegisterViews begins collecting data for the given views.
+func (b *Bindings) RegisterViews(v ...*view.View) {
+	if err := view.Register(v...); err != nil && b.firstErr == nil {
```
Laremere, line 97: This will add an invalid unregister even if there's an error if firstErr has been set.

internal/telemetry/metrics.go
```diff
@@ -22,9 +22,11 @@ import (
 	"go.opencensus.io/tag"
 )
 
+// Default histogram distributions
 var (
-	// HistogramBounds defines a unified bucket boundaries for all histogram typed time metrics in Open Match
-	HistogramBounds = []float64{0, 50, 100, 200, 400, 800, 1600, 3200, 6400, 12800, 25600, 51200}
+	DefaultBytesDistribution        = view.Distribution(64, 128, 256, 512, 1024, 2048, 4096, 16384, 65536, 262144, 1048576)
+	DefaultMillisecondsDistribution = view.Distribution(0.01, 0.05, 0.1, 0.3, 0.6, 0.8, 1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400, 500, 650, 800, 1000, 2000, 5000, 10000, 20000, 50000, 100000)
+	DefaultCountDistribution        = view.Distribution(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536)
```
Laremere, line 29: Should this have a value for 0?

internal/app/backend/backend_service.go
yfei1: still doesn't work.. lint is smart :(

internal/appmain/appmain.go
yfei1: 🤦

internal/telemetry/metrics.go
yfei1: I'm following the usage at here https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ocgrpc/stats_common.go#L57

internal/app/frontend/frontend.go
Laremere: open-match.dev is more consistently used, so I'd prefer that.

internal/app/query/query.go
Laremere: It's not the rate of query calls which is interesting here, it's the rate of tickets being added to the cache.

cmd/default-evaluator/main.go
```diff
@@ -15,11 +15,10 @@
...
 
 import (
-	"open-match.dev/open-match/internal/app/evaluator"
 	"open-match.dev/open-match/internal/app/evaluator/defaulteval"
 	"open-match.dev/open-match/internal/appmain"
 )
 
 func main() {
-	appmain.RunApplication("evaluator", evaluator.BindServiceFor(defaulteval.Evaluate))
+	appmain.RunApplication("evaluator", defaulteval.BinderService)
```
Laremere, line 23: BindService, not BinderService

internal/app/query/query_service.go
```diff
@@ -316,6 +316,7 @@ func (tc *ticketCache) update() {
 
 	stats.Record(context.Background(), cacheTotalItems.M(int64(previousCount)))
 	stats.Record(context.Background(), cacheUpdateLatency.M(float64(time.Since(st))/float64(time.Millisecond)))
+	stats.Record(context.Background(), cacheUpdateLatency.M(len(toFetch)))
```
Laremere: cacheFetchedItems

internal/testing/e2e/in_memory.go
```diff
@@ -86,6 +87,6 @@ func start(t *testing.T, eval evaluator.Evaluator, mmf mmfService.MatchFunction)
 	cfg.Set("logging.level", *testOnlyLoggingLevel)
 	cfg.Set(telemetry.ConfigNameEnableMetrics, *testOnlyEnableMetrics)
 
-	apptest.TestApp(t, cfg, listeners, minimatch.BindService, mmfService.BindServiceFor(mmf), evaluator.BindServiceFor(eval))
+	apptest.TestApp(t, cfg, listeners, minimatch.BindService, mmfService.BindServiceFor(mmf), defaulteval.BindService)
```
Laremere, line 90: This line shouldn't be changing, for here or in cluster.  The default evaluator is no longer used for the e2e tests.

yfei1: That makes sense, thanks!

Laremere: cluster too

PR #1196, opened 2020-04-22
Have the test instance host the mmf and evaluator

Laremere:
This sets things up for specifying a per-test evaluator and mmf.

### DIFF
--- a/internal/testing/e2e/in_memory.go
+++ b/internal/testing/e2e/in_memory.go
@@ -69,16 +69,16 @@ func (iom *inmemoryOM) MustQueryServiceGRPC() pb.QueryServiceClient {
 
 func (iom *inmemoryOM) MustMmfConfigGRPC() *pb.FunctionConfig {
 	return &pb.FunctionConfig{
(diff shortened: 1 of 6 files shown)


internal/testing/e2e/in_memory.go
```diff
@@ -190,9 +190,9 @@ api:
   scale:
     httpport: "51509"
   evaluator:
-    hostname: "om-evaluator"
-    grpcport: "50508"
-    httpport: "51508"
+    hostname: "test"
```
yfei1, line 193: I think you overwrite the evaluator values at L138 so changing the values at here wont take any effects

Laremere: I want this to match exactly that the real cluster's config looks like, will add test which ensures that in a future PR.  (it's still wrong, but I want changing it there, so I wanted to change it here.)

yfei1: idk.. but having the in cluster config being override in the Makefile and the in memory one being set in a go file looks pretty weird @_@

Issue #1203, opened 2020-04-29
rename ignoreListTTL


Laremere:
It should use the newer "pending" terminology.

PR #1206, opened 2020-05-06
Rework config value naming

Laremere:
resolves https://github.com/googleforgames/open-match/issues/1203

ignoreListTTL -> pendingReleaseTimeout
This uses the new terminology for ticket state, and not an implementation specific name.

pageSize/page.size -> queryPageSize
Size was the only variable within page object, so it made little sense.

Dropped the "Ms" reverse hungarian notation on a few of the timing variables (that didn't need to be expressed in MS anyways!)

Moved all of the variables to the top level.  "storage" was a weird location for pageSize and pendingReleaseTimeout (none of the logic goes into redis...), and the synchronizer is an implementation details of the evaluation/synchronization cycle.  

Also added documentation in both the values.yaml and the configmap.


internal/statestore/testing/consts.go
```diff
@@ -25,8 +25,8 @@ const (
 	PoolIdleTimeout = 10 * time.Second
 	// PoolHealthCheckTimeout is the read/write timeout of a healthcheck HTTP request
 	PoolHealthCheckTimeout = 100 * time.Millisecond
-	// IgnoreListTTL is the time to live duration of Open Match ignore list settings
-	IgnoreListTTL = 500 * time.Millisecond
+	// pendingReleaseTimeout is the time to live duration of Open Match ignore list settings
+	pendingReleaseTimeout = 500 * time.Millisecond
```
yfei1, line 29: Would it be 200?

Laremere: This is for a different set of tests.  They don't need to match.

Issue #1168, opened 2020-03-25
Proposal: Configurable time-to-live on assigned tickets


calebatwd:
Tickets backup if client behavior is unable to consistently and reliably remove tickets that are "completed."

Please add a configurable ttl on assigned tickets that will automatically remove tickets that are completed with assignments.

Default value recommendation: 10 minutes?

PR #1211, opened 2020-05-08
Give assigned tickets a time to live, default 10 minutes

Laremere:
Resolves https://github.com/googleforgames/open-match/issues/1168


install/helm/open-match/templates/om-configmap-override.yaml
```diff
@@ -33,4 +33,5 @@ data:
     synchronizer:
       registrationIntervalMs:  {{ index .Values "open-match-core" "registrationIntervalMs" }}
       proposalCollectionIntervalMs: {{ index .Values "open-match-core" "proposalCollectionIntervalMs" }}
+    assignedDeleteTimeout: {{ index .Values "open-match-core" "proposalCollectionIntervalMs" }}
```
sawagh, line 36: you mean assignedDeleteTimeout?

install/helm/open-match/values-production.yaml
```diff
@@ -173,6 +173,7 @@ redis:
 open-match-core:
   enabled: true
   ignoreListTTL: 60000ms
```
sawagh, line 175: may be represent everything in excess of a minute as a minutes value across the yamls? (like you did for this variable in the values.yaml file?

internal/statestore/redis.go
```diff
@@ -529,13 +534,36 @@ func (rb *redisBackend) UpdateAssignments(ctx context.Context, req *pb.AssignTic
...
 		}
 
-		cmds = append(cmds, ticket.GetId(), ticketByte)
+		err = redisConn.Send("SET", ticket.Id, ticketByte, "PX", int64(assignmentTimeout), "XX")
+		if err != nil {
+			return nil, errors.Wrap(err, "error sending ticket assignment set")
+		}
 	}
 
-	_, err = redisConn.Do("MSET", cmds...)
```
sawagh, line 535: Wasnt moving to MSET an optimization? Why move back to EXEC? and if there are reasons, then may be do that consistently for other patterns in the storage layer?

https://github.com/googleforgames/open-match/commit/3de052279b5a725c2980851a29056b8a45ead401

Laremere: Redis does not provide a function which both sets multiple values, and also allows setting the expiration at the same time.  As far as I can see there's only setting expiration one ticket at a time.  So it's fewer commands to set them with their expiration.  As a bonus, it's possible to say only set they key if it exists, which prevents a race condition if the ticket is deleted between reading and writing.

install/helm/open-match/templates/om-configmap-override.yaml
Laremere: great catch!

install/helm/open-match/values-production.yaml
Laremere: Yeah, they should be in sync. Fixed.

Issue #1159, opened 2020-03-17
Proposal: Release all tickets to pool


Laremere:
In the event that the director crashes, it know that any pending tickets are currently reserved for matches which will never receive an assignment.  There is a timeout, but it can speed things along by indicating to open match to immediately return tickets to the pool.

This would additionally be helpful for the frequently requested "Delete all tickets" operation.  Currently we don't have any direct method to do it, and recommend querying all tickets and then deleting manually.  This would exclude any tickets which are currently in a pending state.

```protobuf
message ReleaseTicketsRequest{
  // TicketIds is a list of string representing Open Match generated Ids to be re-enabled for MMF querying
  // because they are no longer awaiting assignment from a previous match result
  repeated string ticket_ids = 1;

  // If true, ticket_ids is ignored, and instead all tickets which are pending are returned to the
  // active pool.
  bool release_all_tickets = 2;
}
```

Implementation wise, this is simply a `DEL proposed_ticket_ids` to remove the key of the sorted set.

sawagh:
Having release_all_tickets and repeated string ticket_ids feels strange. Even though documented that specifying one overrules the other, that is not obvious / intuitive. (something to the effect of oneof is probably clear but slightly more complicated to use)

Alternatively, why not just provide basic primitives and let users build these scenarios? Here is a proposal (achieves this and more)

```
// Just for reference, actual state details worked on in #1100
enum TicketState {
  ACTIVE = 0;
  PENDING = 1;
  ASSIGNED = 2;
}

message ListTicketIdsRequest {
  TicketState state = 1
}

message ListTicketIdsResponse {
  repeated string ids = 1;
}

service QueryService {
  rpc ListTicketIds(ListTicketIdsRequest) returns (stream ListTicketIdsResponse) {
    option (google.api.http) = {
      post: "/v1/queryservice/ticketids:query"
      body: "*"
    };
  }
}
```

If we have the above API, that can then be used with any of the release / delete / get primitives to build scenarios. Thoughts?

sawagh:
Pulling this into v0.10 for visibility - This is very much related to https://github.com/googleforgames/open-match/issues/766 which was a v0.10 issue and I do not think completely addresses the user requirement.

sawagh:
[...]

Issue #764, opened 2019-08-22
Provide ability to delete all the tickets in Open Match


sawagh:
This came up when Psyonix was analyzing some of the failure conditions. Given that Open Match expects tickets to players mapping to be maintained outside of Open Match, it is totally possible that the entity retaining this information loses it (crash?). At this point, say the game lost all information about what Tickets are present in Open Match. This would then have two side effects:
- These leaked tickets would continue to be matched
- There would be no good way to delete this state

Plenty of options:
1. Open Match Provides DeleteAllTickets() <-- Self Explanatory
2. Open Match Provides GetAllTickets() <-- returns all ticket IDs

(2) Provides some information that only Open Match knows and knowing that information may be useful in other scenarios. (1) could be built on top of (2) externally. So in the original problem, when starting up, if the game service (frontend / director) wants to get openmatch to a clean state - GetAllTickets --> followed by delete each could help.

DeleteAllTickets as a functionality may save the user the pain of getting and looping delete - my concern would be conveying the guarantees we provide when this happens - do we prevent any other activity to happen when this is in progress? Is it transnational? I prefer providing a basic GetAllTickets functionality and have the user build scenarios such as delete all using that.

rib:
A convenient way to delete all tickets would be handy during development. I've just ended up in a sitation where I've put a ticket into the system that triggers an error in my director (the ticket has an Any Extension with a type_url that Go doesn't like) and now that I've fixed the issue in my frontend it'd be good to clean out all old tickets.

It's also caught me out when updating my frontend/director that I sometimes have left-over tickets that then interfere with testing.

rib:
I ended up with this sledgehammer as a workaround for now:

```go
func deleteAllTickets() error {
	log.Println("Deleting all pre-existing tickets")

	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	mmlogicAddr := "om-mmlogic.open-match.svc.cluster.local:50503"
	conn, err := grpc.Dial(mmlogicAddr, grpc.WithInsecure())
	if err != nil {
		log.Fatalf("Failed to connect to Matchmaker Logic, got %v", err)
	}
	defer conn.Close()

	mml := pb.NewMmLogicClient(conn)

	frontendAddr := "om-frontend.open-match.svc.cluster.local:50504"
	conn, err = grpc.Dial(frontendAddr, grpc.WithInsecure())
	if err != nil {
[...]

PR #1215, opened 2020-05-08
Add release all tickets API

Laremere:
Release all tickets will move all pending tickets to being active again.  It solves two user requested use cases:

- Director has crashed, so any pending assignments are now lost.
- Deleting all tickets, used in development and in some director setups.  It is followed by a query all ids, and a deletes for those ids.

Resolves https://github.com/googleforgames/open-match/issues/1159
Resolves https://github.com/googleforgames/open-match/issues/764

Implemented as discussed in the latest sync with Unity.  Labeled as beta so we're not locked in by going 1.0


api/backend.proto
```diff
@@ -150,4 +154,17 @@ service BackendService {
...
     };
   }
+
+  // ReleaseAllTickets moves all tickets from the pending state, to the active
+  // state. This enables them to be returned by query, and find different
+  // matches.
+  // 
+  // BETA FEATURE WARNING:  This call and the associated Request and Response
+  // messages are not finalized and still subject to possible change or removal.
+  rpc ReleaseAllTickets(ReleaseAllTicketsRequest) returns (ReleaseAllTicketsResponse) {
```
yfei1, line 164: Use google.protobuf.Empty instead?

Laremere: Thought about it, but I'd prefer to stick with having dedicated request/response types.
The exception is for CRUD operations, which follows the API guide.

PR #1218, opened 2020-05-11
Release  1.0.0-rc.1

Laremere:
Not doing the go.mod for the tutorials because they would reference a non-existent tag.  They can be updated later. (and no longer  need to always match since there won't be backwards incompatible changes anymore.)


install/helm/open-match/values.yaml
```diff
@@ -257,7 +257,7 @@ global:
   # Use this field if you need to override the image registry and image tag for all services defined in this chart
   image:
     registry: gcr.io/open-match-public-images
-    tag: 0.0.0-dev
```
yfei1, line 260: Also need to values-production.yaml

Laremere: done

PR #1233, opened 2020-07-09
Fix typo in backend-service.go

griboue:
Fixed typo


Laremere:
In the future, please use more useful PR name.  "update" is rarely a useful word in PR messages, as any change is an update.  "Fix typo" or "Fix typo in backend-service.go" would both be better.

Laremere:
Thanks for spotting and taking the time to send in a fix though.

Issue #1235, opened 2020-07-17
Support installing more than one Open Match instances in the same namespace


lihaif:
**What happened**:
We want to include Open Match as a sub-chart of our Meta-game helm chart.

In our CI Kubernetes cluster, we deploy all pipelines(each pipeline corresponds to a Meta-game helm deployment) to the same "CI" namespace and run auto-test based on each corresponding Meta-game cluster. The release name of each pipeline looks like "{pipeline-id}-{hash-code}", which will be used as the prefix of "fullname" in the resources of our helm chart. And we are able to deploy multiple Meta-game instances in the same namespace.

But I find that Open Match does not support to install multiple instances in the same namespace. e.g. the name of https://github.com/googleforgames/open-match/blob/57e59c3821b3eade27570ba5ea56c1c33caaaf4a/install/helm/open-match/templates/podsecuritypolicy.yaml#L21 is hard coded. We can not change it easily.

Another issue is that the "Release Name" is not part of the Open Match resource name, which makes it difficult to visually identify which release it belongs to if there are multiple releases in the same namespace.

**What you expected to happen**:
One possible solution is to include the "Release Name" as part of the resource name. For example. below is the code for MongoDB helm chart to get the "mongodb.fullname".

```
{{- define "mongodb.fullname" -}}
{{- if .Values.fullnameOverride -}}
{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- $name := default .Chart.Name .Values.nameOverride -}}
{{- if contains $name .Release.Name -}}
{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
{{- end -}}
{{- end -}}
{{- end -}}
```
Then each resource will use the "fullname" as their name. Then it will have different name for each release, which should solve the issue we have now.

**How to reproduce it (as minimally and precisely as possible)**:
Firstly release a Open Match cluster in "open-match" namespace
```
$ helm upgrade --install release1 --namespace open-match open-match/open-match
```
Then try to do another release in the same namespace. It reports error because there is resource with same name already exists.
```
$ helm upgrade --install release2 --namespace open-match open-match/open-match
```

**Anything else we need to know?**:

**Output of `kubectl version`**:
```
$ kubectl version
[...]

Issue #1244, opened 2020-08-03
Problems with current helm configuration


Laremere:
Preface: I have not looked closely at how all of our helm stuff works, so some of this may be way off.

Our helm charts seem overly complicated, and yet also under powered.  I want to do a deep dive on how it's all set up at some point.  Specifically looking at these items:

- Our charts seem to have deviated from standard practice, see https://github.com/googleforgames/open-match/issues/1235
- The customize subchart should be removed.  We need a chart for the default evaluator, and also a chart for the default evaluator.  If you want a custom evaluator, or are deploying your match function, you should be deploying it like any other k8s service.  Adding our layer on top only confuses things.
- values.yaml and the OM config map should converge, or at least directly include the om config map object in the values.yaml.   This leads to overall system complexity with little point.
- In hindsight, the override configmap is proving to be confusing.  I still maybe think there's some system level stuff which could be configured in a different spot, but maybe not.
- Composition over inclusion: Currently the OM helm chart includes several other charts as sub charts.  This leads to a lot of complexity, and makes certain things more confusing than it needs to be (eg, using your own redis).  It also adds layering problems where all of the configuration you may want to have for redis needs to be pumped through the OM values.yaml.  I don't know if the end result will be simpler without trying, but I'd like to try something like "step 1: Install redis, you can use this chart.  step 2: install OM, configure to point to your redis by setting this config value."  We should provide a nice all in one yaml for the quickstart/demo, but quickly break out after that.
- Specific customization of each component's resources.
- Query should be on a vertical scaling policy, instead of horizontal.

If anyone has opinions, please share.

TBBle:
> * We need a chart for the default evaluator, and also a chart for the default evaluator.

Was one of those "evaluator" occurrences supposed to be "match function"?

Laremere:
I think what I meant to say was one for the demo, fixed.  I specifically don't want a "use this chart for your match function" because we don't know what requirements/configuration your match function needs, and providing a chart which does so is confusing people into thinking they need to use it.

TBBle:
Yeah, that makes sense for open-match-customize. Really, it should be called "open-match-default-replacables" or something, to be clear that this is what it's doing.

[...]

PR #1246, opened 2020-08-05
Use Helm release to name resources

TBBle:
This set of changes ensures that every k8s resource created by the Helm chart includes the `fullname` macro which includes the name of the Helm release.

It's a big change, and is probably much easier to read commit-by-commit, due to the distance between different parts for some of the specific changes.

The main observable effect is that all the resources that were hard-coded to `om-XXX` or `open-match-XXX` are now `RELEASE-open-match-XXX`, although if `RELEASE` contains "open-match" then they will be `RELEASE-XXX`.

In a special case (or rather, removing a special case), e01c77b838fdd86fae4ae8983b030fae552973f4 causes the open-match-scale subchart's resources to be `RELEASE-open-match-scale-XXX`, which means a release named "open-match" gets, e.g., Deployments named `open-match-backend` and `open-match-open-match-scale-backend`. It wasn't clear to be if this is desirable, and if not, that commit can be removed, producing `open-match-scale-backend` instead, but cementing that the subcharts are really just part of the main chart.

Similarly, redis no longer has a `fullnameOverride` set, so its resources will generally be of the form `RELEASE-redis-XXX`, e.g., a StatefulSet named `open-match-redis-master`.

*Note*: Redis PVCs names change with this. Since it's a StatefulSet, the old PVCs are left behind. It's still possible to set `redis.fullnameOverride` to "om-redis" to use any existing PVCs in a namespace, when upgrading. If your release is named `om`, this will be the default name anyway.

As noted in the values.yaml, the `hostName` and `serviceAccount` values now default to empty to generate names, and can be overridden where desired.

To regain the short-names, you can use the YAML file (inside the *Details*) with `--values`. The only thing that will differ is the ConfigMap created by open-match-scale for Grafana dashboards, because it was already named `open-match-scale-dashboard`. If you do this, you can only have one such release in a namespace, and you cannot then have a release named 'om' or 'open-match' installed at the same time, as some resources will conflict.

<details>

```yaml
# Reinstate all the resource names as if I hadn't done anything
fullnameOverride: "om"
open-match-scale:
  fullnameOverride: "om"
  configs:
    default:
      configName: "om-configmap-default"
    override:
      configName: "om-configmap-override"
  scaleBackend:
    hostName: "om-scale-backend"
  scaleFrontend:
    hostName: "om-scale-frontend"

open-match-customize:
  fullnameOverride: "om"

redis:
  fullnameOverride: "om-redis"
  serviceAccount:
    name: "open-match-redis-service"
global:
  kubernetes:
    serviceAccount: "open-match-unprivileged-service"

configs:
  default:
    configName: "om-configmap-default"
  override:
    configName: "om-configmap-override"
```

</details>

Tested to support multiple installs using the following commands from _install/helm/open-match_
```
helm dep build
helm install open-match . --values om-install-all.values.yaml --set-string global.image.tag=1.0.0
helm install another-open-match . --values om-install-all.values.yaml --set-string global.image.tag=1.0.0
```

with TLS keys generated per [TLS Encryption](https://open-match.dev/site/docs/guides/tls/)

where _om-install-all-values.yaml_ is set to install as many of the templated resources as possible, as below:

```yaml
open-match-core:
    enabled: true
    swaggerui:
        enabled: true
    redis:
        enabled: true

open-match-customize:
    enabled: true
    function:
        enabled: true
    evaluator:
        enabled: true

open-match-scale:
    enabled: true

open-match-telemetry:
    enabled: true

open-match-override: true

usingHelmTemplate: false

ci: false

global:
    tls:
        enabled: true
    telemetry:
        grafana:
            enabled: true
        jaeger:
            enabled: true
        prometheus:
            enabled: true
```

I repeated the test with `--set ci=true` appended to both, to confirm that the testing resources do not conflict.

Fixes: #1235

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -363,7 +363,7 @@ install-scale-chart: install-chart-prerequisite build/toolchain/bin/helm$(EXE_EX
 install-ci-chart: install-chart-prerequisite build/toolchain/bin/helm$(EXE_EXTENSION) install/helm/open-match/secrets/
 	$(HELM) upgrade $(OPEN_MATCH_HELM_NAME) $(HELM_UPGRADE_FLAGS) --atomic install/helm/open-match $(HELM_IMAGE_FLAGS) \
 		--set query.replicas=1,frontend.replicas=1,backend.replicas=1 \
-		--set evaluator.hostName=test \
+		--set evaluator.hostName=open-match-test \
 		--set evaluator.grpcPort=50509 \
 		--set evaluator.httpPort=51509 \
 		--set open-match-core.registrationInterval=200ms \
--- a/Makefile
+++ b/Makefile
@@ -412,6 +412,7 @@ install/yaml/03-prometheus-chart.yaml: build/toolchain/bin/helm$(EXE_EXTENSION)
 		--set global.telemetry.prometheus.enabled=true \
 		install/helm/open-match > install/yaml/03-prometheus-chart.yaml
 
+# We have to hard-code the Prometheus Server URL as we are excluding Prometheus, so Helm cannot determine the URL from the Prometheus subchart
 install/yaml/04-grafana-chart.yaml: build/toolchain/bin/helm$(EXE_EXTENSION)
 	mkdir -p install/yaml/
 	$(HELM) template $(OPEN_MATCH_HELM_NAME) $(HELM_TEMPLATE_FLAGS) $(HELM_IMAGE_FLAGS) \
@@ -419,6 +420,7 @@ install/yaml/04-grafana-chart.yaml: build/toolchain/bin/helm$(EXE_EXTENSION)
 		--set open-match-core.redis.enabled=false \
 		--set open-match-telemetry.enabled=true \
 		--set global.telemetry.grafana.enabled=true \
+		--set-string global.telemetry.grafana.prometheusServer="http://$(OPEN_MATCH_HELM_NAME)-prometheus-server.$(OPEN_MATCH_KUBERNETES_NAMESPACE).svc.cluster.local:80/" \
 		install/helm/open-match > install/yaml/04-grafana-chart.yaml
 
 install/yaml/05-jaeger-chart.yaml: build/toolchain/bin/helm$(EXE_EXTENSION)
--- a/Makefile
+++ b/Makefile
@@ -389,9 +389,12 @@ install/yaml/: TAG = $(BASE_VERSION)
 endif
 install/yaml/: update-chart-deps install/yaml/install.yaml install/yaml/01-open-match-core.yaml install/yaml/02-open-match-demo.yaml install/yaml/03-prometheus-chart.yaml install/yaml/04-grafana-chart.yaml install/yaml/05-jaeger-chart.yaml install/yaml/06-open-match-override-configmap.yaml install/yaml/07-open-match-default-evaluator.yaml
 
+# We have to hard-code the Jaeger endpoints as we are excluding Jaeger, so Helm cannot determine the endpoints from the Jaeger subchart
 install/yaml/01-open-match-core.yaml: build/toolchain/bin/helm$(EXE_EXTENSION)
 	mkdir -p install/yaml/
 	$(HELM) template $(OPEN_MATCH_HELM_NAME) $(HELM_TEMPLATE_FLAGS) $(HELM_IMAGE_FLAGS) \
+		--set-string global.telemetry.jaeger.agentEndpoint="$(OPEN_MATCH_HELM_NAME)-jaeger-agent:6831" \
+		--set-string global.telemetry.jaeger.collectorEndpoint="http://$(OPEN_MATCH_HELM_NAME)-jaeger-collector:14268/api/traces" \
 		install/helm/open-match > install/yaml/01-open-match-core.yaml
 
 install/yaml/02-open-match-demo.yaml: build/toolchain/bin/helm$(EXE_EXTENSION)
--- a/Makefile
+++ b/Makefile
@@ -462,7 +462,7 @@ set-redis-password:
 		read REDIS_PASSWORD; \
 		stty echo; \
 		printf "\n"; \
-		$(KUBECTL) create secret generic om-redis -n $(OPEN_MATCH_KUBERNETES_NAMESPACE) --from-literal=redis-password=$$REDIS_PASSWORD --dry-run -o yaml | $(KUBECTL) replace -f - --force
+		$(KUBECTL) create secret generic open-match-redis -n $(OPEN_MATCH_KUBERNETES_NAMESPACE) --from-literal=redis-password=$$REDIS_PASSWORD --dry-run -o yaml | $(KUBECTL) replace -f - --force
 
 install-toolchain: install-kubernetes-tools install-protoc-tools install-openmatch-tools
 install-kubernetes-tools: build/toolchain/bin/kubectl$(EXE_EXTENSION) build/toolchain/bin/helm$(EXE_EXTENSION) build/toolchain/bin/minikube$(EXE_EXTENSION) build/toolchain/bin/terraform$(EXE_EXTENSION)
--- a/install/helm/open-match/subcharts/open-match-customize/templates/_helpers.tpl
+++ b/install/helm/open-match/subcharts/open-match-customize/templates/_helpers.tpl
@@ -1,0 +1,20 @@
+{*
+ Copyright 2019 Google LLC
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*}
+
+{{/* vim: set filetype=mustache: */}}
+{{- define "openmatchcustomize.function.hostName" -}}
+{{- .Values.function.hostName | default (printf "%s-function" (include "openmatch.fullname" . ) ) -}}
+{{- end -}}
--- a/install/helm/open-match/templates/_helpers.tpl
+++ b/install/helm/open-match/templates/_helpers.tpl
@@ -192,6 +192,22 @@ targetCPUUtilizationPercentage: {{ .Values.global.kubernetes.horizontalPodAutoSc
 {{- printf "%s-configmap-override" (include "openmatch.fullname" . ) -}}
 {{- end -}}
 
+{{- define "openmatch.jaeger.agent" -}}
+{{- if index .Values "open-match-telemetry" "enabled" -}}
+{{- if index .Values "open-match-telemetry" "jaeger" "enabled" -}}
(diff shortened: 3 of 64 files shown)


TBBle:
My build failed, but I cannot see why, as I don't have access to the logs.

TBBle:
I tried to run the build on our own GCP account, but it fails with
```
PS D:\third-party-repos\open-match> gcloud builds submit --config=cloudbuild.yaml --substitutions=_OM_VERSION=DEV .
Creating temporary tarball archive of 483 file(s) totalling 13.0 MiB before compression.
Uploading tarball of [.] to [gs://GCP_ID_cloudbuild/source/1596626035.019253-a76e8f7428934d93bc573650e12090a6.tgz]
ERROR: (gcloud.builds.submit) INVALID_ARGUMENT: Request contains an invalid argument.
```

There's no clues as to which argument is invalid, nor any logs in Cloud Build that I can see.

```json
{
  "error": {
    "code": 400,
    "message": "Request contains an invalid argument.",
    "status": "INVALID_ARGUMENT"
  }
}
```

Laremere:
Due to limitations from google cloud build, you are required to join the Open Match mailing list: https://groups.google.com/forum/#!forum/open-match-discuss

install/helm/open-match/subcharts/open-match-customize/templates/_helpers.tpl
```diff
@@ -0,0 +1,40 @@
...
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+{{- end }}
+
+{{- define "openmatch.function.hostName" -}}
+{{- .Values.function.hostName | default (printf "%s-function" (include "openmatch.fullname" . ) ) -}}
+{{- end -}}
```
lihaif, line 40: Why `openmatch.function.hostName` is defined here while `openmatch.evaluator.hostName` is defined in  `install/helm/open-match/templates/_helpers.tpl`. Both of them are customizable components and should be at the same level.

lihaif: `openmatch.evaluator.hostName` is also used in `om-configmap-override.yaml` out of the sub-chart. Defining it in the parent chart means that this sub-chart can not be installed independently. But it may not be an issue.

TBBle: `openmatch.function.hostName` is defined here because it's only _used_ here, while, as you have noted, `openmatch.evaluator.hostName` is also used (once) in the parent chart. See #1244. The configuration for evaluator hostname (and matchFunction hostname) are still done in the top-level chart's values.yaml.

I probably should have renamed `openmatch.function.hostName` to `openmatchcustomize.function.hostName`, on consideration.

I'll fix that, and I'll make sure that the subcharts don't carry straight-up copies of the parent chart templates, since I haven't replicated all the used named templates down into the subcharts.

The structure of the charts means that the sub-charts are a long way from being independently installed, as there is still a log of dependence in the subcharts on parent chart named templates. I didn't try to address this here, except when it was easy, and I'm not sure it's a goal to have them independently installable.

I suggested in #1244 rolling the sub-charts back into the main chart, partly for this reason, because it sends a slightly weird signal on first glance.

TBBle:
I've joined the discussion group, and I can see the build logs now.

TBBle:
CI failure was
```
--- FAIL: TestMatchFunctionMatchCollision (0.45s)
    fetch_matches_test.go:158: 
        	Error Trace:	fetch_matches_test.go:158
        	Error:      	Should be true
        	Test:       	TestMatchFunctionMatchCollision
        	Messages:   	218.840219ms
```

which should not have any connection to my changes, but appears to have been a 20ms overshoot for a 200ms interval, but it's not clear if this is a failure, because it's not clear what this is actually trying to protect/ensure.

We'll see if a new test run fails the same way.

TBBle:
Good. Test rerun got past the timeout issue, and is now seeing the changes.

TBBle:
I still need to fix the two Grafana dashboards which hard-code the pod names (redis.json and go-processes.json) to use labels instead, but apart from that this should be ready for review.

TBBle:
Okay, this should now be complete. I believe I've looked at and addressed (or ignored) every occurrance of `om-` or `open-match-` in this codebase.

The docs are going to need work to align with this too.

Both the Jaeger and Grafana setup were already invalid if you didn't name your release `open-match`, as they had hard-coded names which the subcharts generate based on the release name. This was fixed as well, although it pushes some hard-coding into the Makefile when trying to generate partial manifests that refer to data only known by previous or future partial manifests.

It's also probably feasible to combine Helm options weirdly and get invalid setups, but hopefully that's limited to situations that would be invalid anyway, e.g., enabling Grafana without Prometheus won't give you a usable outcome, but it never really did...

Laremere:
Hey thanks for the PR.

I've already starting looking through it, but there's a lot going on here so the review will take a bit, and probably take a couple rounds...

Are you able to handle the documentation changes as well?

TBBle:
Yeah, I'm happy to do the documentation changes as well. I didn't want to start on those until I had some feedback on the work itself, since it lives in a separate repo so I can't make it part of this PR anyway, and I don't want to rework the documentation changes if there's UX-visible changes needed in this PR.

Makefile
```diff
@@ -389,9 +389,12 @@ install/yaml/: TAG = $(BASE_VERSION)
 endif
 install/yaml/: update-chart-deps install/yaml/install.yaml install/yaml/01-open-match-core.yaml install/yaml/02-open-match-demo.yaml install/yaml/03-prometheus-chart.yaml install/yaml/04-grafana-chart.yaml install/yaml/05-jaeger-chart.yaml install/yaml/06-open-match-override-configmap.yaml install/yaml/07-open-match-default-evaluator.yaml
 
+# We have to hard-code the Jaeger endpoints as we are excluding Jaeger, so Helm cannot determine the endpoints from the Jaeger subchart
```
Laremere, line 392: So... I get why you're doing this, but it's really uggly.  Are there any other setups which would mean this doesn't have to happen?  Just probing for options.

install/helm/open-match/templates/_helpers.tpl
```diff
@@ -135,3 +155,72 @@ minReplicas: {{ .Values.global.kubernetes.horizontalPodAutoScaler.minReplicas }}
 maxReplicas: {{ .Values.global.kubernetes.horizontalPodAutoScaler.maxReplicas }}
 targetCPUUtilizationPercentage: {{ .Values.global.kubernetes.horizontalPodAutoScaler.targetCPUUtilizationPercentage }}
 {{- end -}}
+
+{{- define "openmatch.serviceAccount.name" -}}
```
Laremere, line 159: All of these names are very redundant - is there any way to simplify?

Makefile
TBBle: Not easily. We're generating a config that includes Jaeger settings, separately from deploying Jaeger itself. To fix this, we'd want to be able to have the jaeger-deployment step generate someting that Open Match could pick up to gain knowledge of where Jaeger lives.

We use Jaeger Operator with side-car injection in our project, which removes all of this pain.

install/helm/open-match/templates/_helpers.tpl
TBBle: I'm not sure what you mean here. `openmatch` is the namespace, and it's the name of the service account. Which part is looking redundant to you?

TBBle:
Frankly, _this chart setup_ has been done in a very complex (and rather non-idiomatic) way, which is why it's such a pain to work with. It's not really a Helm thing, it's mostly because it's trying to install a _lot_ of stuff optionally, in very manual ways, and has not really taken good advantage of Helm.

The entire telemetry subchart would be greatly simplified, for example, if we made use of operators rather than directly installing (probably-duplicate) daemonsets and deployments for Prometheus, Grafana, Jaeger, etc.

I haven't pushed on simplification here at all, because I wanted to leave that for #1244 instead; I feel there's already plenty going on in this PR already.

install/helm/open-match/templates/_helpers.tpl
TBBle: Oh do you mean having the big-long list of named templates, rather than the name itself? If we were to drop, for example, being able to override the service names for these services, then we could just use `{{ include "openmatch.fullname" . }}-XXX` everywhere the named template is used.

We could have a name-making named template where the `XXX` above could be a parameter, if that would be more-readable at the call-site (I think it'd be aesthetically worse, personally)

However, a _big_ advantage of the named templates as they stand now, is that someone subcharting Open Match, i.e. me, doesn't need to hard-code the names of the services, I can just ask the subchart what the value of `openmatch.frontend.hostName` is in my own chart. Pretty much exactly what we do for getting the various subchart service URLs, e.g. redis, here.

Makefile
TBBle: I just realised I wrote "so Helm" instead of "as Helm" in this comment. And left out that the reason for this is because we're not installing the relevant subchart at the time.

When using Helm to install Helm (rather than trying to generate partial manifests), this isn't an issue, the Helm templates just ask the sub-charts's named template for the relevant name.

Laremere:
Would it be easier to start from scratch?

TBBle:
I don't think so, particularly. Easier to evolve from working state to working state. And we'd need to have the same discussions either way about what aspects of the current Helm chart UX can be or should be changed, either way.

tools/certgen/certgen.go
```diff
@@ -30,16 +30,16 @@ var (
 	// serviceAddresses is a list of all the HTTP hostname:port combinations for generating a TLS certificate.
 	// It appears that gRPC does not care about validating the port number so we only add the HTTP addresses here.
 	serviceAddressList = []string{
-		"om-backend:51505",
+		"open-match-backend:51505",
 		"om-demo:51507",
```
Laremere, line 34: why aren't these values changing.

install/helm/open-match/templates/_helpers.tpl
Laremere: Ah, yeah that seems pretty useful.  It pains me greatly, but seems worth it.

Laremere:
So I think were I am is that this PR looks pretty good, I don't spot in obvious flaws (sans outstanding comments), and the tests are happy.  I'm not happy with how the helm charts keep growing in complexity, but I'll approve the PR after the comments are resolved.

Makefile
Laremere: The current stacking of the config map was an attempt to simplify things, specifically for the yaml install path.  The blame goes on me for suggesting it, but I think it's proving to be less helpful than I thought.

tools/certgen/certgen.go
TBBle: I have no idea where they come from. They're not part of the Helm chart generated code. I think they're used in the tutorials?

TBBle: Some of them come from https://github.com/googleforgames/open-match/blob/master/install/02-open-match-demo.yaml for example, which appears to be hand-maintained. I assume that will be cleaned up as part of #1244 cleaning up the openmatch-customize subchart.

TBBle:
Am I correct in assuming that the `0.0.0-dev` version in the Helm repository is an autobuild from master?

Also, note to @TBBle, do _not_ forget the doc updates.

Laremere:
I'm actually not sure - I haven't used the helm repository and others set it up.

Issue #1102, opened 2020-02-04
Explore automatic logging of GRPC response errors


Laremere:
When error handling, it is generally bad practice to both log an error, and also return it.  This results in noisy logs for errors which may be properly handled, or logged again at a lower level with more context.

Generally, the pattern should be one of:
- Handle the error without propagating if possible (eg, retries where appropriate).
- Wrap the error with additional useful info/context if useful.  Then return the wrapped error to the caller.

One place where we reasonably violate this is in GRPC services, where we log the error before returning it to GRPC.  However this leads to inconsistency in the code, where some functions do or don't log returned errors, and only manual review is keeping it correct.  It would be nice to use a GRPC plugin or something which is able to intercept the error and log it. (I haven't checked, is this already done?)

Once errors are logged by the system, the service code should no longer log and return errors.

yfei1:
Yes, this is possible. I digged into the grpc interceptors while working on the monitoring stuffs a few months ago but haven't prioritize this task since this is not the main topic for v0.9. I can take it - some of our grpc interceptors need extra configurations to work properly.

akremsa:
Should we consider this ticket as done?

Laremere:
yes, thanks

PR #1248, opened 2020-08-23
Reduce logging in statestore - redis

### LINKED ISSUE #1228: Unnecessary logging when waiting for assignments is canceled.
Currently, when a waiting for assignment, a client will start the call and receive a message when the assignment has been updated.  Then it drops the call, because it doesn't need it anymore.

On the next iteration of the wait for assignment loop, it then logs in two places (at least) that the context is canceled:
https://github.com/googleforgames/open-match/blob/master/internal/statestore/redis.go#L584

https://github.com/googleforgames/open-match/blob/master/internal/statestore/redis.go#L228

This is partially a result of over logging that needs to be cleaned up as part of https://github.com/googleforgames/open-match/issues/1102

However, there's another issue here:
We've now stated that getting an assignment is a terminal state for a ticket (besides being deleted), so there's no point in the wait for assignment call to continue querying the database.  It will always get the same answer until it errors due to the ticket not being there, or canceled by the caller.

Should the FE change so that it will terminate the call once it successfully gets an assignment?  I think that behavioral change falls into the backwards compatibility promises for 1.0, but just barely.

Otherwise, perhaps we need to both: Fix the over logging that's occurring, and suppress logging for canceled context on wait for assignment, though that starts to get iffy on suppressing errors the user might care about.  We could get smart about when to suppress (eg, only after assignment was found), but that's getting even more magical and weird.

akremsa:
Closes #1228 

Previously _context canceled_ message was mentioned twice: 
1) connect method
2) GetAssignments method, after GetTicket returns an error.
Logs before the fix:
image

I suggest to log everything on a very top level of API and just wrap everything which comes from lower levels.
Also, I've made logs more informative adding an ID to each record.

Logs after the fix:
image

### DIFF
--- a/internal/statestore/redis.go
+++ b/internal/statestore/redis.go
@@ -220,45 +220,23 @@ func redisURLFromAddr(addr string, cfg config.View, usePassword bool) string {
 	return redisURL + addr
 }
 
-func (rb *redisBackend) connect(ctx context.Context) (redis.Conn, error) {
-	redisConn, err := rb.redisPool.GetContext(ctx)
-	if err != nil {
-		return nil, status.Errorf(codes.Unavailable, "failed to connect to redis: %v", err)
-	}
-	return redisConn, nil
-}
-
 // CreateTicket creates a new Ticket in the state storage. If the id already exists, it will be overwritten.
 func (rb *redisBackend) CreateTicket(ctx context.Context, ticket *pb.Ticket) error {
-	if ticket == nil {
-		return errors.New("nil ticket provided")
-	}
-
-	redisConn, err := rb.connect(ctx)
+	redisConn, err := rb.redisPool.GetContext(ctx)
 	if err != nil {
-		redisLogger.WithFields(logrus.Fields{
-			"error": err.Error(),
-		}).Errorf("error in CreateTicket, id: %s", ticket.Id)
-		return err
+		return status.Errorf(codes.Unavailable, "CreateTicket, id: %s, failed to connect to redis: %v", ticket.GetId(), err)
 	}
 	defer handleConnectionClose(&redisConn)
 
 	value, err := proto.Marshal(ticket)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internal/statestore/redis.go
```diff
@@ -223,18 +223,22 @@ func redisURLFromAddr(addr string, cfg config.View, usePassword bool) string {
...
 
 // CreateTicket creates a new Ticket in the state storage. If the id already exists, it will be overwritten.
 func (rb *redisBackend) CreateTicket(ctx context.Context, ticket *pb.Ticket) error {
+	if ticket == nil {
+		return status.Error(codes.InvalidArgument, "nil ticket provided")
+	}
+
 	redisConn, err := rb.connect(ctx)
 	if err != nil {
+		redisLogger.WithFields(logrus.Fields{
```
Laremere, line 239: Per the linked issue: https://github.com/googleforgames/open-match/issues/1102

It's not good to both log an error and also return it, this is the cause of the initial issue.

internal/statestore/redis.go
```diff
@@ -335,8 +345,15 @@ func (rb *redisBackend) DeleteTicket(ctx context.Context, id string) error {
 
 // IndexTicket indexes the Ticket id for the configured index fields.
 func (rb *redisBackend) IndexTicket(ctx context.Context, ticket *pb.Ticket) error {
+	if ticket == nil {
```
Laremere, line 348: We already have validation in the calling code, eg https://github.com/googleforgames/open-match/blob/master/internal/app/frontend/frontend_service.go

Here and elsewhere in your changes we don't need to be validating at this level.  If there's any missing validation, it should be added in the grpc service.

akremsa:
@Laremere PR is updated, please check.
The current output is:
image

Laremere:
Great work!  Please follow up with cleaning up elsewhere (the non-streaming calls).

PR #1249, opened 2020-08-23
Use require in tests to avoid nil pointer exceptions

akremsa:
Several fixes in order to avoid nil pointer exceptions in cases if tests fail.


akremsa:
Ok, I get the idea.

akremsa:
@Laremere, please check

Laremere:
Thanks for the PR!

PR #1250, opened 2020-08-25
Added a PR template

akremsa:
I suggest using a pull request template. 

Do we have any kind of labels related to PRs? I've checked a bunch of PRs, didn't notice that.

Let's discuss what can be added here as well.


Laremere:
Thanks!

PR #1251, opened 2020-08-26
More consistent helm package versions

TBBle:

There's a few fixes here for the generated Helm charts, some low-hanging fruit from #1244 and a fix for some issues noticed while working on #1246.

The major visible fix is that Helm charts packaged from non-release builds will now have a unique AppVersion (the container image tag), so that it's possible to know which version of Open Match you have installed. I haven't changed the (SemVer) Version, as that's not low-hanging fruit per #1244.

This also has the effect that such Helm charts will point to the specific container image tag from which they were built, rather than using a floating image tag. This will avoid surprising issues in an Open Match deployment from a non-release build. It's generally a bad idea to use floating image tags in a Kubernetes Deployment. The manifest YAML generated from the same builds did not use the floating image tags, so this problem was only visible to Helm users.

In order to reduce the risk of similar issues in future, the manifest YAML is generated from the packaged chart, not the chart source. At this point, this probably makes no difference, but as other improvements flow from #1246, it will reduce one point of divergence between the manifest YAML and the Helm package. CI deployments still use the chart source because they need to generate TLS certificates. Fixing this (and hence supporting TLS certificates for the packaged chart) is also not low-hanging fruit.

I also fixed an issue where the open-match-telemetry subchart was not having its dependencies updated, meaning it was specified to use one version of the third-party charts it depends on, but was deploying much older versions in practice.

Improvements towards #1244

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -294,7 +294,11 @@ lint-chart: build/toolchain/bin/helm$(EXE_EXTENSION) build/toolchain/bin/ct$(EXE
 	$(CHART_TESTING) lint --all --chart-yaml-schema $(TOOLCHAIN_BIN)/etc/chart_schema.yaml --lint-conf $(TOOLCHAIN_BIN)/etc/lintconf.yaml --chart-dirs $(REPOSITORY_ROOT)/install/helm/
 	$(CHART_TESTING) lint-and-install --all --chart-yaml-schema $(TOOLCHAIN_BIN)/etc/chart_schema.yaml --lint-conf $(TOOLCHAIN_BIN)/etc/lintconf.yaml --chart-dirs $(REPOSITORY_ROOT)/install/helm/
 
-build/chart/open-match-$(BASE_VERSION).tgz: build/toolchain/bin/helm$(EXE_EXTENSION) lint-chart
+# TODO: Restructure Helm charts so that the default for global.images.tag can be `.Chart.AppVersion`, and then remove this patching.
+patch-chart-image-tag:
+	$(SED_REPLACE) 's|    tag: .*|    tag: $(TAG)|g' $(REPOSITORY_ROOT)/install/helm/open-match/values.yaml
+
+build/chart/open-match-$(BASE_VERSION).tgz: build/toolchain/bin/helm$(EXE_EXTENSION) lint-chart patch-chart-image-tag
 	mkdir -p $(BUILD_DIR)/chart/
 	$(HELM) package -d $(BUILD_DIR)/chart/ --version $(BASE_VERSION) --app-version $(TAG) $(REPOSITORY_ROOT)/install/helm/open-match
 
@@ -318,7 +322,7 @@ install-chart-prerequisite: build/toolchain/bin/kubectl$(EXE_EXTENSION) update-c
 HELM_UPGRADE_FLAGS = --cleanup-on-fail -i --no-hooks --debug --timeout=600s --namespace=$(OPEN_MATCH_KUBERNETES_NAMESPACE) --set global.gcpProjectId=$(GCP_PROJECT_ID) --set open-match-override.enabled=true --set redis.password=$(REDIS_DEV_PASSWORD)
 # Used for generate static yamls. Install om-configmap-override.yaml as needed.
 HELM_TEMPLATE_FLAGS = --no-hooks --namespace $(OPEN_MATCH_KUBERNETES_NAMESPACE) --set usingHelmTemplate=true
-HELM_IMAGE_FLAGS = --set global.image.registry=$(REGISTRY) --set global.image.tag=$(TAG)
+HELM_IMAGE_FLAGS = --set global.image.registry=$(REGISTRY)
 
 install-demo:
 	cp $(REPOSITORY_ROOT)/install/02-open-match-demo.yaml $(REPOSITORY_ROOT)/install/tmp-demo.yaml
@@ -328,7 +332,7 @@ install-demo:
 	rm $(REPOSITORY_ROOT)/install/tmp-demo.yaml
 
 # install-large-chart will install open-match-core, open-match-demo with the demo evaluator and mmf, and telemetry supports.
-install-large-chart: install-chart-prerequisite install-demo build/toolchain/bin/helm$(EXE_EXTENSION) install/helm/open-match/secrets/
+install-large-chart: install-chart-prerequisite install-demo build/toolchain/bin/helm$(EXE_EXTENSION) install/helm/open-match/secrets/ patch-chart-image-tag
 	$(HELM) upgrade $(OPEN_MATCH_HELM_NAME) $(HELM_UPGRADE_FLAGS) --atomic install/helm/open-match $(HELM_IMAGE_FLAGS) \
 		--set open-match-telemetry.enabled=true \
 		--set open-match-customize.enabled=true \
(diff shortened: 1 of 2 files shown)


TBBle:
I've just looked at the last-published 0.0.0-dev chart, and I realised that it _did_ package a set of TLS secrets. I'm not sure if that is intentional or not, since the Makefile was set up to only depend on the secrets for the CI deployments, not the YAML or Helm packages, but I guess there's an ordering thing going on there.

TBBle:
While testing an Open Match version without this fix, that the currently-published 0.0.0-dev Helm chart doesn't work, because it is looking for images like _gcr.io/open-match-public-images/openmatch-backend:0.0.0-dev_, but the images from the auto-builds are pushed to _gcr.io/open-match-build/openmatch-backend:0.0.0-dev_. This PR won't fix that, but won't make it worse.

This should probably be added to 256a831 since that fixes a different issue in the tag in a similar way.

Makefile
```diff
@@ -285,16 +285,23 @@ $(foreach IMAGE,$(IMAGES),clean-$(IMAGE)-image): clean-%-image:
...
+	(cd $(REPOSITORY_ROOT)/install/helm/open-match/subcharts/open-match-telemetry; $(HELM) dependency update)
 
 lint-chart: build/toolchain/bin/helm$(EXE_EXTENSION) build/toolchain/bin/ct$(EXE_EXTENSION)
 	(cd $(REPOSITORY_ROOT)/install/helm; $(HELM) lint $(OPEN_MATCH_HELM_NAME))
 	$(CHART_TESTING) lint --all --chart-yaml-schema $(TOOLCHAIN_BIN)/etc/chart_schema.yaml --lint-conf $(TOOLCHAIN_BIN)/etc/lintconf.yaml --chart-dirs $(REPOSITORY_ROOT)/install/helm/
 	$(CHART_TESTING) lint-and-install --all --chart-yaml-schema $(TOOLCHAIN_BIN)/etc/chart_schema.yaml --lint-conf $(TOOLCHAIN_BIN)/etc/lintconf.yaml --chart-dirs $(REPOSITORY_ROOT)/install/helm/
 
-build/chart/open-match-$(BASE_VERSION).tgz: build/toolchain/bin/helm$(EXE_EXTENSION) lint-chart
+patch-chart-image-values:
+	$(SED_REPLACE) 's|    registry: .*|    registry: $(REGISTRY)|g' $(REPOSITORY_ROOT)/install/helm/open-match/values.yaml
```
Laremere, line 298: values.yaml is commited, so running any command which calls into this would cause unwanted churn on these file, right?

Why are you replacing the current helm_image_tags?

TBBle: Because right now, the generated chart contains *floating* image tags (0.0.0-dev), and I don't think `imagePullPolicy: Always` is set here.

So installing this chart will mean your nodes are running whatever version of the image happened to be built from master at the time of first pull *for that node*. Over time, this means you'll end up with a mix of different versions of the same image, which seems a huge risk.

TBBle: Patching the registry is needed because the auto-build images are not pushed to `gcr.io/$(OPEN_MATCH_PUBLIC_IMAGES_PROJECT_ID)` (which is what's in the committed values.yaml), but to `gcr.io/$(OPEN_MATCH_BUILD_PROJECT_ID)`. 

The idea here is to test the same thing we ship. Before this change, the behaviour was to generate templates from the Helm chart, and patch _those generated files_ for CI, leaving the Helm chart untested and invalid out-of-the-box.

Another option would be to make the helm calls for CI, _and the users of the 0.0.0-dev chart_, pass `--set global.image.registry=gcr.io/$(OPEN_MATCH_BUILD_PROJECT_ID)` to Helm on each invocation (or when using it as a subchart, hard-coding that in the parent chart's values.yaml).

TBBle: Anyway, the change made by this target should not be committed, since it's patching in build-specific values.

Laremere: Right, I don't want commands making changes to files which shouldn't be committed (exception: anything in .gitignore, obviously).  At best it'd be annoying to revert, especially when there's other changes to the same file, at worst changes accidentally make there way into being committed.

Why is simply setting the variables when building the charts not the right course of action here?

Sorry for the delay reviewing this - I wasn't involved in writing the charts nor the build system so I need time to become familiar and figure out the best solution here.

TBBle: You can't set variables except `.Chart.Version` and `.Chart.AppVersion` when calling `helm package`. The best you can do is change the values.yaml before calling `helm package`. That's why I suggested as a TODO to move the image tag be `.Chart.AppVersion`, because we _can_ set that at package-build time, to remove 50% of this editing.

To get rid of the other 50% of the editing, we'd need to instead have a macro reacting to `.Chart.AppVersion` or `.Chart.Version`, and recognising an auto-build, to replace the existing hard-coded default repo with _two_ hard-coded default repos, selected by, e.g. the presence of `-dev` in `.Chart.Version`.

However, both of the above changes are complicated by the current muliti-chart setup, as the sub-charts _always_ have version `0.0.0-dev`, even in release builds, because we don't use `helm package` for them, but source them inline, which means there's no opportunity to set `.Chart.Version` and `.Chart.AppVersion` at package time.

We could, as a separate change, modify the build system to do `helm package` for the subcharts rather than `helm dep update` on the parent charts, perhaps. But per #1244, the whole subcharting setup in Open Match is pretty messy, and I'd rather spent the time unwinding that, if I was going to spend time on it. (And any Open Match time I spend has higher priorities, like the outstanding documentation updates for #1235).

Perhaps with some clever scripting, the values.yaml changes could be reverted as part of the Make execution? Although obviously that won't fly if the process fails and the user just runs it again, because then on next run, it'd be changing _from_ the modified files anyway, and change them back to the modified files.

To be fair, this is a pretty common problem I've hit with container image repositories and registries, if you have separate 'release' and 'auto-build' output locations. There was [a proposal for Helm to make image-registry handling more specific](https://github.com/helm/helm/issues/7754) and allow solving issues like this more streamlined, but it didn't seem to go anywhere.

Laremere: > You can't set variables except .Chart.Version and .Chart.AppVersion when calling helm package. 

(╯°□°)╯︵ ┻━┻

TBBle: There was briefly a feature to do exactly that, but it broke other workflows, e.g., stripped all the comments from the values.yaml, and was backed out.

Laremere: Was thinking about this more, would it make sense to switch to "building" charts?

Basically:

- Copy chart folder from install/ into build/ (which is in .gitignore)
- Build subcharts and put them in the proper folder
- Edit the version and image info
- Build the actual chart

This would also get rid of committed zip file sub charts (which kind of also shows that there are build steps to the charts already, they're just a bit of a hack today.)

TBBle: Yup, that would definitely be an option. And if you're doing that, I'd just copy the current subcharts into the charts/ folder, rather than building them separately and copying the tarballs in, since (as far as I recall) they are not _deployed_ separately anywhere anyway. In effect, Helm's just extracting those those tarballs internally and treating the whole thing as one big directory anyway, so it removes a needless step of "building" the subcharts, when they don't really have an independent life outside the open-match chart.

You'd still run `helm dep up` or `helm dep build` (I forget which one is in the OM workflow) to update external charts, but that won't touch chart _directories_ in the charts/ folder.

Laremere:
There's been a lot of good discussion here, but I think solution as is won't be merged.  I'm closing this PR to reflect that.

Better helm building/packaging is a high priority maintenance issue for OM.

Issue #1200, opened 2020-04-28
Sample/Docs on Skill/Range Based Matchfunctions


syntxerror:
The goal of this sample is to provide the steps and considerations when using Open Match with matchmaking rating(mmr) to construct a matchmaker. Most games rely on mmr to pair players together for a fair and enjoyable match so the sample will generate a random skill level that best fits a typical normalized distribution, allow overlapping Tickets to meet requirements for multiple profiles, and handle collisions by rating match quality.

## Background

Although not all matchmakers use mmr based logic for grouping players together (e.g. Custom Games) most do. One standard for many matchmakers is [Glicko-2](https://en.wikipedia.org/wiki/Glicko_rating_system), which takes a player rating range with calculated deviation bounds and matches players together for a more enjoyable game. 

## In-Scope/Out-of-Scope

In-Scope

1. Rank/Range Based rating
2. Overlapping Tickets
    1. Deviations in player tickets that allows them to meet requirements in multiple profiles
3. Custom Evaluator
4. Updated Docs

Out-of-Scope

1. Not plug and play
2. Glicko-2 is not a requirement
3. External player profile connections (player profiles in databases)
4. No distributed matchmaking

## Design Ideas

The design of the sample should hit three main points:

1. Adding a "skill rating" or mmr field to Tickets
2. Generating Profiles that will allow skill overlap
3. Writing a basic match function to select tickets that are within skill range and handle collisions

### Adding Skill Ratings to Tickets 

To simulate a Glicko-2-like “New Ratings Deviation” in a truly traditional matchmaking system for the sample it will require a skill level with some deviation added. Also taking into account the normal skill distribution seen in the bell curve below, we should ensure that the skill levels we give our tickets follow normalized distribution.

![image](https://user-images.githubusercontent.com/38893532/82931530-0ef20500-9f55-11ea-872a-d5edfe783ab8.png)

### Creating Profiles

Within our Director, we should create profiles that allow overlap and not negate our deviation. Deviation should allow Tickets to fit a minimum of one profile and a maximum of three. This will prevent the player from crossing a skill bound that is too large for them. For the sample, the skill overlap will be ±.75.

### Handling Collisions/Ensuring Match Quality

To ensure good match quality, a function to determine the quality of the match is necessary. This will also result in handling collisions when a Ticket has generated several match proposals.

syntxerror:
This issue wasn't closed when the PR mentioning this issue wasn't closed

PR #1252, opened 2020-08-26
Skill Based Example using Default Evaluator

jonfoust:
This PR is an example for skill based match functions. Staying true to our documentation on evaluators, this sample uses the default evaluator along with a function to access match score/quality as shown in our current default evaluator tutorial. This was needed based on questions/feedback from Slack discussions.

The sample is a clone of the Default Evaluator tutorial [folder](https://github.com/googleforgames/open-match/tree/master/tutorials/default_evaluator) with updates for skill based matchmaking vs. the mode based + queue entry time currently shown. Match score function returns the difference between highest and lowest skill ratings (the lower the difference, the better match will be). 

Works on #1200 

This will be a two part sample. One for default evaluator and another for custom evaluator which will come later.


jonfoust:
This method is up for discussion: cloning the tutorial folder vs. updated docs + code samples only. Looking for suggestions.

markmandel:
Quick question - should this have unit tests? (In open match, do you run example unit tests with your regular CI stuff? I'm not sure, but thought I'd throw this out there as a suggestion).

More unit testing gooder anyway.

jonfoust:
Moved to Open Match Ecosystem googleforgames/open-match-ecosystem#11

PR #1253, opened 2020-08-30
Improved app tests

akremsa:
Switched from _require_ to _assert_.
Also made output where we compare error codes more readable.

Current output:
```
Error:      	Not equal: 
            	expected: 0xa
            	actual  : 0xe
```

After the fix:
```
Error:      	Not equal: 
            	expected: "Aborted"
            	actual  : "NotFound"
```


aLekSer (approved):
Looks good.

Laremere:
Nitpick: Switched from *assert* to *require*.

PR #1254, opened 2020-09-01
Added a new exception to golangci config

akremsa:
I've noticed that in tests we use a workaround like `tt:=tt` in order to avoid linter fails.
In my opinion, this is not an elegant solution, and I've spent some time researching if there is a way to fix it, and found several issues:
https://github.com/golangci/golangci-lint/issues/281
https://github.com/kyoh86/scopelint/issues/4

I think adding this warning to the exception list makes our code a bit cleaner and easy to understand. Personally, when I've encountered this line in tests for the first time I thought that this was probably a mistake and removed that code and noticed a linter warning much later. 

What do you think?

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -186,6 +186,7 @@ issues:
   # excluded by default patterns execute `golangci-lint run --help`
   exclude:
     - abcdef
+    - Using the variable on range scope .* in function literal
 
   # Excluding configuration per-path, per-linter, per-text and per-source
   exclude-rules:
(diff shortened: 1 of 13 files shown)


Laremere:
See this example of what the link is catching: https://play.golang.org/p/hUyCvFOuAXr

The difference between same code (which runs the passed function immediately), and unsafe code (which may run code at different points) is very subtle so I prefer to keep this pattern.  This is particularly bad for tests because Run /may/ run in parallel, or may not, depending on other pieces of code.

akremsa:
Ok, thanks for the feedback.

PR #1262, opened 2020-09-17
Replaced GS bucket links with substitution variables

galalmounir:

In order to use the cloud build file on a separate GCP project different infrastructures need to be deployed, including a couple of GS Buckets. Unfortunately, the cloud build file has the open-match build pipeline buckets for artifacts and logs hardcoded and not exposed as substitution variables.

This PR creates two substitution variables for the buckets (`_ARTIFACTS_BUCKET`, and `_LOGS_BUCKET`), allowing a google cloud build to succeed by using a different value for these variables.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -153,7 +153,7 @@ steps:
 
 artifacts:
     objects:
-        location: gs://open-match-build-artifacts/output/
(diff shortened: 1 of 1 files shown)


Laremere (approved):

Laremere:
Please bring your branch up to date with master so that I can merge, thanks!

PR #1264, opened 2020-09-23
Improved tests for statestore - redis

akremsa:
I've noticed that redis.go has low unit test coverage and decided to improve that.

Before:
```
go test -coverprofile=before.out
PASS
coverage: 52.3% of statements
ok  	open-match.dev/open-match/internal/statestore	1.394s
```

After:
```
go test -coverprofile=after.out
PASS
coverage: 84.8% of statements
ok  	open-match.dev/open-match/internal/statestore	1.007s
```


internal/statestore/redis_test.go
```diff
@@ -247,6 +357,515 @@ func TestConnect(t *testing.T) {
...
+
+	// Verify all tickets are created and returned
+	verifyTickets(service, tickets)
+
+	c, err := redis.Dial("tcp", fmt.Sprintf("%s:%s", cfg.GetString("redis.hostname"), cfg.GetString("redis.port")))
+	require.NoError(t, err)
+	// Add the first ticket to the pending release and verify changes are reflected in the result
+	redis.Strings(c.Do("ZADD", "proposed_ticket_ids", time.Now().UnixNano(), ids[0]))
+
+	// Verrify 1 ticket is indexed
```
aLekSer, line 756:
```suggestion
	// Verify 1 ticket is indexed
```

akremsa: Thanks

internal/statestore/redis_test.go
```diff
@@ -247,6 +357,515 @@ func TestConnect(t *testing.T) {
...
+			description:     "empty id passed, err expected",
+			ticketID:        "",
+			expectedCode:    codes.NotFound,
+			expectedMessage: "Ticket id: not found",
+		},
+		{
+			description:     "wrong id passed, err expected",
+			ticketID:        "123456",
+			expectedCode:    codes.NotFound,
+			expectedMessage: "Ticket id:123456 not found",
```
aLekSer, line 479: Do we need a whitespace, but that could do this in a separate PR.
```suggestion
			expectedMessage: "Ticket id: 123456 not found",
```

akremsa:
@Laremere could you please take a look?

internal/statestore/redis_test.go
akremsa: ok

aLekSer (approved):
Fine with me.

Laremere (approved):
There was already reasonably good coverage of this through the e2e tests (unfortunately coverage doesn't work across packages.)  This seems like enough of a positive, and don't want to waste your work, so approving.
I think in the future if there are instances where these tests need to change and coverage is handled by e2e tests, I'll be fine with PR authors either fixing this test, or removing test cases here (again, as long as it's also covered well by e2e.)

Laremere:
Thanks for the PR!

PR #1268, opened 2020-10-15
Adding Exclude property to DoubleRangeFilter and test coverage.

### LINKED ISSUE #1260: Allow excluding min/max in DoubleRangeFilter
We have a need to subdivide a continuous variable into ranges that are even and predicable, say, 0.0 - 4.0 into 4 buckets:
`[0.0, 1.0)  [1.0, 2.0), [2.0, 3.0), [3.0, 4.0)`
Each bucket with these boundaries would have no overlap. The current inclusive/inclusive boundaries make this very difficult to do without trying to hand make the boundaries to be hopefully close enough but not equal to the next value that there will not be another value in between the ending of one bucket and the start of the next one ie:
`[0.0, 0.99999999...], [1, 1.99999999...]`

We also have used this type of representation for discrete values as well (like enumerations) where it is even more important that the buckets do not overlap.

We propose no default changes to how things work, by default a query would still run with min/max as inclusive boundaries. 

However, we propose adding a new field to the DoubleRangeFilter contract that introduces the idea that either or both bounds could be exclusive:
```
message DoubleRangeFilter {
  // Name of the ticket's search_fields.double_args this Filter operates on.
  string double_arg = 1;

  // Maximum value.
  double max = 2;

  // Minimum value.
  double min = 3;

  enum Exclude {
    NONE = 0;
    MIN = 1;
    MAX = 2;
    BOTH = 3;
  }
  Exclude exclude = 4;
}
```
This will enable all of the scenarios we currently use and more

An alternative to achieve this is attempt to understand the minimum value between the discrete bucket ranges and do some math that will hopefully be close enough to the next buckets minimum range that there is no possibility of a value existing between it. The main concern here is that the computation of these ranges may be done on different platform architectures and may result in slightly varying results at high precision:
So instead of the desired list of filters for a few non overlapping queries:
`[0.0, 1.0)  [1.0, 2.0), [2.0, 3.0), [3.0, 4.0)`
We could try:
`[0.0, 0.99999999...)  [1.0, 1.9999999...), [2.0, 2.9999999...), [3.0, 3.99999999...)`

This has the potential to miss some values between the max of one range and the next range, in addition to being cumbersome to compute and look at compared to the mathematically concise way of defining these.

**Additional context**
We currently run a matchmaker that we would like to port over to using open match as the ticket backend; our existing logic for generating queries that is exposed to customers has the behavior we describe, and it would be a breaking change we would like to avoid to have to convey to customers that they will now need to simulate ranges like this.

HazWard:
This set of changes gives more control over the excluded/included bounds of DoubleRangeFilters.

Closes #1260


adumovic:
@googlebot I fixed it

Laremere:
PR mostly looks good, I would still prefer bools over having the enum, is there more reasoning behind choosing the enum?  See comments in: https://github.com/googleforgames/open-match/issues/1260

Laremere:
CLA issue is because of the email `HazWard@users.noreply.github.com`

Laremere:
Git amend the PR with your Unity email to fix, I think.  (and if possible, set that as your default github email would avoid this problem, I think?)

HazWard:
Thanks for the tip, I did the merge from the website so it didn't use my work email.

HazWard:
We went with enums because we use them often and it gave us a single fields to indicate which bounds should be excluded.

api/messages.proto
```diff
@@ -103,6 +103,14 @@ message DoubleRangeFilter {
...
   // Minimum value.
   double min = 3;
+
+  enum Exclude {
+    NONE = 0;
+    MIN = 1;
+    MAX = 2;
+    BOTH = 3;
+  }
+  Exclude exclude = 4;
```
Laremere, line 113: Add a comment explaining what this means, with comments on each of the enum values.

Laremere, line 113: Add beta notice (see return ticket api for example).

internal/filter/filter.go
```diff
@@ -106,10 +106,27 @@ func (pf *PoolFilter) In(ticket *pb.Ticket) bool {
 		if !ok {
 			return false
 		}
-		// Not simplified so that NaN cases are handled correctly.
-		if !(v >= f.Min && v <= f.Max) {
-			return false
+
+		switch f.Exclude {
```
Laremere, line 110: Given this implementation, the testing is not through enough.

There are probably ways to reorder this logic to have fewer statements, but given the complexity of these statements, I think we want full testing on these statements.  That means for each conditional (eg, `v >= f.Min`) testing less than, equal to, and greater than cases.  This would securely lock down the backward compatibility and behavior of these fields.  Please use a helper (like simpleDoubleRange) or other simplification so that we can easily see all of the tests cases are covered and that they are correct.

Laremere: Actually, it might make sense for simpleDoubleRange to take the exclusivity argument.

Laremere (changes requested):

internal/filter/filter.go
Laremere: Also, switching to bools would simplify this work (not that we should optimize for writing fewer test cases, but it would show simpler system behavior...) as you could split out the min and max if statements.  However I floated the API choices with a few other people and there's not a clear "this API is better" so I have no preference for which one we put in beta.

api/messages.proto
HazWard: Added a comment + beta notice

Laremere (changes requested):
I still don't see full test coverage - given that you've moved and (mostly) copied 1 line of logic behind a switch statement with 4 clauses, full coverage implies there to be roughly 4 times the number of tests.  I've thought about it more, and suggest grouping by the case, eg:

`
simpleDoubleRange("minIsNan", 5, math.NaN(), 10, pb.DoubleRangeFilter_NONE),
simpleDoubleRange("minIsNan", 5, math.NaN(), 10, pb.DoubleRangeFilter_MIN),
simpleDoubleRange("minIsNan", 5, math.NaN(), 10, pb.DoubleRangeFilter_MAX),
simpleDoubleRange("minIsNan", 5, math.NaN(), 10, pb.DoubleRangeFilter_BOTH),

simpleDoubleRange("maxIsNan", 5, 0, math.NaN(), pb.DoubleRangeFilter_NONE),
simpleDoubleRange("maxIsNan", 5, 0, math.NaN(), pb.DoubleRangeFilter_MIN),
simpleDoubleRange("maxIsNan", 5, 0, math.NaN(), pb.DoubleRangeFilter_MAX),
simpleDoubleRange("maxIsNan", 5, 0, math.NaN(), pb.DoubleRangeFilter_BOTH),

`
(empty line between the groupings)

The cases which are affected by the enum would become obvious, as there won't be full groupings.  The names don't strictly need to be different, though it'd probably be helpful for the cases which are different.

api/messages.proto
```diff
@@ -103,6 +103,26 @@ message DoubleRangeFilter {
 
   // Minimum value.
   double min = 3;
+
+  enum Exclude {
+    // No bounds should be excluded when evaluating the filter, i.e.: MIN >= x <= MAX
```
Laremere, line 108: Should be MIN <= x <= MAX.  (same problem for all comments.)

Laremere: Everything else looks great, though.  Awesome work here!

Laremere (approved):

PR #1271, opened 2020-10-23
Unavailable gRPC match functions forces us to wait the proposalCollectionInterval before failing

HazWard:
Whenever a profile points to a function that is not available, a call to `FetchMatches` with that profile will force us to wait for the full length of `proposalCollectionInterval` before getting back an error saying that the function took longer than the collection interval to return results. It would be expected that if the function is not available, we should fail fast. This change adds an additional check before calling the function to see if the provided address is available. 

**Which issue(s) this PR fixes**: No issues created for this.

Open to suggestions on a more robust way to do this.

### DIFF
--- a/install/helm/open-match/values.yaml
+++ b/install/helm/open-match/values.yaml
@@ -171,8 +171,8 @@ open-match-core:
   pendingReleaseTimeout: 1m
   # Time after a ticket has been assigned before it is automatically delted.
   assignedDeleteTimeout: 10m
-  # Time before a function is considered unavailable after trying to connect to it.
-  functionConnectionTimeout: 300ms
+  # Time before an RPC connection is considered unavailable after trying to connect to it.
+  rpcConnectionTimeout: 300ms
   # Maximum number of tickets to return on a single QueryTicketsResponse.
   queryPageSize: 10000
 
--- a/install/helm/open-match/values.yaml
+++ b/install/helm/open-match/values.yaml
@@ -171,6 +171,8 @@ open-match-core:
   pendingReleaseTimeout: 1m
   # Time after a ticket has been assigned before it is automatically delted.
   assignedDeleteTimeout: 10m
+  # Time before a function is considered unavailable after trying to connect to it.
+  functionConnectionTimeout: 300ms
   # Maximum number of tickets to return on a single QueryTicketsResponse.
   queryPageSize: 10000
 
--- a/internal/app/backend/backend_service.go
+++ b/internal/app/backend/backend_service.go
@@ -198,7 +198,10 @@ func callGrpcMmf(ctx context.Context, cc *rpc.ClientCache, profile *pb.MatchProf
 	var conn *grpc.ClientConn
 	conn, err := cc.GetGRPC(address)
 	if err != nil {
-		return status.Error(codes.InvalidArgument, "failed to establish grpc client connection to match function")
+		if code := status.Code(err); code != codes.Unknown {
+			return err
+		}
+		return status.Errorf(codes.InvalidArgument, "failed to establish grpc client connection to match function: %s", err.Error())
 	}
 	client := pb.NewMatchFunctionClient(conn)
 
--- a/internal/testing/e2e/fetch_matches_test.go
+++ b/internal/testing/e2e/fetch_matches_test.go
@@ -16,6 +16,7 @@ package e2e
 
 import (
 	"context"
+	"errors"
 	"io"
 	"math/rand"
 	"sync"
@@ -23,7 +24,6 @@ import (
 	"time"
 
 	"github.com/golang/protobuf/proto"
-	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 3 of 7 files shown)


internal/app/backend/backend_service.go
```diff
@@ -180,9 +182,12 @@ func synchronizeRecv(ctx context.Context, syncStream synchronizerStream, m *sync
 }
 
 // callMmf triggers execution of MMFs to fetch match proposals.
-func callMmf(ctx context.Context, cc *rpc.ClientCache, req *pb.FetchMatchesRequest, proposals chan<- *pb.Match) error {
+func callMmf(ctx context.Context, cc *rpc.ClientCache, req *pb.FetchMatchesRequest, functionConnectionTimeout time.Duration, proposals chan<- *pb.Match) error {
 	defer close(proposals)
 	address := fmt.Sprintf("%s:%d", req.GetConfig().GetHost(), req.GetConfig().GetPort())
+	if err := isAddressAvailable(address, functionConnectionTimeout); err != nil {
```
Laremere, line 188: This feels like a hack/bandaid.  
I've found that gRPC is weird and operates with different primitives than people tend to be used to with http requests, however they're very well thought out, and will have a correct way to accomplish things.

Specifically, gRPC client doesn't necessarily represent a specific tcp connection, instead it may be connected to multiple  instances and go between working and broken.  What we're saying here is "if this connection isn't currently alive, we want to fail faster than we otherwise would".  gRPC has a couple constructs which would be useful here:

https://pkg.go.dev/google.golang.org/grpc#ClientConn.WaitForStateChange
https://pkg.go.dev/google.golang.org/grpc#ClientConn.GetState

To implement this change, I think we should solve it at a system level instead of just the MMF (fixing this problem with evaluator too).  So that means changing the client cache: https://github.com/googleforgames/open-match/blob/master/internal/rpc/clientcache.go

Something like,
```golang
func (cc *ClientCache) GetGRPC(address string) (*grpc.ClientConn, error) {
	val, exists := cc.cache.Load(address)
	c, ok := val.(cachedGRPCClient)
	if !ok || !exists {
		conn, err := GRPCClientFromEndpoint(cc.cfg, address)
		if err != nil {
			return nil, err
		}
		c = cachedGRPCClient{
			client: conn,
		}
		cc.cache.Store(address, c)
	}
 
        timeoutDuration := readConnectionReadyTimeout()
        ctx := context.WithTimeout(context.Background(), timeoutDuration)
        for {
            s := c.client.GetState()
            if s == connectivity.Ready {
                break
            }
            if !c.client.WaitForStateChange(ctx, s) {
              return nil, status.Errorf(codes.Unavailable, "timeout waiting for connection ready after %s, stuck in state %s", timeoutDuration, s)
            }
         }

	return c.client, nil
}
```

There might be an even better way to specify this, but these apis are complex so I'm not sure :/
Maybe ask on the gRPC slack?

(also see if it's reasonable to pass the context from the call here, don't use it for dialing but only this connection's specific waiting.)

Also, please write a test covering this behavior.

HazWard: Thanks for the feedback, it felt a bit hacky originally but I'm still a bit new to the gRPC primitives so I wasn't sure what I could use.

install/helm/open-match/values.yaml
```diff
@@ -171,6 +171,8 @@ open-match-core:
   pendingReleaseTimeout: 1m
   # Time after a ticket has been assigned before it is automatically delted.
   assignedDeleteTimeout: 10m
+  # Time before a function is considered unavailable after trying to connect to it.
+  functionConnectionTimeout: 300ms
```
Laremere, line 175: Client cache is used for more than connecting to a MMF.  Perhaps `rpcConnectionTimeout` with the comment noting MMFs fall under this?

internal/app/backend/backend_service.go
Laremere: I mostly looked at https://pkg.go.dev/google.golang.org/grpc to find something which would do what we want.

Laremere:
Please see the test failure.  You can run the go tests locally as well to detect failures (as long as they don't happen only on a real cluster).  Looks like I missed something in my code suggestion.

HazWard:
The new test point to a real cluster calling `om-test`. If it points to a real cluster are there steps available to run the tests?

Laremere:
If you look at the failure, ("Details", then "View more details on Google Cloud Build") you can see the logs, and which step it failed at.

You can see what each step does in cloudbuild.yaml: 
https://github.com/googleforgames/open-match/blob/master/cloudbuild.yaml#L109

This appears to be a local failure, where running the go tests without a cluster will fail.

internal/testing/e2e/fetch_matches_test.go
```diff
@@ -841,3 +842,27 @@ func TestHTTPMMF(t *testing.T) {
...
+		Profile: &pb.MatchProfile{},
+	})
+	require.Nil(t, err)
+
+	om.SetEvaluator(func(ctx context.Context, in <-chan *pb.Match, out chan<- string) error {
+		return nil
+	})
+
+	_, err = stream.Recv()
+	require.Error(t, err)
```
Laremere, line 867: Also check that the call takes more than rpcConnectionTimeout but less than registrationInterval + proposalCollectionInterval.

HazWard: I worry that by adding checks like this we'll probably get some timing issues.

Laremere: We've done it in several places without issue.

Laremere, line 867: Also check that the error has code invalid?

HazWard: Since the FetchMatches call wraps the `mmfErr` and the `syncErr` we can't really assert the code.

install/helm/open-match/values.yaml
HazWard: Updated

Laremere (approved):

PR #1273, opened 2020-10-28
Add Backfill support to internal statestore

akremsa:
Added Backfill support to internal statestore. 
Methods: create, get, delete, update.
Renamed _redis.go/redis_test.go_ to _ticket.go/ticket_test.go_ because now there are 2 entities - ticket and backfill and I find such naming to be more consistent.


internal/statestore/backfill_test.go
```diff
@@ -0,0 +1,344 @@
...
+		{
+			description:     "empty id passed, no err expected",
+			ticketID:        "",
+			expectedCode:    codes.OK,
+			expectedMessage: "",
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
```
hsorellana, line 191: what is this convention for??

akremsa: This is an agreement to avoid late binding of tc variable in case of a closure in a for loop.
It's not a case for this particular loop, but we agreed with @Laremere to use such a pattern everywhere.

internal/statestore/backfill.go
```diff
@@ -0,0 +1,152 @@
...
+		return nil, status.Errorf(codes.Internal, "nil updateFunc provided")
+	}
+
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Unavailable, "UpdateBackfill, id: %s, failed to connect to redis: %v", backfill.GetId(), err)
+	}
+	defer handleConnectionClose(&redisConn)
+
+	for i := 0; i < 3; i++ {
```
aLekSer, line 77: I assume we should avoid magic numbers here could go some constant or configurable value.

internal/statestore/backfill.go
```diff
@@ -0,0 +1,152 @@
...
+			return nil, err
+		}
+
+		err = redisConn.Send("MULTI")
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to MULTI, id: %s", backfill.GetId())
+		}
+
+		// update current backfill by invoking an updateFunc which is implemented on the caller side
+		backfillToSet, err := updateFunc(currentBackfill, backfill)
```
aLekSer, line 94: pre-computations can be done before performing Redis queries.

internal/statestore/backfill.go
akremsa: In my opinion, this single case is not worth adding an additional parameter and propagating it everywhere for instance to makefile, helm, or so on. It's really hard to catch and test that scenario live where you can see a big difference in this parameter. But it's better to limit this loop anyway rather than making an infinite one.

internal/statestore/backfill.go
```diff
@@ -0,0 +1,152 @@
...
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to WATCH, id: %s", backfill.GetId())
+		}
+
+		currentBackfill, err := getBackfill(redisConn, backfill.GetId())
+		if err != nil {
+			return nil, err
+		}
+
+		err = redisConn.Send("MULTI")
```
aLekSer, line 88: I wonder should this be before `getBackfill` so that the we have Redis locked before the read and write?
I mean no one should write in between this read and the following write.

internal/statestore/backfill.go
akremsa: What will it change?

internal/statestore/backfill.go
akremsa: This is not a lock actually. There is no difference at all in this case.

internal/statestore/backfill.go
akremsa: I can not do that. If I put MULTI before _getBackfill_, I won't be able to get backfill immediately. That is why WATCH is used here.

internal/statestore/backfill_test.go
aLekSer: I see one of the answers could be found here:
https://github.com/kyoh86/scopelint/issues/4#issuecomment-471661062

internal/statestore/backfill.go
aLekSer: No I mean `updateFunc` before `MULTI`. WDYT?

akremsa: Ok. Yes, that probably makes sense.

api/messages.proto
```diff
@@ -204,3 +204,27 @@ message Match {
   // Deprecated fields.
   reserved 5, 6;
 }
+
+// BETA FEATURE WARNING:  This call and the associated Request and Response
+// messages are not finalized and still subject to possible change or removal.
+// Represents a backfill entity which is used to fill partially full matches.
+message Backfill {
```
Laremere, line 211: Please bring your branch up to date with master.  Thanks!

internal/statestore/backfill.go
```diff
@@ -0,0 +1,152 @@
+package statestore
```
Laremere, line 1: Splitting the files up seems ok to me (redis.go is getting quite big, this would only make it bigger), however the existing redis.go has more than just tickets in it.  It also has generic redis connection stuff.  

Please keep common redis stuff to redis.go, splitting out tickets stuff to tickets.go (preferred), or add everything to redis.go directly.

I would prefer that this split/movement/cleanup was done in a separate PR, though I won't require it for approval this time.

internal/statestore/backfill.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		err = errors.Wrapf(err, "failed to delete the backfill from state storage, id: %s", id)
+		return status.Errorf(codes.Internal, "%v", err)
+	}
+
+	return nil
+}
+
+// UpdateBackfill updates an existing Backfill with new data.
+func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfill, updateFunc func(current *pb.Backfill, new *pb.Backfill) (*pb.Backfill, error)) (*pb.Backfill, error) {
```
Laremere, line 66: It's not clear to me where this update function will be used?

This logic seems to be an optimistic update.
If the backfill is updated by the MMF, as soon as the backfill is updated by anything else there's no reason to retry - the whole fetch match cycle is optimistic and has failed.
When updating from the gameserver, it needs to forcefully update and set the value.

When acknowledging from the gameserver, this field isn't actually updated?
How are the non-user visible fields (pending tickets, last acknowledge time) going to work with this?

internal/statestore/backfill.go
akremsa: I like the idea of leaving 3 files - redis, backfill, ticket. Let's do that right now, anyway I've already started doing that.

internal/statestore/backfill.go
akremsa: > When acknowledging from the gameserver, this field isn't actually updated?

Cold you please clarify which field are you talking about? 

> How are the non-user visible fields (pending tickets, last acknowledge time) going to work with this?

We expect last acknowledge time to be an in-memory variable. 
Regarding pending tickets - next PR will introduce methods to store _ticketIDs_ by _backfillID-generation_ key.

akremsa: I've added an isGS flag to this update function. Is it better?

hsorellana (approved):

internal/statestore/backfill.go
Laremere: > Could you please clarify which field are you talking about?
The redis field, the whole backfill ticket has no changes.

> We expect last acknowledge time to be an in-memory variable.
How does this work?

> Regarding pending tickets - next PR will introduce methods to store ticketIDs by backfillID-generation key.
Hm, that might work, but I'm not quite seeing it.

@aLekSer @akremsa Would you please update the design doc with the fields which are used in redis and the general strategy for working with them?  I think that would help avoid churn on code review.

aLekSer: Yes, I think this should be clear in the design doc, updating it.

> When acknowledging from the gameserver, this field isn't actually updated?

On acknowledgment we expect not to make any updates to Redis, only read from Redis and return back to the GameServer the current state.

>How does this work?

There is no acknowledge time, so with every `acknowledge` request it would be stored in an LRU cache or in a simple `map[backfillId]timestamp`. It was mentioned in discussion that we should not propagate this timestamp to the database, because it would cause frequent updates.

internal/statestore/backfill.go
```diff
@@ -0,0 +1,180 @@
...
+
+	"github.com/golang/protobuf/proto"
+	"github.com/gomodule/redigo/redis"
+	"github.com/pkg/errors"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"open-match.dev/open-match/pkg/pb"
+)
+
+// CreateBackfill creates a new Backfill in the state storage. If the id already exists, it will be overwritten.
```
scosgrave, line 28: > If the id already exists, it will be overwritten.

This sounds dangerous. What are the cons of just having the code that creates the entry also create the ID for it?

internal/statestore/public.go
```diff
@@ -63,8 +68,19 @@ type Service interface {
 	// ReleaseAllTickets releases all pending tickets back to active
 	ReleaseAllTickets(ctx context.Context) error
 
-	// Closes the connection to the underlying storage.
-	Close() error
+	// Backfill
```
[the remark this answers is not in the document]
hsorellana: ... and embeding it into the main interface. Something like: 

```go
type Service interface {
    BackfillService
    ...
}
```

internal/statestore/backfill.go
aLekSer: We were supposing to keep State Store simple as it was, and generate this ID on Frontend. I will sent a PR soon, it would be similar to `CreateTicket`.

internal/statestore/backfill.go
yeukovichd: May be this is a good idea or not but can we add simple methods to state store like:

1. UpdateBackfill(backfill) - just updates backfill value by key
2. UpdateBackfillTickets(backfill, ticketIds) - just update list of tickets associated with backfill. Stored in Redis like [backfillId_tickets] -> [ticket ids]
3. Lock - distributed lock
4. Unlock - distributed unlock

To backend:

1. UpdateBackfill(backfill)

- lock
- get backfill
- check if exists
- release associated tickets
- increment generation
- update backfill
- unlock

2. SysUpdateBackfill(backfill, ticketIds)
- lock
- get backfill
- check if exists
- compare generation
- update backfill
- update associated tickets
- unlock

aLekSer: On AcknowledgeBackfill:

- lock
- get backfill
- check if exists
- assign associated tickets
- unlock


Return Backfill without any tickets.

aLekSer: release associated tickets - remove tickets from pending state. Now they are available for QueryTickets.

akremsa:
@Laremere 
There are 2 ways how we can update storing BF and ticketIDs.

1) Use a single key for BF and associated ticketIDs. In this case, ticketIDs can be added either to backfill protobuf or just use an internal structure like:
`type RedisBackfill struct {
backfill Backfill
ticketIDs []string
}`
In this case, we can keep using transactions (WATCH, MULTI) because we have only 1 instance of backfill entity with all information inside. 
Side effect: there is an error on an UpdateBackfill step, but tickets were put to PendingRelease state. We have separate keys for backfill and pending_tickets. 
Do we need to perform there 2 operations - _Update_ and _Pending Release_ in a scope of a single transaction? 
If no - then we can use this approach.

2) Use 2 keys: **backfillID** for backfill, **backfillID-tickets** for backfill-tickets mapping. In this case, we can use only distributed locks.

We need to choose an approach we are going to follow. I can make an update in the upcoming PR.

internal/statestore/public.go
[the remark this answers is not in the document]
Laremere: I don't see how this simplifies things?  It adds a layer of indirection, and another interface.  So to understand the service interface, I have to follow the code to the BackfillService, which isn't used on its own anywere.

Laremere:
The lock approach seems more initially trackable - if we're not locking we need to be very sure there aren't race conditions.

If we want to store items in redis that aren't a public proto, that's fine - there's internal protobufs too, which can be used.

Need to not only consider the backfills here - when a backfill is acknowledged, the tickets which are acknowledged need to be assigned and then removed from the backfill's pending tickets.

internal/statestore/backfill.go
Laremere: Storing acknowledge timestamps in the frontend or backends won't work, as there are multiple instances of those services.

Lock/unlock shouldn't be visible from the app code for frontend/backend/etc, as the requirement to do locking is because of our use of redis ( I think? ), so I think it would be better to do more meaningful methods inside statestore, which may use private methods to locking, etc.

api/messages.proto
```diff
@@ -261,6 +261,6 @@ message Backfill {
 
   // Generation gets incremented on GameServers update operations.
   // Prevents the MMF from overriding a newer version from the game server.
-  // It is not to be updated by the MMF.
+  // It is not supposed to be updated by the MMF.
```
Laremere, line 264: Perhaps:
"Do not read or write to this field, it is for internal tracking and changing the value will cause bugs."?

internal/statestore/backfill.go
Laremere: The xids algorithm used to create the ids ensures that they are unique with no system wide synchronization.  Calling clients are forbidden from choosing an id during create.  So no conflicts will occur.

internal/statestore/backfill.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		err = errors.Wrapf(err, "failed to delete the backfill from state storage, id: %s", id)
+		return status.Errorf(codes.Internal, "%v", err)
+	}
+
+	return nil
+}
+
+// UpdateBackfill updates an existing Backfill with a new data. Caller has to provide a custom updateFunc if this function is called not for the game server.
+func (rb *redisBackend) UpdateBackfill(ctx context.Context, isGS bool, backfill *pb.Backfill, updateFunc func(current *pb.Backfill, new *pb.Backfill) (*pb.Backfill, error)) (*pb.Backfill, error) {
```
Laremere, line 80: This seems like to unrelated functions that are put together:

if isGS is true, then the function uses the backfill parameter, if it's false it uses the updateFunc parameter.  The only shared logic is the standard connection code.

internal/statestore/redis.go
```diff
@@ -222,418 +213,3 @@ func redisURLFromAddr(addr string, cfg config.View, usePassword bool) string {
...
-	if err != nil {
-		return status.Errorf(codes.Unavailable, "ReleaseAllTickets, failed to connect to redis: %v", err)
-	}
-	defer handleConnectionClose(&redisConn)
-
-	_, err = redisConn.Do("DEL", proposedTicketIDs)
-	return err
-}
-
-func handleConnectionClose(conn *redis.Conn) {
```
Laremere, line 615: This is used by backfill as well, should stay in redis.go?

internal/statestore/public.go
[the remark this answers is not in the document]
hsorellana: My suggestion comes only from the fact that the current Service interface has 10+ funcs and interfaces are supposed to be small.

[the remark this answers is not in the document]
Laremere: That advice is for the total number of methods on an interface - embedding another interface doesn't reduce that.

It's ok for this interface to be large - It's precisely what it needs to be.  This is very specific interface with one to few implementations, where the users of the interface definitely need the entire implementation.

The small interface guidelines are for making interfaces widely implementable and usable.  Eg don't have a function take ReadCloser when just Reader will do - as some implementations can't close.

internal/statestore/redis.go
akremsa: yes, missed that. thank you

internal/statestore/backfill.go
scosgrave: 👍

akremsa:
@Laremere updated PR, please check.

1) I decided to store backfill and ticketIDs together. In this way, we avoid a case when ticketIDs are set successfully, but next call to set backfill fails (or vice versa) and there is no way to rollback because we agreed to use locks instead of transactions.
2) Added an additional internal protobuf structure to store backfill and ticketIDs.
3) Mutex is returned to the caller because otherwise it should be stored inside statestore between lock and unlock calls which I've found a bit problematic.

There should be the following flow using statestore:
```
statestore := New(cfg)
mutex := statestore.NewMutex("key1")
mutex.Lock()
statestore.ReleaseAllTickets(ctx)
statestore.UpdateBackfill(ctx, b, tids)
mutex.Unlock()
```

internal/statestore/backfill_test.go
```diff
@@ -0,0 +1,289 @@
+package statestore
```
aLekSer, line 1: License header.

api/messages.proto
```diff
@@ -261,6 +261,6 @@ message Backfill {
 
   // Generation gets incremented on GameServers update operations.
   // Prevents the MMF from overriding a newer version from the game server.
-  // It is not to be updated by the MMF.
```
aLekSer, line 264: It seems that you need to include `/api/backend.swagger.json`.

internal/statestore/redis.go
```diff
@@ -20,29 +20,48 @@ import (
...
+func (rb *redisBackend) NewMutex(key string) *redisLocker {
+	//TODO: make expiry duration configurable
+	m := redsync.NewMutex(key, rs.WithExpiry(5*time.Minute))
+	return &redisLocker{mutex: m}
+}
+
+// Lock locks r. In case it returns an error on failure, you may retry to acquire the lock by calling this method again.
+func (r redisLocker) Lock() error {
+	return r.mutex.Lock()
+}
```
aLekSer, line 58: While performing E2E tests on Update Backfill, found that context is needed here:
```suggestion
// Lock locks r. In case it returns an error on failure, you may retry to acquire the lock by calling this method again.
func (r redisLocker) Lock(ctx context.Context) error {
	return r.mutex.LockContext(ctx)
}
```
Otherwise nil pointer dereference in `func GetRedisPool(cfg config.View) *redis.Pool {` -`dialFunc`:
```
			if ctx.Err() != nil 
```

internal/statestore/backfill_test.go
```diff
@@ -0,0 +1,303 @@
...
+				require.NoError(t, err)
+			} else {
+				require.Error(t, err)
+				require.Equal(t, tc.expectedCode.String(), status.Convert(err).Code().String())
+				require.Contains(t, status.Convert(err).Message(), tc.expectedMessage)
+			}
+		})
+	}
+
+	// pass an expired context, err expected
```
Laremere, line 80: This is a different test, it belongs in its own test function.

Here,  and for the other backfill functions.

Laremere (changes requested):

internal/statestore/public.go
```diff
@@ -63,8 +68,22 @@ type Service interface {
...
+	GetBackfill(ctx context.Context, id string) (*pb.Backfill, []string, error)
+
+	// DeleteBackfill removes the Backfill with the specified id from state storage. This method succeeds if the Backfill does not exist.
+	DeleteBackfill(ctx context.Context, id string) error
+
+	// UpdateBackfill updates an existing Backfill with a new data. ticketIDs can be nil.
+	UpdateBackfill(ctx context.Context, backfill *pb.Backfill, ticketIDs []string) error
+
+	// NewMutex returns a new distributed mutex with given name
+	NewMutex(key string) *redisLocker
```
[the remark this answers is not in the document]
Laremere: Also, with a new interface, there's no need for the redisLocker type.

aLekSer:
`redis.go` file is big, so my comments got hidden:
Comment to this line:
```
func (r redisLocker) Lock() error {
```
I faced an issue while using this call in my development branch and test `TestDoDeleteBackfill`, test case "expect unavailable code since context is canceled before being called" which is a replica of existent `ticket` version.
It is failed while waiting for `Unavailable` error, when run Lock without a context:
https://github.com/aLekSer/open-match/blob/c6c7ecc702d3539f1f60d7413a75833577cc956c/internal/app/frontend/frontend_service_test.go#L319
Suggested version of Lock function:
```

// Lock locks r. In case it returns an error on failure, you may retry to acquire the lock by calling this method again.
func (r redisLocker) Lock(ctx context.Context) error {
	err := r.mutex.LockContext(ctx)
	if err != nil {
		return status.Errorf(codes.Unavailable, "Lock, failed to connect to redis: %v", err)
	}
	return err
}

```

internal/api/internal.proto
```diff
@@ -0,0 +1,29 @@
+// Copyright 2020 Google LLC
```
aLekSer, line 1: @akremsa Please take a look at conversation about this file naming, which might be `statestore.proto` : https://github.com/googleforgames/open-match/pull/1279#discussion_r525408856

internal/statestore/redis.go
akremsa: Nil pointer exceptionon appears because there is no `err != nil` check in dialFunc, but it should be there. I will add context anyway.

internal/statestore/redis.go
```diff
@@ -20,29 +20,48 @@ import (
...
 
+// redisLocker is a wrapper over redsync mutex in order to let callers use it but not to expose redsync package outside
+type redisLocker struct {
+	mutex *rs.Mutex
+}
+
+// NewMutex returns a new distributed mutex with given name
+func (rb *redisBackend) NewMutex(key string) *redisLocker {
+	//TODO: make expiry duration configurable
+	m := redsync.NewMutex(key, rs.WithExpiry(5*time.Minute))
```
aLekSer, line 51: @akremsa Can you add a prefix to the NewMutex:
```suggestion
	m := redsync.NewMutex("lock/"+key, rs.WithExpiry(5*time.Minute))
```

akremsa: Yes. 
The reason for this change is that miniredis store regular records and mutex keys in the same place and if we use the same key - backfill_id for both - miniredis can not acquire a lock during tests run.

aLekSer: I assume that similar thing would happen for normal Redis also.

akremsa: did some tests - redis behaves the same

internal/statestore/backfill_test.go
akremsa: @Laremere  Why? I use high-level names for tests where I include multiple test cases. 
For instance -TestCreateBackfill means that I test CreateBackfill under different conditions, and passing an expired context can be considered as another test case. 

Exactly the same happens here:
https://github.com/googleforgames/open-match/blob/master/internal/app/frontend/frontend_service_test.go#L45

Expire context scenario is included in a general TestDoCreateTickets function. In my case, I didn't add preAction, but tested expired context case at the end of the function.

If you want to extract expired context test case to its own function it should be done in all tests, not only in backfill_test.go, which can take some time I guess. What do you think?

aLekSer (approved):
Thanks for applying all comments.

Laremere (approved):

PR #1276, opened 2020-10-29
Add comments when displaying makefile usage

hsorellana:
Improve readability of the Makefile when executing "make" command and displaying comments for commands available.


Makefile
```diff
@@ -708,9 +737,15 @@ define fast_test_folder
 	$(foreach dir, $(wildcard $(1)/*/.), $(call fast_test_folder, $(dir)))
 endef
 
+## # Run go tests
+## make test
+##
 test: $(ALL_PROTOS) tls-certs third_party/
 	$(call test_folder,.)
 
+## # Run fast go tests
```
Laremere, line 746: Nit: This runs the same tests, a better description would be
Run go tests only with worse flake and race detection

Laremere (approved):

Makefile
hsorellana: fixed!

PR #1278, opened 2020-11-03
Add ability to filter backfills

yeukovichd:
This PR has changes that allow to filter tickets and backfills using search criteria


internal/testing/e2e/query_tickets_test.go
```diff
@@ -153,7 +153,10 @@ func returnedByQuery(t *testing.T, tc testcases.TestCase) (found bool) {
 	om := newOM(t)
 
 	{
-		resp, err := om.Frontend().CreateTicket(context.Background(), &pb.CreateTicketRequest{Ticket: tc.Ticket})
+		ticket := pb.Ticket{}
+		ticket.SearchFields = tc.SearchFields
```
Laremere, line 157: Nit:
```
ticket := &pb.Ticket{
  SearchFields: tc.SearchFields,
}
```
Here and below.

Prefer setting fields on struct during initialization like this if possible.

Laremere (commented):
Fantastic PR, this works really well!

One nit, then we're good to go.

internal/testing/e2e/query_tickets_test.go
yeukovichd: Done

Laremere (approved):

PR #1279, opened 2020-11-09
Frontend Create Backfill

aLekSer:
Initial version of CreateBackfill. It stores an empty regular TicketIds list.

Wait on StateStare changes.

### DIFF
--- a/internal/app/frontend/frontend_service.go
+++ b/internal/app/frontend/frontend_service.go
@@ -120,6 +120,7 @@ func doCreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest, store 
 	backfill.Id = xid.New().String()
 	backfill.CreateTime = ptypes.TimestampNow()
 
+
 	sfCount := 0
 	sfCount += len(backfill.GetSearchFields().GetDoubleArgs())
 	sfCount += len(backfill.GetSearchFields().GetStringArgs())
@@ -127,17 +128,9 @@ func doCreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest, store 
 	stats.Record(ctx, searchFieldsPerBackfill.M(int64(sfCount)))
 	stats.Record(ctx, totalBytesPerBackfill.M(int64(proto.Size(backfill))))
 
-	err := store.CreateBackfill(ctx, backfill)
+	err := store.CreateBackfill(ctx, backfill, []string{})
 	if err != nil {
 		return nil, err
-	}
-	// TODO: add IndexBackfill functionality
-	/*
-		err = store.IndexBackfill(ctx, ticket)
-		if err != nil {
-			return nil, err
-		}
-	*/
 
 	return backfill, nil
 }
--- a/internal/app/frontend/frontend_service.go
+++ b/internal/app/frontend/frontend_service.go
@@ -120,7 +120,6 @@ func doCreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest, store 
 	backfill.Id = xid.New().String()
 	backfill.CreateTime = ptypes.TimestampNow()
 
-
 	sfCount := 0
 	sfCount += len(backfill.GetSearchFields().GetDoubleArgs())
 	sfCount += len(backfill.GetSearchFields().GetStringArgs())
@@ -131,7 +130,7 @@ func doCreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest, store 
 	err := store.CreateBackfill(ctx, backfill, []string{})
 	if err != nil {
 		return nil, err
-
+	}
 	return backfill, nil
 }
 
@@ -235,7 +234,8 @@ func (s *frontendService) DeleteBackfill(ctx context.Context, req *pb.DeleteBack
 
 // GetBackfill fetches a Backfill object by its ID.
 func (s *frontendService) GetBackfill(ctx context.Context, req *pb.GetBackfillRequest) (*pb.Backfill, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+	bf, _, err := s.store.GetBackfill(ctx, req.GetBackfillId())
+	return bf, err
 }
 
 // UpdateBackfill updates a Backfill object, if present.
--- a/internal/app/frontend/frontend_service.go
+++ b/internal/app/frontend/frontend_service.go
@@ -94,6 +94,54 @@ func doCreateTicket(ctx context.Context, req *pb.CreateTicketRequest, store stat
 	return ticket, nil
 }
 
+// CreateBackfill creates a new Backfill object.
+// it assigns an unique Id to the input Backfill and record it in state storage.
+// A Backfill is considered as ready for matchmaking once it is created.
+//   - If SearchFields exist in a Backfill, CreateBackfill will also index these fields such that one can query the ticket with query.QueryBackfills function.
+func (s *frontendService) CreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest) (*pb.Backfill, error) {
+	// Perform input validation.
+	if req.Backfill == nil {
+		return nil, status.Errorf(codes.InvalidArgument, ".backfill is required")
+	}
+	if req.Backfill.CreateTime != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "backfills cannot be created with create time set")
+	}
+
+	return doCreateBackfill(ctx, req, s.store)
+}
+
+func doCreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest, store statestore.Service) (*pb.Backfill, error) {
+	// Generate an id and create a Backfill in state storage
+	backfill, ok := proto.Clone(req.Backfill).(*pb.Backfill)
+	if !ok {
+		return nil, status.Error(codes.Internal, "failed to clone input ticket proto")
+	}
+
+	backfill.Id = xid.New().String()
+	backfill.CreateTime = ptypes.TimestampNow()
+
+	sfCount := 0
+	sfCount += len(backfill.GetSearchFields().GetDoubleArgs())
+	sfCount += len(backfill.GetSearchFields().GetStringArgs())
+	sfCount += len(backfill.GetSearchFields().GetTags())
+	stats.Record(ctx, searchFieldsPerBackfill.M(int64(sfCount)))
+	stats.Record(ctx, totalBytesPerBackfill.M(int64(proto.Size(backfill))))
+
+	err := store.CreateBackfill(ctx, backfill)
+	if err != nil {
+		return nil, err
+	}
+	// TODO: add IndexBackfill functionality
+	/*
+		err = store.IndexBackfill(ctx, ticket)
+		if err != nil {
+			return nil, err
+		}
+	*/
+
+	return backfill, nil
+}
+
 // DeleteTicket immediately stops Open Match from using the Ticket for matchmaking and removes the Ticket from state storage.
 // The client must delete the Ticket when finished matchmaking with it.
 //   - If SearchFields exist in a Ticket, DeleteTicket will deindex the fields lazily.
@@ -184,11 +232,6 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
 // AcknowledgeBackfill is used to notify OpenMatch about GameServer connection info.
 // This triggers an assignment process.
 func (s *frontendService) AcknowledgeBackfill(ctx context.Context, req *pb.AcknowledgeBackfillRequest) (*pb.Backfill, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
-}
-
-// CreateBackfill creates a new Backfill object.
-func (s *frontendService) CreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest) (*pb.Backfill, error) {
 	return nil, status.Error(codes.Unimplemented, "not implemented")
 }
 
--- a/internal/app/frontend/frontend_service.go
+++ b/internal/app/frontend/frontend_service.go
@@ -100,6 +100,9 @@ func doCreateTicket(ctx context.Context, req *pb.CreateTicketRequest, store stat
 //   - If SearchFields exist in a Backfill, CreateBackfill will also index these fields such that one can query the ticket with query.QueryBackfills function.
 func (s *frontendService) CreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest) (*pb.Backfill, error) {
 	// Perform input validation.
+	if req == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "request is nil")
+	}
 	if req.Backfill == nil {
 		return nil, status.Errorf(codes.InvalidArgument, ".backfill is required")
 	}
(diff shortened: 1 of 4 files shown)


internal/app/frontend/frontend_service.go
```diff
@@ -94,6 +94,47 @@ func doCreateTicket(ctx context.Context, req *pb.CreateTicketRequest, store stat
...
+	sfCount += len(backfill.GetSearchFields().GetDoubleArgs())
+	sfCount += len(backfill.GetSearchFields().GetStringArgs())
+	sfCount += len(backfill.GetSearchFields().GetTags())
+	stats.Record(ctx, searchFieldsPerBackfill.M(int64(sfCount)))
+	stats.Record(ctx, totalBytesPerBackfill.M(int64(proto.Size(backfill))))
+
+	err := store.CreateBackfill(ctx, backfill, []string{})
+	if err != nil {
+		return nil, err
+
```
hsorellana, line 134: There's a missing closing curly brace here

aLekSer: That's true, I cannot run the test without state store changes.

internal/api/internal.proto
```diff
@@ -0,0 +1,29 @@
+// Copyright 2020 Google LLC
```
scosgrave, line 1: I think `statestore.proto` might be a more appropriate name for this file. It matches the style of using the module name in the proto file name that `synchronizer.proto` has started.

Laremere: messages.proto would also be appropriate, as we put all message definitions that aren't request/response types in a similarly named file in the public api.

Laremere: Oh that file is actually being added by https://github.com/googleforgames/open-match/pull/1273

scosgrave: My initial take on how this file should be named was not right. Laramere educated me on how these files are named. The naming pattern is not based on the module name. Instead, services and their req/resp usually go into their own proto file, and objects that aren't coupled to a service's API usually go in a messages.proto file. 
+1 for messages.proto

aLekSer: I will copy a following thread back to the author @akremsa

internal/app/frontend/frontend_service.go
```diff
@@ -94,6 +94,46 @@ func doCreateTicket(ctx context.Context, req *pb.CreateTicketRequest, store stat
...
+func doCreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest, store statestore.Service) (*pb.Backfill, error) {
+	// Generate an id and create a Backfill in state storage
+	backfill, ok := proto.Clone(req.Backfill).(*pb.Backfill)
+	if !ok {
+		return nil, status.Error(codes.Internal, "failed to clone input ticket proto")
+	}
+
+	backfill.Id = xid.New().String()
+	backfill.CreateTime = ptypes.TimestampNow()
```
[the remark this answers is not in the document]
aLekSer: This is an error to have not nil createTime:
```
		return nil, status.Errorf(codes.InvalidArgument, "backfills cannot be created with create time set")
```

internal/app/frontend/frontend_service.go
```diff
@@ -94,6 +94,46 @@ func doCreateTicket(ctx context.Context, req *pb.CreateTicketRequest, store stat
 	return ticket, nil
 }
 
+// CreateBackfill creates a new Backfill object.
+// it assigns an unique Id to the input Backfill and record it in state storage.
+// A Backfill is considered as ready for matchmaking once it is created.
+//   - If SearchFields exist in a Backfill, CreateBackfill will also index these fields such that one can query the ticket with query.QueryBackfills function.
+func (s *frontendService) CreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest) (*pb.Backfill, error) {
```
akremsa, line 101: why don't you check req for nil?

internal/app/frontend/frontend_service_test.go
```diff
@@ -274,3 +330,62 @@ func TestDoGetTicket(t *testing.T) {
...
+	for _, test := range tests {
+		test := test
+		t.Run(test.description, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(utilTesting.NewContext(t))
+			store, closer := statestoreTesting.NewStoreServiceForTesting(t, viper.New())
+			defer closer()
+
+			test.preAction(ctx, cancel, store)
+
+			backfill, l, err := store.GetBackfill(ctx, fakeBackfill.GetId())
```
akremsa, line 381: require.NotNil(t, backfill)

internal/testing/e2e/backfill_test.go
```diff
@@ -0,0 +1,52 @@
...
+		},
+	}}}
+	b1, err := om.Frontend().CreateBackfill(ctx, bf)
+	require.Nil(t, err)
+
+	b2, err := om.Frontend().CreateBackfill(ctx, bf)
+	require.Nil(t, err)
+
+	// Different Ids should be generated
+	assert.NotEqual(t, b1.Id, b2.Id)
```
akremsa, line 43: why do you mix assert and require? 
please use require

internal/testing/e2e/backfill_test.go
```diff
@@ -0,0 +1,52 @@
...
+	b1.Id = b2.Id
+	b1.CreateTime = b2.CreateTime
+	// Other than CreateTune abd Id fields, they should be equal
+	assert.Equal(t, b1, b2)
+	require.Nil(t, err)
+	get, err := om.Frontend().GetBackfill(ctx, &pb.GetBackfillRequest{BackfillId: b1.Id})
```
akremsa, line 49: please use common naming like **result** or **actual**, it will be easier to read this code.

internal/testing/e2e/backfill_test.go
```diff
@@ -0,0 +1,52 @@
...
+func TestCreateGetBackfill(t *testing.T) {
+	om := newOM(t)
+	ctx := context.Background()
+
+	bf := &pb.CreateBackfillRequest{Backfill: &pb.Backfill{SearchFields: &pb.SearchFields{
+		StringArgs: map[string]string{
+			"search": "me",
+		},
+	}}}
+	b1, err := om.Frontend().CreateBackfill(ctx, bf)
```
akremsa, line 36: require.NotNil(t, b1)

internal/testing/e2e/backfill_test.go
```diff
@@ -0,0 +1,52 @@
...
+	require.Nil(t, err)
+
+	b2, err := om.Frontend().CreateBackfill(ctx, bf)
```
akremsa, line 39: require.NotNil(t, b2)

internal/testing/e2e/backfill_test.go
```diff
@@ -0,0 +1,52 @@
...
+	require.Nil(t, err)
+
+	// Different Ids should be generated
+	assert.NotEqual(t, b1.Id, b2.Id)
+	b1.Id = b2.Id
+	b1.CreateTime = b2.CreateTime
+	// Other than CreateTune abd Id fields, they should be equal
```
akremsa, line 46: What does it mean? Could you rephrase it, please?

internal/app/frontend/frontend_service.go
```diff
@@ -94,6 +94,46 @@ func doCreateTicket(ctx context.Context, req *pb.CreateTicketRequest, store stat
...
+	// Perform input validation.
+	if req.Backfill == nil {
```
akremsa, line 103: Please test these req validations in e2e tests

internal/app/frontend/frontend_service_test.go
```diff
@@ -88,6 +88,62 @@ func TestDoCreateTickets(t *testing.T) {
...
+	for _, test := range tests {
+		test := test
+		t.Run(test.description, func(t *testing.T) {
+			store, closer := statestoreTesting.NewStoreServiceForTesting(t, cfg)
+			defer closer()
+
+			ctx, cancel := context.WithCancel(utilTesting.NewContext(t))
+			test.preAction(cancel)
+
+			res, err := doCreateBackfill(ctx, &pb.CreateBackfillRequest{Backfill: test.backfill}, store)
```
akremsa, line 135: require.NotNil(t, res)

akremsa: Why do you test doCreateBackfill instead of CreateBackfill?

akremsa (changes requested):
Some minor changes requested

internal/app/frontend/frontend_service.go
```diff
@@ -94,6 +94,46 @@ func doCreateTicket(ctx context.Context, req *pb.CreateTicketRequest, store stat
...
+	sfCount += len(backfill.GetSearchFields().GetDoubleArgs())
+	sfCount += len(backfill.GetSearchFields().GetStringArgs())
+	sfCount += len(backfill.GetSearchFields().GetTags())
+	stats.Record(ctx, searchFieldsPerBackfill.M(int64(sfCount)))
+	stats.Record(ctx, totalBytesPerBackfill.M(int64(proto.Size(backfill))))
+
+	err := store.CreateBackfill(ctx, backfill, []string{})
+	if err != nil {
+		return nil, err
+	}
```
hsorellana, line 133: Before returning, after my PR is merged, here should be a store.IndexBackfill call also

aLekSer: That's expected to have. I will add in a different PR.

internal/app/frontend/frontend_service_test.go
aLekSer: I am testing real handlers as part of E2E test.

internal/testing/e2e/backfill_test.go
aLekSer: I have updated this comment, but in a separate branch, sorry. I meant that all fields other then CreateTime and Ids are the same if we CreateBackfill with the same input args.

internal/app/frontend/frontend_service_test.go
```diff
@@ -88,6 +88,62 @@ func TestDoCreateTickets(t *testing.T) {
 	}
 }
 
+func TestDoCreateBackfill(t *testing.T) {
```
Laremere, line 91: While I think this follows the patterns of the testing for the other calls, I'm unsure about the value these tests are adding:

- I don't think calling with canceled context actually finds problems, at most if checks that one use of context is done correctly, but not that all uses are correct.
- It took a bit to get there, but I'm pretty confident that "preAction" is an anti-pattern at this point.  It's adding too much logic to tests.
- See also, the if err != nil check in the test code is definitely logic.  This should be removed, and is evidence of the tests being over-generalized.

My recommendation:
Get the test coverage using the e2e tests for:
- creating
- getting backfill created
- getting backfill which isn't present, and getting not found
(first two are already done, as far as I can tell)

At most, keep only the canceled context tests here, but I'm fine not having those.
The other tests that follow these patterns for existing features should probably be cleaned up, but that's out of scope for this project.

aLekSer: Thanks for your feedback, will address all comments

internal/app/frontend/frontend_service_test.go
akremsa: This is a unit test, it has nothing to do with e2e one. _doCreateBackfill_ is not an entry point to the current package. Please, test exported methods only, as you do in TestDoGetBackfill.

aLekSer: First of all we follow the same pattern for `TestDoCreateTicket()`, second Scott asked me to use E2E to cover those handlers:
https://github.com/googleforgames/open-match/pull/1279/files#r528961563

akremsa: If there is the same approach in TestDoCreateTicket it doesn't mean that it is done right. You should not rely on any other tests (e2e, etc) while working on unit tests. Here you **ignore** everything which comes before _doCreateBackfill_.

akremsa: @Laremere do you mean that we should test public API in e2e tests, rather than unit ones?
Because at this moment I don't see a clear boundary between tickets e2e and tickets unit tests, for instance.
If so - there is no point to add and support tests in _frontend_ package right now.

akremsa: With _TestDoCreateTickets_ half of a method is tested in frontend package, another part (validations) in e2e, which is a bit confusing.

akremsa (approved):

internal/app/frontend/frontend_service_test.go
```diff
@@ -88,6 +89,97 @@ func TestDoCreateTickets(t *testing.T) {
...
+						}}}},
+			expectedCode:    codes.OK,
+			expectedMessage: "",
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.description, func(t *testing.T) {
+			res, err := fs.CreateBackfill(ctx, tc.request)
```
akremsa, line 152: Ok, now it's better. But why do you retain TestDoCreateTickets now?

aLekSer:
Hello @Laremere I have applied all comments, if I did something wrong please leave feedback.

internal/app/frontend/frontend_service_test.go
```diff
@@ -88,6 +89,97 @@ func TestDoCreateTickets(t *testing.T) {
...
+			if tc.expectedCode == codes.OK {
```
Laremere, line 153: Avoid clauses like this in tests: instead this should be two separate test cases.

internal/app/frontend/frontend_service_test.go
```diff
@@ -88,6 +89,97 @@ func TestDoCreateTickets(t *testing.T) {
...
+				require.NotNil(t, res)
+			} else {
+				require.Error(t, err)
+				require.Equal(t, tc.expectedCode.String(), status.Convert(err).Code().String())
+				require.Contains(t, status.Convert(err).Message(), tc.expectedMessage)
+			}
+		})
+	}
+
+	// expect error with canceled context
```
Laremere, line 164: Should be in a separate test case.

internal/testing/e2e/backfill_test.go
```diff
@@ -0,0 +1,68 @@
...
+	// All fields other than CreateTime and Id fields should be equal
+	require.Equal(t, b1, b2)
+	require.Nil(t, err)
+	actual, err := om.Frontend().GetBackfill(ctx, &pb.GetBackfillRequest{BackfillId: b1.Id})
+	require.Nil(t, err)
+	require.NotNil(t, actual)
+	require.Equal(t, b1, actual)
+
+	// Get Backfill which is not present - NotFound
+	actual, err = om.Frontend().GetBackfill(ctx, &pb.GetBackfillRequest{BackfillId: b1.Id + "new"})
```
Laremere, line 64: I think this should be a separate test case?

Laremere (approved):
Adding a few comments, don't want to block this on nit for testing, so merging with comments.  Cleanup to match my desired state doesn't involve only the new tests here, so it's fine.

internal/testing/e2e/backfill_test.go
aLekSer: OK, I will make it as a separate PR.

Issue #1240, opened 2020-07-29
Feature Request: Backfill Support


calebatwd:
**Is your feature request related to a problem? Please describe.**
Matches can be created without being full. At the same time, matches that were “full,” can have changing rosters from the perspective of the game host. Developers need a way to fill partially full matches. Additionally, the reasons and criteria used to fill matches change over time as the game progresses (e.g. highly skilled players leave one team, a friend has joined from outside matchmaking, etc).

For users of the matchmaker, solving this backfill problem should result in quicker times to match, and matches that are filled with more players; it will be more likely that any joined server will be full or quickly filled. From a developer perspective, it is very hard to work with a backfill system that competes with tickets being allocated to new servers. Methods to achieve this generally take the form of introducing lag to compensate for startup delay of servers, and failure to do so correctly can break matchmaking or overwhelm the system.

Here are a few common issues associated with directly running a match function as a backfill request:

**Backfill Storms** - If backfilling encounters issues that prevent successful backfill for a short period, matchmaking generates new game hosts since players aren't being drained into existing games. The game hosts start and try backfilling, resulting in further failure. More game hosts. More failed backfills. If left unresolved, the issue balloons and eventaully becomes a scale problem. Etc. Etc. This is especially problematic for long running game hosts (like lobby servers or mmos)

**Immediacy Problem** - Servers scheduled for allocation need to be immediately available for backfill. Without this, during the lag between a server allocation request and it coming online for backfill requests, many extraneous servers can be created leading to a cascading effect that can lead to service failure and extremely sparse server density.

**Player Density** - Similarly to the immediacy problem above, backfill requests that aren’t processed during every match function window are missed and can result in new matches being created despite having partially full games. This decreases the ability for the developer to fill those games, and ultimately ties game-host player-density directly to the rate at which a game developer is able to meet the matchmaking sync windows. This necessitates the need for some state to exist in the matchmaker to be available during matchmaking windows on a consistently recurring basis.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Laremere:
Previous discussion on backfill:
https://github.com/googleforgames/open-match/issues/1149
https://github.com/googleforgames/open-match/issues/1148
https://github.com/googleforgames/open-match/issues/1114

None of them got very far, so closing them in favor of this discussion which will be more targeted at finding the solution and implementing changes to OM.

scosgrave:
[...]

PR #1280, opened 2020-11-11
Query backfills

hsorellana:
This PR is WIP which adds the ability to query backfill objects. Still needs a lot of work.

Don't pay attention to Frontend and Statestore func additions, since they are unrelated to this PR but i needed some functionality from those services to get QueryService working

This is related to [Backfill](https://github.com/googleforgames/open-match/issues/1240) feature.


Laremere:
We need to be caching results here, and not loading only all backfills for every query.  So we need a backfill cache like we have a tickets cache.

Some context: https://github.com/googleforgames/open-match/issues/1123

We used to use in redis filtering for tickets, and the performance was frankly awful.  As the number of queries goes up, the load on redis (which is single threaded!) goes up linearly.  By joining multiple query calls into one cache update, the load only scales relative to the number of query service instances.  Additionally, it reduces the number of times a given ticket/backfill is loaded from redis (per version) to only once per query instance, instead of once per query.

internal/app/query/query_service.go
```diff
@@ -135,7 +135,48 @@ func (s *queryService) QueryTicketIds(req *pb.QueryTicketIdsRequest, responseSer
 }
 
 func (s *queryService) QueryBackfills(req *pb.QueryBackfillsRequest, responseServer pb.QueryService_QueryBackfillsServer) error {
```
akremsa, line 137: what about checking req for nil?

hsorellana: Here my grpc knowledge escapes me, but can a nil request reach this handler??

akremsa: Is this method supposed to be directly called by a grpc-client?

internal/app/query/query_service.go
```diff
@@ -135,7 +135,48 @@ func (s *queryService) QueryTicketIds(req *pb.QueryTicketIdsRequest, responseSer
...
+				results = append(results, backfill)
+			}
+		}
+	})
+	if err != nil {
+		err = errors.Wrap(err, "QueryBackfills: failed to run request")
+		return err
+	}
+
+	stats.Record(ctx, backfillsPerQuery.M(int64(len(results))))
```
akremsa, line 162: I think it's worth putting _len(results)_ to a separate variable. You call it 4 times in this method.

internal/app/query/query_service.go
hsorellana: Yes! Now i see your point... 👍

internal/app/query/query_service.go
hsorellana: Fixed

aLekSer:
We have discuss a need to make an update to the design: include Backfill `Version` field.
Point to have it: easy way to invalidate Backfill updates from the backend, when the result of the MMF (outdated Backfill queried) has a previous version and a resulting match proposal should not be taken into account, as an attempt to Update Backfill with an outdated version in it would be rejected.

Imagine there is a version 1 `Backfill` sits in the query. It was used to form a matchProposal and lead to updated `Backfill`. Redis has a `Backfill` version 2 now.
Next round of MMF for some reason `queryBackfills` and received old version 1 `Backfill`, form a matchProposal but this match even if it passes Evaluator. An update of the `Backfill` from 1->2nd version would be rejected, since Redis already know version 2. So no Tickets would be associated with this `Backfill`.

Two options to achieve the aforementioned behaviour:
1. Include  `Version` along with `Generation` into `Backfill` protobuf definition. It would also be contained in the matchProposal as part of the `Match`
1. Include `Version` to the InternalBackfill representation also stored in Redis. (Along with `Backfill`, `TicketIds` - `Version` field). This leads to a need to change `Match` protobuf  - new field `backfill_version`, along with current `Backfill` and `allocate_gameserver`

hsorellana:
ping @Laremere ☝️

internal/testing/e2e/query_tickets_test.go
```diff
@@ -150,16 +150,14 @@ func TestTicketNotFound(t *testing.T) {
...
-		Name:         "OwnTest",
-		Pool:         &pb.Pool{},
-		SearchFields: nil,
+	for _, tc := range testcases.IncludedTestCases() {
+		tc := tc
+		t.Run("QueryBackfill/"+tc.Name, func(t *testing.T) {
+			if !returnedBackfillByQuery(t, tc) {
+				require.Fail(t, "Expected to find backfill in pool but didn't.")
+			}
+		})
```
hsorellana, line 159: Hey @Laremere i could use some help here. I'm using the currently existing test cases for querying backfills. Do you think are useful scenarios which doesn't provide false positives?? Or should we revisit this??

internal/app/query/query_service.go
```diff
@@ -325,3 +370,151 @@ func (tc *ticketCache) update() {
...
+	for id := range bc.backfills {
+		if _, ok := currentAll[id]; !ok {
+			delete(bc.backfills, id)
+			deletedCount++
+		}
+	}
+
+	toFetch := []string{}
+	for id := range currentAll {
+		if _, ok := bc.backfills[id]; !ok {
```
aLekSer, line 499: Here we don't include all Backfills and updated Backfills
```
		if b, ok := bc.backfills[id]; !ok || b.Version != id.Version {
```

internal/statestore/redis.go
```diff
@@ -637,3 +639,112 @@ func (rb *redisBackend) newExponentialBackoffStrategy() backoff.BackOff {
...
+
+// IndexBackfill adds the backfill to the index.
+func (rb *redisBackend) IndexBackfill(ctx context.Context, backfill *pb.Backfill) error {
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		return status.Errorf(codes.Unavailable, "IndexBackfill, id: %s, failed to connect to redis: %v", backfill.GetId(), err)
+	}
+	defer handleConnectionClose(&redisConn)
+
+	err = redisConn.Send("SADD", allBackfills, backfill.Id)
```
aLekSer, line 718: I think we should switch to `hset backfill-to-version backfillID 32` or to something like next:
```
	err = redisConn.Send("SREM", allBackfills, backfill.Id+"/"+string(backfill.Version-1))
	err = redisConn.Send("SADD", allBackfills, backfill.Id+"/"+string(backfill.Version))
```

internal/statestore/redis.go
```diff
@@ -637,3 +639,112 @@ func (rb *redisBackend) newExponentialBackoffStrategy() backoff.BackOff {
...
+
+	return nil
+}
+
+// DeindexBackfill removes specified Backfill from the index. The Backfill continues to exist.
+func (rb *redisBackend) DeindexBackfill(ctx context.Context, id string) error {
+	return nil
+}
+
+func (rb *redisBackend) GetIndexedBackfills(ctx context.Context) (map[string]struct{}, error) {
```
aLekSer, line 732: This function should return pairs: BackfillIds, and Versions.
Can we move Index to another PR?

internal/statestore/redis.go
aLekSer: Please include `TestGetIndexedIDSet()` in your smaller PR

aLekSer:
I vote for splitting this PR into at least 2 parts. First part should include `IndexBackfill` which would be used on Create and Update Backfills.
We need to introduce `version` for Backfills as well as use `hset` or more complex keys in set to include version, and remove old versions on `IndexBackfill`.
To switch in `query_service.go` to:
```
		if b, ok := bc.backfills[id]; !ok || b.Version != id.Version {
			toFetch = append(toFetch, id)
		}
```

internal/app/query/query_service.go
```diff
@@ -270,11 +316,10 @@ collectAllWaiting:
 	// Send WaitGroup to query calls, letting them run their query on the ticket
 	// cache.
 	for _, req := range reqs {
-		tc.wg.Add(1)
 		select {
 		case req.runNow <- struct{}{}:
+			tc.wg.Add(1)
 		case <-req.ctx.Done():
-			tc.wg.Done()
```
aLekSer, line 277: It might be more safe to move this into a separate small PR.

internal/app/query/query_service.go
aLekSer: Now this is covered by line above:
```
for id, cachedBackfill := range bc.backfills {
 		indexedBackfillGeneration, ok := indexedBackfills[id]
 		if !ok || cachedBackfill.Generation < int64(indexedBackfillGeneration) {
```

aLekSer (commented):
Now tests are fine. Please roll back TicketCache and move this to another PR. So that we don't update the core logic.

aLekSer (approved):
Approving the PR. The main functionality is working now. Tests are flaky at the moment, but most of the time passes locally. Let's skip them and unblock @yeukovichd MMF and Backend Changes.

aLekSer:
@Laremere Scott can you please review this PR it is needed to proceed with MMF changes.

internal/app/query/query_service.go
```diff
@@ -135,7 +135,53 @@ func (s *queryService) QueryTicketIds(req *pb.QueryTicketIdsRequest, responseSer
...
+	}
+
+	pool := req.GetPool()
+	if pool == nil {
+		return status.Error(codes.InvalidArgument, ".pool is required")
+	}
+
+	pf, err := filter.NewPoolFilter(pool)
+	if err != nil {
+		return err
```
akremsa, line 149: Should it be an exact status code like in the return statements above? If so, please check other returns in this method.

aLekSer: `QueryTickets` does not have. But we can make it better.

aLekSer: I assume nothing to be done. `NewPoolFilter` already included those `codes`.

aLekSer:
TestBackfillNotFound is [failing](https://00f74ba44bd2d4df0ececce404f6a2876b0b69d7d5-apidata.googleusercontent.com/download/storage/v1/b/open-match-build-logs/o/log-fb73c1e6-3843-4dcd-8bd6-90a98252f6bb.txt?jk=AFshE3U1qi6XzVWCzSg0cZfiY33DKqyfHo2mfu-C_IFoUfWVirQ13lHygiVx3ARrR_nhc4zbGi55TFcBKP3jSY3LlvikhaNJ1EbNTWyxmkFipfi4Xq6ze1qsEMNxVYR89BLfPVyS6EqVIaBJIqTwEkNlykos2-4-CbBua-OugEbFrDa2wXVwYGcQOZZcYzhRunDSwpeBqc_IoK_y3xJJGy4oRREuaBt1LmK2Ljj7nnjsq0k4Ij4wDIsP-HUeEzqh4rSkbzIvdNVSwiCHJrYTpNJ3VRQDiCjon3SiuYETyVBk2VtBJHAMuLPn_xfgmiCFrRX3lB0ItZVX_YoVRag_6qXgX6fmVhTApNOnROryHxdTczx7ZAVlwuZ5Xyp_OH_DAvwFSe70vT9fLTCMYzxQVYDZKQtlbQE6ZXQH4STWu4L-ZGH021S_hJjdZ2Y8XmMplG2uO_pvi-V2Dy6vxYFBj4ZJ1rM4FXCXXfm-Ey-V3mSHP-hRqOGTZ4C4E7FoLgT19ad3jmyIX0o_RMYlQLyhKps4CPmp1Rjz-U7UOYMMNbLKa7kMp_MRB8XMIJdxzVQ3scV0_vuP3U23Exaf4Lw1QhJ-9lif1eHSDOHwznyDLXL_5Qr4u8ZmBPI5oUD6DOwKzVN2TfBT7e2c5t_qYEcRlOaZ8cVOUhrGJrl6nqdnpmxo7q2WLv99o6VRgB3AUMPGitmLI5AdBuPBMxoOEq5227vBHOjc4eF55RhPr5QCxG45msE63wlzupFCsA5AFYK7tN8GbQwOwdWYdA8SLlhMHPX086Ynfx4q5BI8nsv8wM1Py1lBL9HCX1UxOOlbT0QmUPlijQJ6-8rJeLDhIEDSRTAmNB9r2bmVqx23nHRryieeEFtZcTIeDZbpGzw0Fe9gjJC3RRZUPkBQegngzRlLkRJyYWD3Pq31mqovQiXw&isca=1) (hard to reproduce locally): 
`
Step #8 - "Test: Services": --- FAIL: TestBackfillNotFound (3.44s)
Step #8 - "Test: Services":     --- FAIL: TestBackfillNotFound/QueryBackfill/double_range_minMaxAreNan#02 (0.06s)
Step #8 - "Test: Services":         query_tickets_test.go:262: 
Step #8 - "Test: Services":             	Error Trace:	query_tickets_test.go:262
`

internal/app/query/query_service_test.go
```diff
@@ -66,3 +70,58 @@ func TestGetPageSize(t *testing.T) {
...
+			Id:         "backfill-01",
+			Generation: 1,
+		}
+		bf2 := &pb.Backfill{
+			Id:         "backfill-02",
+			Generation: 1,
+		}
+		ctx := context.Background()
+		storeAndIndex(ctx, store, bf1, bf2)
+		bfCache.update()
```
akremsa, line 108: add `require.Equal(t, 2, len(bfCache.backfills))`?

internal/app/query/query_service.go
akremsa: Yes, but what should be done with other returns down below?

internal/app/query/query_service.go
```diff
@@ -42,8 +41,9 @@ var (
 // queryService API provides utility functions for common MMF functionality such
 // as retreiving Tickets from state storage.
 type queryService struct {
-	cfg config.View
-	tc  *ticketCache
+	cfg     config.View
+	tc      *ticketCache
```
akremsa, line 45: probably rename to ticketCache?

aLekSer: or `tCache`. We don't want to change that, because we didn't want to change Ticket core logic and produce some bugs. So let's stay as is.

internal/app/query/query_service.go
```diff
@@ -325,3 +371,146 @@ func (tc *ticketCache) update() {
...
+		if !ok || cachedBackfill.Generation < int64(indexedBackfillGeneration) {
+			delete(bc.backfills, id)
+			deletedCount++
+		}
+	}
+
+	toFetch := []string{}
+	for id := range indexedBackfills {
+		if _, ok := bc.backfills[id]; !ok {
+			toFetch = append(toFetch, id)
```
akremsa, line 500: Why can't you call bc.store.GetBackfill right here, instead of iterating through toFetch down below? We can avoid 1 extra for loop.

aLekSer: Here could be a TODO or something, in order to keep this similar to runRequest of Tickets. Later on, we can optimise it.

aLekSer: Tests are flaky even without this enhancement. The main difference is `hset` vs  usual `set`.

internal/app/query/query_service.go
akremsa: It is used just in 3 places and it won't cause bugs because it won't compile if something goes wrong.
Just noticed a different naming style for backfill and ticket cache.

internal/app/query/query_service.go
hsorellana: The NewPoolFilter returns status.Error(...) if some error happens, so here that error it is only returned

internal/app/query/query_service.go
hsorellana: Here i have some doubts because, if we keep the same behavior from TicketCache, there are some metrics i need to implement and one of them uses the `toFetch` variable. So on one hand we could do what you say Alexey, or we keep the same metrics as for tickets. 
cc: @Laremere

internal/app/query/query_service.go
```diff
@@ -325,3 +371,146 @@ func (tc *ticketCache) update() {
...
+		}
+	}
+
+	for _, backfillToFetch := range toFetch {
```
Laremere, line 504: I'm worried this loop will be noticeably slow, because it could potentially make many separate requests to redis.  Especially true because it's expected for backfills to change a lot and need to be refreshed.  For tickets, there's a bulk get which is done using a single redis call.  I'm not opposed to merging it in this state, and adding that later (as long as it's high priority to add.)

internal/app/query/query_service.go
```diff
@@ -325,3 +371,146 @@ func (tc *ticketCache) update() {
...
+
+	// This error is used for the return value when a request fails to reach
+	// the storage. In case of multiple requests are being handled, this error
+	// is returned to every single request.
+	err error
+}
+
+func newBackfillCache(b *appmain.Bindings, cfg config.View) *backfillCache {
+	bc := &backfillCache{
+		store:           statestore.New(cfg),
```
Laremere, line 399: Both caches should share the same statestore, move this to be an argument to the newCache, and create it once.

internal/app/query/query_service.go
```diff
@@ -325,3 +371,146 @@ func (tc *ticketCache) update() {
 	logger.Debugf("Ticket Cache update: Previous %d, Deleted %d, Fetched %d, Current %d", previousCount, deletedCount, len(toFetch), len(tc.tickets))
 	tc.err = nil
 }
+
+type backfillCache struct {
```
Laremere, line 375: Simply copy pasting here created a lot of duplicated code, which I mostly don't like because the common part is tricky concurrency stuff.

I think the best solution here is to refactor the ticketCache into simply `cache`.  Include in the struct (and `newCache`) an update function with type `func(store statestore.Service, state interface{}) error`, moving both update functions to static functions which can be passed into the constructor.  You'll need a default value for the state as well.  

Then change f in request to be  ` f func(interface{})`.  For the update and request functions/closures, simply do a cast.

It's not pretty, and generics would be really nice here.  However all of the casts will trivially be tested by the e2e tests, so I'm not really worried about a bug there.  I think the cost of a few casts is worth de-duplicating the concurrency code.
Thoughts?

Laremere (changes requested):

internal/app/query/query_service_test.go
```diff
@@ -66,3 +70,59 @@ func TestGetPageSize(t *testing.T) {
 		})
 	}
 }
+
+func TestBackfillCache(t *testing.T) {
```
Laremere, line 74: Honestly, I'd skip testing here (as I did with tickets), and just make sure all these cases are covered by e2e.

internal/testing/e2e/query_tickets_test.go
Laremere: I'm not sure what you're asking?  These tests seem good to me.

internal/app/query/query_service_test.go
Laremere: Specifically, the "call create and see backfill in query, then call update and the backfill in query is updated" test case on e2e is super important.  Be good to also check with delete properly making it go away.  (Do we have that test case for tickets?  If not we should...)

Doing those tests would make these tests redudant.

internal/app/query/query_service.go
aLekSer: I also saw lots of duplicate code here, just thought it might be united into one set of functions later on at the refactoring stage. In order to leave beta (backfill) and release (ticket) code separate.
Overall there is only one condition which is changed, it is comparing Generation number in a memory (map) and in HSET in Redis.

internal/app/query/query_service_test.go
hsorellana: applied!

internal/testing/e2e/query_tickets_test.go
hsorellana: I was asking if current test cases are useful for testing this. But if you think the test cases are good for querying backfills, i'll leave them untouched

internal/statestore/redis.go
hsorellana: Already implemented

internal/app/query/query_service.go
hsorellana: I standarized the field naming as it always have been

internal/app/query/query_service.go
```diff
@@ -135,7 +135,53 @@ func (s *queryService) QueryTicketIds(req *pb.QueryTicketIdsRequest, responseSer
...
+	err = s.bfc.requestBackfills(ctx, func(backfills map[string]*pb.Backfill) {
+		for _, backfill := range backfills {
+			if pf.In(backfill) {
+				results = append(results, backfill)
+			}
+		}
+	})
+	if err != nil {
+		err = errors.Wrap(err, "QueryBackfills: failed to run request")
+		return err
```
akremsa, line 163: @hsorellana I mean should you return here some status codes?

hsorellana: ouh you're right!

internal/app/query/query_service.go
Laremere: Some further examples to clarify:

```golang
type ticketCache struct {
	store statestore.Service

	requests chan *cacheRequest

	// Single item buffered channel.  Holds a value when runQuery can be safely
	// started.  Basically a channel/select friendly mutex around runQuery
	// running.
	startRunRequest chan struct{}

	wg sync.WaitGroup

        update func(statestore.Service, interface{}) error

	// Mutlithreaded unsafe fields, only to be written by update, and read when
	// request given the ok.
        value interface{}
	err     error
}
```


```golang
func (s *queryService) QueryTicketIds(req *pb.QueryTicketIdsRequest, responseServer pb.QueryService_QueryTicketIdsServer) error {
       /// <code removed here for example simplicity>
	var results []string
	err = s.tc.request(ctx, func(value interface{}) {
                tickets := value.(map[string]*pb.Ticket)
		for id, ticket := range tickets {
			if pf.In(ticket) {
				results = append(results, id)
			}
		}
	})
	if err != nil {
		err = errors.Wrap(err, "QueryTicketIds: failed to run request")
		return err
	}
       /// <code removed here for example simplicity>
}
```

```golang
func updateTicketCache(store statestore.Service, value interface{}) error {
  tickets := value.(map[string]*pb.Ticket)
   /// etc....
}
```

aLekSer: Covered by this PR #1300 . We can close this PR I think.

PR #1287, opened 2020-11-17
Improve comments for RPC funcs

hsorellana:
This PR adds some modifications to the comments of the Backfill funcs inside the FrontendService since they weren't so easy to read and understand its usage.


api/frontend.proto
```diff
@@ -111,8 +111,7 @@ message GetBackfillRequest {
 // BETA FEATURE WARNING: This Request message is not finalized and still subject
 // to possible change or removal.
 message UpdateBackfillRequest {
-  // A Backfill object with SearchFields defined.
-  Backfill backfill_ticket = 1;
+  Backfill backfill = 1;
```
aLekSer, line 114: I like this more as it is consistent with `CreateBackfillRequest`. 👍

api/frontend.swagger.json
```diff
@@ -141,9 +125,9 @@
         ]
       },
       "delete": {
-        "summary": "Side effects: Any tickets waiting for this backfill will be returned to the active pool, no longer pending.",
+        "summary": "DeleteBackfill receives a backfill ID and deletes its resource.\nAny tickets waiting for this backfill will be returned to the active pool, no longer pending.",
         "description": "BETA FEATURE WARNING:  This call and the associated Request and Response\nmessages are not finalized and still subject to possible change or removal.",
-        "operationId": "FrontendService_DeleteBackfill",
```
aLekSer, line 146: Not sure, there are so many changes to `swagger.json` files. Do you use a different version of it?

aLekSer (commented):
Changes in `proto` file is good. A bit worried of why `swagger.json` so significantly changed. Can you test this with `kind` and `make proxy-ui` to see if it looks good on `http://localhost:51500/`:
```
make create-kind-cluster
make get-kind-kubeconfig
make push-images
make install-chart
```

hsorellana:
i used an online swagger editor to verify how the output json file looks like. And it looks like 
Screen Shot 2020-11-17 at 18 55 37

aLekSer:
Tried to generate `swagger.json` files, they are the same as in this PR.

api/frontend.proto
```diff
@@ -186,15 +186,19 @@ service FrontendService {
...
   // messages are not finalized and still subject to possible change or removal.
   rpc GetBackfill(GetBackfillRequest) returns (Backfill) {
     option (google.api.http) = {
       get: "/v1/frontendservice/backfills/{backfill_id}"
     };
   }
-
-  // Side effects: Any tickets waiting for this backfill will be returned to the active pool, no longer pending.
+  
+  // UpdateBackfill updates a Backfill object with an ID set and SearchFields defined.
```
Laremere, line 199: extensions will be updated to.  Perhaps something like "UpdateBackfill updates search_fields and extensions for the backfill with the provided id"?

hsorellana: Applied!

pkg/pb/evaluator.pb.go
```diff
@@ -1,288 +1,160 @@
...
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
 // Code generated by protoc-gen-go. DO NOT EDIT.
-// versions:
-// 	protoc-gen-go v1.25.0
-// 	protoc        v3.10.1
```
hsorellana, line 18: Look here @Laremere. When i run `make all-protos` to generate the changes from proto files, it deletes the licensing here. How should we proceed with this?? Do i leave it like it is or what?? Little help, please

Laremere: If that's what the commands do - let's leave it like that.

Laremere (approved):

pkg/pb/evaluator.pb.go
hsorellana: ok, great!

aLekSer (approved):
Looks nice.

PR #1288, opened 2020-11-19
Redis: Backfill last acknowledged

aLekSer:
Add SortedSet to store Backfill in a sorted order to retrieve all expired Backfills.

Work on #1240 

This PR is dependent on StateStore changes.
We can select a specific TTL config variable later during implementation of Cleanup Service.

```
func (rb *redisBackend) AcknowledgeBackfill(ctx context.Context, id string) (error) {
func (rb *redisBackend) GetExpiredBackfills(ctx context.Context) ([]string, error) {
// Used in DeleteBackfil()
func (rb *redisBackend) deleteExpiredBackfillID(ctx context.Context, backfillID string) error
```


hsorellana:
Hey @aLekSer, look at this. "Please have them confirm that by leaving a comment that contains only `@googlebot I consent`. in this pull request"

aLekSer:
@hsorellana that would not be an issue once Alexey's PR would be merged. Then this PR would have only my changes.

internal/statestore/backfill.go
```diff
@@ -134,3 +139,70 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+	expiredBackfillIds, err := redis.Strings(redisConn.Do("ZRANGEBYSCORE", backfillLastAckTime, startTimeInt, endTimeInt))
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error getting expired backfills %v", err)
+	}
+
+	return expiredBackfillIds, nil
+}
+
+// DeleteExpiredBackfillIDs - delete expired BackfillIDs from a sorted set
+func (rb *redisBackend) DeleteExpiredBackfillIDs(ctx context.Context, backfillIDs []string) error {
```
Laremere, line 189: It's not clear to me how this will be used?

internal/statestore/backfill.go
```diff
@@ -134,3 +139,70 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+// GetExpiredBackfillIDs - get all backfill IDs which are expired
+func (rb *redisBackend) GetExpiredBackfillIDs(ctx context.Context) ([]string, error) {
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Unavailable, "GetExpiredBackfillIDs, failed to connect to redis: %v", err)
+	}
+	defer handleConnectionClose(&redisConn)
+
+	// the same TTL is used for both Backfills and pendingRelease Tickets
+	ttl := rb.cfg.GetDuration("pendingReleaseTimeout")
```
Laremere, line 174: This should be something like 80% to 90% of ttl, since the backfill should expire before the tickets waiting on it, as the tickets timing out but the backfill NOT timing out but being acknowledged would be a problem.

internal/statestore/backfill.go
aLekSer: It would be used by a cleanup service:
1st step `backfillIds, err := GetExpiredBackfills()`
2nd - Return associated tickets for all Bakcfills back to the pool.
Some AcknowledgeBackfill() could occur in between 
3rd step `DeleteExpiredBackfillIDs(backfillIds)`, note that there could be more expired Backfills but we only delete thos e processed.

internal/statestore/backfill.go
```diff
@@ -134,3 +139,70 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+
+// AcknowledgeBackfill - store Backfill's last acknowledgement time
+func (rb *redisBackend) AcknowledgeBackfill(ctx context.Context, id string) error {
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		return status.Errorf(codes.Unavailable, "AcknowledgeBackfill, id: %s, failed to connect to redis: %v", id, err)
+	}
+	defer handleConnectionClose(&redisConn)
+
+	currentTime := time.Now().UnixNano()
```
akremsa, line 151: @Laremere is there an agreement of not using UTC()? `time.Now()` is used everywhere.

Laremere: Assumption is all computers in a cluster will have clocks that are synchronized enough.

internal/statestore/backfill.go
```diff
@@ -134,3 +139,70 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+
+// DeleteExpiredBackfillIDs - delete expired BackfillIDs from a sorted set
+func (rb *redisBackend) DeleteExpiredBackfillIDs(ctx context.Context, backfillIDs []string) error {
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		return status.Errorf(codes.Unavailable, "DeleteExpiredBackfillIDs, failed to connect to redis: %v", err)
+	}
+	defer handleConnectionClose(&redisConn)
+
+	cmds := make([]interface{}, 0, len(backfillIDs)+1)
```
akremsa, line 196: good one

internal/statestore/backfill.go
Laremere: Why not just use the same mechanics as the delete method?  Reduces implementation complexity and possible bugs.

internal/statestore/backfill_test.go
```diff
@@ -301,3 +302,54 @@ func TestDeleteBackfill(t *testing.T) {
 	require.Equal(t, codes.Unavailable.String(), status.Convert(err).Code().String())
 	require.Contains(t, status.Convert(err).Message(), "DeleteBackfill, id: 12345, failed to connect to redis:")
 }
+
+func TestAcknowledgeBackfill(t *testing.T) {
```
akremsa, line 306: Actually, it's not an _AcknowledgeBackfill_ unit test now. You test here the whole acknowledge scenario. This is nice, but I would call this test something like _TestAcknowledgeLifecycle_. 
To write a unit test you need to test exactly one single unit (method) under the different conditions. 
Talking about backfill it's enough to check that after calling _AcknowledgeBackfill_ the passed ID is added to _backfill_last_ack_time_.
So I suggest you to rename this method and add _TestAcknowledgeBackfill_ where you will test only _AcknowledgeBackfill_ method. What do you think?

internal/statestore/public.go
```diff
@@ -84,6 +84,15 @@ type Service interface {
 
 	// NewMutex returns an interface of a new distributed mutex with given name
 	NewMutex(key string) RedisLocker
+
+	// AcknowledgeBackfill - store Backfill's last accessed time
+	AcknowledgeBackfill(ctx context.Context, id string) error
```
akremsa, line 89: nit: I suggest sticking to the current comments format like:
_AcknowledgeBackfill stores Backfill's last accessed time_

internal/statestore/backfill.go
```diff
@@ -134,3 +139,70 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+	defer handleConnectionClose(&redisConn)
+
+	currentTime := time.Now().UnixNano()
+	cmds := make([]interface{}, 0)
+	cmds = append(cmds, backfillLastAckTime, currentTime, id)
+
+	_, err = redisConn.Do("ZADD", cmds...)
+	if err != nil {
+		err = errors.Wrap(err, "failed to store backfill's last acknowledgement time")
+		return status.Error(codes.Internal, err.Error())
```
akremsa, line 158: nit: I think here you can use _status.Errorf(...., err)_ as you do above. Probably we need to elaborate the same error logging format across this file.

internal/statestore/backfill.go
aLekSer: Some DeleteBackfill() request can have some issues, others no, so Backend have an opportunity to select on its own.

internal/statestore/backfill.go
```diff
@@ -134,3 +139,70 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+	}
+	defer handleConnectionClose(&redisConn)
+
+	// the same TTL is used for both Backfills and pendingRelease Tickets
+	ttl := rb.cfg.GetDuration("pendingReleaseTimeout")
+	curTime := time.Now()
+	endTimeInt := curTime.Add(-ttl).UnixNano()
+	startTimeInt := 0
+
+	// Filter out tickets that are fetched but not assigned within TTL time (ms).
```
akremsa, line 179: Filter out tickets backfill ids ...

internal/statestore/backfill.go
aLekSer: Got an idea now. This can be done as part of `DeleteBackfill`

internal/statestore/public.go
aLekSer: // AcknowledgeBackfill - stores Backfill's last acknowledgement time

aLekSer:
@Laremere Scott, note that I would add config parameter for Backfill Time in a separate PR to reduce complexity of this PR. It is ready to review, hope that I got all comments right.

Laremere:
Force push is adding more work to re-review.  Friendly reminder to not do that (otherwise "see changes since your last view" breaks.)
If you're having trouble pushing (eg, you used update branch button at some point), then do `git pull` before you push.

aLekSer:
@Laremere Do you mean that we should not perform rebase on a new master also? I get used to do it on Agones.

internal/statestore/backfill_test.go
```diff
@@ -300,4 +309,130 @@ func TestDeleteBackfill(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Check that Acknowledge timestamp stored valid in Redis
+	pool := GetRedisPool(cfg)
+	conn := pool.Get()
+	res, err := redis.Int64(conn.Do("ZSCORE", backfillLastAckTime, bf1))
+	require.NoError(t, err)
+	require.True(t, (time.Unix(res/1e9, res%1e9).Sub(curTime)) < time.Second)
+
+	cfg = createInvalidRedisConfig()
```
akremsa, line 399: This is a completely different test case. Please move it to another test function or add table-driven tests here.

internal/statestore/backfill_test.go
```diff
@@ -300,4 +309,130 @@ func TestDeleteBackfill(t *testing.T) {
 	require.Error(t, err)
 	require.Equal(t, codes.Unavailable.String(), status.Convert(err).Code().String())
 	require.Contains(t, status.Convert(err).Message(), "DeleteBackfill, id: 12345, failed to connect to redis:")
+
+	// test that Backfill also deleted from last acknowledged sorted set
+	_, err = redis.Int64(conn.Do("ZSCORE", backfillLastAckTime, bfID))
```
akremsa, line 314: Shouldn't it be moved to a happy scenario condition?

internal/statestore/backfill_test.go
```diff
@@ -300,4 +309,130 @@ func TestDeleteBackfill(t *testing.T) {
...
+
+	// GetExpiredBackfillIDs should return only expired BF
+	service := New(cfg)
+	require.NotNil(t, service)
+	defer service.Close()
+	ctx := utilTesting.NewContext(t)
+	bfIDs, err := service.GetExpiredBackfillIDs(ctx)
+	require.NoError(t, err)
+	require.Len(t, bfIDs, 1)
+	require.Equal(t, expID, bfIDs[0])
```
akremsa, line 437: good!

internal/statestore/backfill.go
```diff
@@ -134,3 +139,68 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+	expiredBackfillIds, err := redis.Strings(redisConn.Do("ZRANGEBYSCORE", backfillLastAckTime, startTimeInt, endTimeInt))
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error getting expired backfills %v", err)
+	}
+
+	return expiredBackfillIds, nil
+}
+
+// deleteExpiredBackfillID - delete expired BackfillID from a sorted set
+func (rb *redisBackend) deleteExpiredBackfillID(ctx context.Context, backfillID string) error {
```
akremsa, line 190: you can pass redis connection here from a caller method. what do you think?

internal/statestore/backfill_test.go
aLekSer: Created a new one

internal/statestore/backfill.go
```diff
@@ -134,3 +139,62 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+
+// deleteExpiredBackfillID - delete expired BackfillID from a sorted set
+func (rb *redisBackend) deleteExpiredBackfillID(conn redis.Conn, backfillID string) error {
+	cmds := make([]interface{}, 0, 2)
+	cmds = append(cmds, backfillLastAckTime, backfillID)
+
+	_, err := conn.Do("ZREM", cmds...)
+	if err != nil {
+		return status.Errorf(codes.Internal, "failed to delete expired backfill ID %v",
+			errors.Wrap(err, "failed to delete expired backfill ID from Sorted Set"))
```
akremsa, line 197: You use the same two error messages here, please clean it up.

internal/statestore/backfill_test.go
```diff
@@ -275,20 +286,18 @@ func TestDeleteBackfill(t *testing.T) {
...
-				require.Equal(t, tc.expectedCode.String(), status.Convert(errActual).Code().String())
-				require.Contains(t, status.Convert(errActual).Message(), tc.expectedMessage)
+			if tc.backfillID != "" {
+				_, errGetTicket := service.GetTicket(ctx, tc.backfillID)
+				require.Error(t, errGetTicket)
+				require.Equal(t, codes.NotFound.String(), status.Convert(errGetTicket).Code().String())
 			}
+
+			// test that Backfill also deleted from last acknowledged sorted set
+			_, err = redis.Int64(conn.Do("ZSCORE", backfillLastAckTime, tc.backfillID))
```
akremsa, line 299: there is no point to perform this check if you are testing the case with an empty backfill id. I think it's better to put it under _if tc.backfillID != ""_ statement.

akremsa (approved):

Laremere:
> @Laremere Do you mean that we should not perform rebase on a new master also? I am used to do it on Agones.

We squash merge all PRs anyways, so rebasing isn't necessary.

internal/statestore/backfill_test.go
```diff
@@ -275,19 +286,23 @@ func TestDeleteBackfill(t *testing.T) {
...
-				}
-			} else {
-				require.Error(t, errActual)
-				require.Equal(t, tc.expectedCode.String(), status.Convert(errActual).Code().String())
-				require.Contains(t, status.Convert(errActual).Message(), tc.expectedMessage)
+			if tc.backfillID != "" {
+				// test that Backfill last acknowledged is in a sorted set
+				ts, err := redis.Int64(conn.Do("ZSCORE", backfillLastAckTime, tc.backfillID))
+				require.NoError(t, err)
+				require.True(t, ts > 0, "timestamp is valid")
```
akremsa, line 293: timestamp is NOT valid

internal/statestore/backfill.go
```diff
@@ -140,6 +142,65 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+
+// GetExpiredBackfillIDs gets all backfill IDs which are expired
+func (rb *redisBackend) GetExpiredBackfillIDs(ctx context.Context) ([]string, error) {
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Unavailable, "GetExpiredBackfillIDs, failed to connect to redis: %v", err)
+	}
+	defer handleConnectionClose(&redisConn)
+
+	// Use a fraction 80% of pendingRelease Tickets TTL
```
aLekSer, line 176: Will update this in this or the next PR

internal/statestore/backfill.go
```diff
@@ -140,6 +142,65 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+	if err != nil {
+		return status.Errorf(codes.Unavailable, "AcknowledgeBackfill, id: %s, failed to connect to redis: %v", id, err)
+	}
+	defer handleConnectionClose(&redisConn)
+
+	currentTime := time.Now().UnixNano()
+	cmds := make([]interface{}, 0)
+	cmds = append(cmds, backfillLastAckTime, currentTime, id)
+
+	_, err = redisConn.Do("ZADD", cmds...)
```
Laremere, line 158: There's no need for the oddness with cmds, just do:
`_, err = redisConn.Do("ZADD", backfillLastAckTime, currentTime, id)`

internal/statestore/backfill.go
```diff
@@ -28,7 +29,8 @@ import (
 )
 
 const (
-	allBackfills = "allBackfills"
+	backfillLastAckTime = "backfill_last_ack_time"
```
Laremere, line 32: This also needs to be set during backfill creation, so that a server which never starts up can be cleaned up.

internal/statestore/backfill.go
```diff
@@ -140,6 +142,65 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+
+	return expiredBackfillIds, nil
+}
+
+// deleteExpiredBackfillID deletes expired BackfillID from a sorted set
+func (rb *redisBackend) deleteExpiredBackfillID(conn redis.Conn, backfillID string) error {
+	cmds := make([]interface{}, 0, 2)
+	cmds = append(cmds, backfillLastAckTime, backfillID)
+
+	_, err := conn.Do("ZREM", cmds...)
```
Laremere, line 196: Here too.

internal/statestore/backfill_test.go
```diff
@@ -300,6 +315,147 @@ func TestDeleteBackfill(t *testing.T) {
...
+	require.NoError(t, err)
+
+	bfIDs, err := service.GetExpiredBackfillIDs(ctx)
+	require.Len(t, bfIDs, 0)
+	require.NoError(t, err)
+	err = service.AcknowledgeBackfill(ctx, bf1)
+	require.NoError(t, err)
+	err = service.AcknowledgeBackfill(ctx, bf2)
+	require.NoError(t, err)
+	// Sleep until the pending release expired and verify we still have all the tickets
```
Laremere, line 353: After acknowledging, but before sleeping, verify the expired backfills.

Laremere (changes requested):

internal/statestore/backfill.go
aLekSer: Yes, that's true, made a TODO for this I wanted to make this as part of separate PR.
But I think can be added here also.

aLekSer: Added additional call acknowledgeBackfill to StateStore CreateBackfill.

internal/statestore/backfill.go
```diff
@@ -59,6 +61,8 @@ func (rb *redisBackend) CreateBackfill(ctx context.Context, backfill *pb.Backfil
 	if res.(int64) == 0 {
 		return status.Errorf(codes.AlreadyExists, "backfill already exists, id: %s", backfill.GetId())
 	}
+
+	acknowledgeBackfill(redisConn, backfill.GetId())
```
akremsa, line 65: you are ignoring error here

aLekSer: yes, thanks for noticing, will update

akremsa: TestCreateBackfill should be updated according to this new behavior.

internal/statestore/backfill_test.go
```diff
@@ -300,6 +315,165 @@ func TestDeleteBackfill(t *testing.T) {
...
+	cfg, closer := createRedis(t, false, "")
+	defer closer()
+
+	startTime := time.Now()
+	service := New(cfg)
+	require.NotNil(t, service)
+	defer service.Close()
+	ctx := utilTesting.NewContext(t)
+	bf1 := "mockBackfillID"
+	err := service.CreateBackfill(ctx, &pb.Backfill{
```
akremsa, line 405: here you need to create backfill directly in redis to test AcknowledgeBackfill function separately, because CreateBackfill uses AcknowledgeBackfill under the hood.

akremsa: Not sure if backfill should be created here at all, it is not used anyway.
AcknowledgeBackfill accepts an id as a string.

aLekSer:
@Laremere all comments got resolved.

Laremere (approved):

PR #1290, opened 2020-11-20
Backfill indexing

hsorellana:
This PR adds `IndexBackfill`, `DeindexBackfill` and `GetIndexedBackfills` funcs to the statestore, funcs used in `QueryService` but, for code sanity, it was decided to put them in a separate PR

This is related to #1280 which is also related to #1240


internal/statestore/backfill.go
```diff
@@ -134,3 +143,58 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+
+// IndexBackfill adds the backfill to the index.
+func (rb *redisBackend) IndexBackfill(ctx context.Context, backfill *pb.Backfill) error {
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		return status.Errorf(codes.Unavailable, "IndexBackfill, id: %s, failed to connect to redis: %v", backfill.GetId(), err)
+	}
+	defer handleConnectionClose(&redisConn)
+
+	err = redisConn.Send("SADD", allBackfills, backfill.Id)
```
aLekSer, line 155: Here we should store version of the Backfill.
It could be hashmap `hset` with backfillId as key and version as value.

internal/statestore/backfill.go
```diff
@@ -134,3 +144,64 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+		return nil, status.Errorf(codes.Unavailable, "GetIndexedBackfills, failed to connect to redis: %v", err)
+	}
+	defer handleConnectionClose(&redisConn)
+
+	bfIndex, err := redis.StringMap(redisConn.Do("HGETALL", allBackfills))
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error getting all indexed backfill ids %v", err)
+	}
+
+	r := make(map[string]int)
```
akremsa, line 196: r := make(map[string]int, len(bfIndex))?

In the happy case, there won't be any allocations in a further for loop.

internal/statestore/backfill.go
```diff
@@ -134,3 +144,64 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+
+// GetIndexedBackfills returns the ids of all backfills currently indexed.
+func (rb *redisBackend) GetIndexedBackfills(ctx context.Context) (map[string]int, error) {
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Unavailable, "GetIndexedBackfills, failed to connect to redis: %v", err)
+	}
+	defer handleConnectionClose(&redisConn)
+
+	bfIndex, err := redis.StringMap(redisConn.Do("HGETALL", allBackfills))
```
akremsa, line 191: What if bfIndex is empty?
I think you need to return an error with a proper description in this case. 
Something like "no indexed backfills found".

aLekSer: This is not actually an error, at the start up time there would be 0 backfills, or after all backfills would be deleted. So I think error should be `nil` because nothing bad have happened. We can add a warning log I think at a later PR.

internal/statestore/backfill.go
hsorellana: nice catch!

internal/statestore/backfill.go
```diff
@@ -26,7 +27,13 @@ import (
 	"open-match.dev/open-match/pkg/pb"
 )
 
-// CreateBackfill creates a new Backfill in the state storage if one doesn't exist. The xids algorithm used to create the ids ensures that they are unique with no system wide synchronization. Calling clients are forbidden from choosing an id during create. So no conflicts will occur.
+const (
+	allBackfills = "allBackfills"
+)
+
+// CreateBackfill creates a new Backfill in the state storage if one doesn't exist.
```
Laremere, line 34: What's with all the comment formatting changes?  They aren't relevant to this PR...

internal/statestore/backfill.go
```diff
@@ -134,3 +144,64 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+	bfIndex, err := redis.StringMap(redisConn.Do("HGETALL", allBackfills))
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error getting all indexed backfill ids %v", err)
+	}
+
+	r := make(map[string]int, len(bfIndex))
+	for bfID, bfGeneration := range bfIndex {
+		gen, err := strconv.Atoi(bfGeneration)
+		if err != nil {
+			// TODO: handle err when generation cannot be parsed
```
Laremere, line 200: Log with error?  I would also be good with just returning an error.

Logging is less disastrous if this happens in a production scenario,  but returning an error is more likely to cause of problem to be caught before things get to production.  Either way, it's a "this should never happen" type thing.

internal/statestore/backfill.go
```diff
@@ -134,3 +144,64 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+
+// IndexBackfill adds the backfill to the index.
+func (rb *redisBackend) IndexBackfill(ctx context.Context, backfill *pb.Backfill) error {
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
+		return status.Errorf(codes.Unavailable, "IndexBackfill, id: %s, failed to connect to redis: %v", backfill.GetId(), err)
+	}
+	defer handleConnectionClose(&redisConn)
+
+	// TODO: review usage of version or stick with Generation
```
Laremere, line 156: Here (and other todo) can you add a task on your backlog for this?  I think generation is going to work here, so I'm worried nobody is going to look at this code and the todo will stick around...

Laremere: Otherwise, create an issue on github, saying these todos need to be cleaned up, and reference the issue in the todo, eg

// TODO(#1234): clean the gutters

internal/statestore/backfill.go
hsorellana: Those are just some minor upgrades i found when reading the code that could be done without affecting the context of the PR

internal/statestore/backfill.go
hsorellana: Yes, absolutely! I think there's a conversation needed to happen so about the Version/Generation issue, and as soon that clears, i'll delete the TODOs i've put

internal/statestore/backfill.go
hsorellana: The doubt that i had here is that technically `strconv.Atoi` will never return in this context, because it is converting the Generation field which, when indexing, it is always the same data type. (this happens [here](https://github.com/googleforgames/open-match/blob/649fe89908361a6a8baa5e4ebd0dfe1f237daa77/internal/statestore/backfill.go#L157)). 
But dismissing the error seems a little too much code responsability.

internal/statestore/backfill.go
Laremere: Please avoid unrelated style cleanups in PRs.  Standalone changes to do cleanup is fine, as long as there's no fighting about style (which is just a waste of everyone's time.)

internal/statestore/backfill.go
Laremere: Right, so returning an error for a really "this should never happen" is a good way to say loudly that something that shouldn't happen is happening.  Just do that?

Either  way, I don't see why this has to wait under a TODO in this PR, instead of making a choice either way.

hsorellana: Ok, i see your point. And this TODO its just a remainder of asking about it.

internal/statestore/backfill.go
hsorellana: there's a new test case with empty backfill index, and it doesn't raise an error, just returns an empty map

akremsa: ok, thanks for the reply

internal/statestore/instrumented.go
```diff
@@ -141,3 +141,24 @@ func (is *instrumentedService) NewMutex(key string) RedisLocker {
 	defer span.End()
 	return is.s.NewMutex(key)
 }
+
+// IndexBackfill adds the backfill to the index.
+func (is *instrumentedService) IndexBackfill(ctx context.Context, backfill *pb.Backfill) error {
+	_, span := trace.StartSpan(context.Background(), "statestore/instrumented.IndexBackfill")
```
yeukovichd, line 147: why is context.Background is used instead of ctx?

hsorellana: probably slipped in conflict resolution. Will fix it

akremsa (approved):

internal/statestore/backfill_test.go
```diff
@@ -301,3 +301,122 @@ func TestDeleteBackfill(t *testing.T) {
...
+
+	ctx := utilTesting.NewContext(t)
+
+	verifyBackfills := func(service Service, backfills []*pb.Backfill) {
+		ids, err := service.GetIndexedBackfills(ctx)
+		require.Nil(t, err)
+		require.Equal(t, len(backfills), len(ids))
+
+		for _, tt := range backfills {
+			_, ok := ids[tt.GetId()]
```
aLekSer, line 390: Can you please add one more test case when you Index a new Backfill with Generation > 1 and get an output
```suggestion
			gen, ok := ids[tt.GetId()]
			require.Equal(t, tt.Generation, gen)
```

internal/statestore/backfill_test.go
```diff
@@ -301,3 +301,122 @@ func TestDeleteBackfill(t *testing.T) {
...
+	defer service.Close()
+
+	ctx := utilTesting.NewContext(t)
+
+	verifyBackfills := func(service Service, backfills []*pb.Backfill) {
+		ids, err := service.GetIndexedBackfills(ctx)
+		require.Nil(t, err)
+		require.Equal(t, len(backfills), len(ids))
+
+		for _, tt := range backfills {
```
aLekSer, line 389:
```suggestion
		for _, bf := range backfills {
```

internal/statestore/backfill_test.go
```diff
@@ -301,3 +301,122 @@ func TestDeleteBackfill(t *testing.T) {
...
+	verifyBackfills(service, []*pb.Backfill{})
+
+	// two indexed backfills exists
+	backfills := generateBackfills(ctx, t, service, 2)
+	verifyBackfills(service, backfills)
+
+	// deindex one backfill, one backfill exist
+	err := service.DeindexBackfill(ctx, backfills[0].Id)
+	require.Nil(t, err)
+	verifyBackfills(service, backfills[1:2])
```
aLekSer, line 405:
```
newGen : = 1000
backfills[1].Generation = newGen
		bf, service.UpdateBackfill(ctx, backfills[1], []string{}))
		require.NoError(t, service.IndexBackfill(ctx, backfills[1]))
```

aLekSer (changes requested):
Thanks for this PR, overall is great. But I think `GetIndexedBackfills()` result should be more properly unit tested.

internal/statestore/backfill_test.go
hsorellana: I don't see exactly what you want tested here. If you are concerned about `HSET` the documentation says if the key already exists, it is overwritten with the new value. And the scenario you are proposing i have it tested in the backfill caching update tests

Redis documentation for  [`HSET`](https://redis.io/commands/hset) says:
> If key does not exist, a new key holding a hash is created. If field already exists in the hash, it is overwritten.

internal/statestore/backfill_test.go
hsorellana: Applied changes

internal/statestore/backfill_test.go
aLekSer: Yes, just wanted to test update the of the index, just to check it is updated

internal/statestore/backfill.go
aLekSer: #1294 Was created

aLekSer (approved):
Thanks for updating UT, also great to see how this PR would be used in E2E test in `QueryBackfills` PR.

internal/statestore/backfill.go
Laremere: Remove the todo, and this can be merged.

hsorellana: Go for merging @Laremere 👍

Laremere (approved):

PR #1292, opened 2020-11-25
Frontend: UpdateBackfill and DeleteBackfill handlers

aLekSer:

UD methods for Backfill structure.

Work on #1240

This PR contains another PR with CRud part of Frontend: #1279 

So review please only changes from this commit and later commits:
https://github.com/googleforgames/open-match/pull/1292/commits/9f970a7a8b85ebe2018743c4344a671614c7c810


internal/app/frontend/frontend_service.go
```diff
@@ -187,22 +330,9 @@ func (s *frontendService) AcknowledgeBackfill(ctx context.Context, req *pb.Ackno
...
-func (s *frontendService) CreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest) (*pb.Backfill, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
-}
-
-// DeleteBackfill deletes a Backfill by its ID.
-func (s *frontendService) DeleteBackfill(ctx context.Context, req *pb.DeleteBackfillRequest) (*empty.Empty, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
-}
-
+// DeleteBackfill deletes a Back
```
hsorellana, line 333: this comment is unrelated here

internal/app/frontend/frontend_service.go
```diff
@@ -94,6 +94,149 @@ func doCreateTicket(ctx context.Context, req *pb.CreateTicketRequest, store stat
...
+//   - If SearchFields exist in a Backfill, CreateBackfill will also index these fields such that one can query the ticket with query.QueryBackfills function.
+func (s *frontendService) CreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest) (*pb.Backfill, error) {
+	// Perform input validation.
+	if req == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "request is nil")
+	}
+	if req.Backfill == nil {
+		return nil, status.Errorf(codes.InvalidArgument, ".backfill is required")
+	}
+	if req.Backfill.CreateTime != nil {
```
yeukovichd, line 109: Why is this check required? If we look at doCreateBackfill it always assigns new value to CreateTime.

internal/app/frontend/frontend_service.go
```diff
@@ -94,6 +94,149 @@ func doCreateTicket(ctx context.Context, req *pb.CreateTicketRequest, store stat
...
+
+	err := m.Lock(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if _, err = m.Unlock(ctx); err != nil {
+
+			logger.WithFields(logrus.Fields{
+				"error": err.Error(),
```
yeukovichd, line 165: Maybe it is better to use WithError and specify fields such as backfill_id?

internal/app/frontend/frontend_service.go
aLekSer: This is a leftover of CreateTicket and is part of #1279 This PR would be rebased on top of master after #1279 would be merged.

aLekSer:
Unrelated to the change CI failure:
```

time="2020-11-26T00:14:25Z" level=error msg="context canceled" app=openmatch component=api.test
--- FAIL: TestMatchFunctionMatchCollision (0.43s)
    fetch_matches_test.go:158: 
        	Error Trace:	fetch_matches_test.go:158
        	Error:      	Should be true
        	Test:       	TestMatchFunctionMatchCollision
        	Messages:   	381.29587ms
```

internal/app/frontend/frontend_service.go
```diff
@@ -137,6 +137,106 @@ func doCreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest, store
...
+		if _, err = m.Unlock(ctx); err != nil {
+			logger.WithError(err).Error("error on mutex unlock")
+		}
+	}()
+	bfStored, associatedTickets, err := s.store.GetBackfill(ctx, bfID)
+	if err != nil {
+		return nil, err
+	}
+
+	err = s.store.DeleteTicketsFromPendingRelease(ctx, associatedTickets)
```
Laremere, line 174: This must be done after the ticket is updated:
If the server performs this action, then dies, the tickets will think they can return to the active pool and find other matches, but the backfill will still think it can put tickets in this backfill, potentially assigning them twice.

The other way around, the backfill will release it's claim in the tickets, and if this server dies, then the tickets will just time out and then re-enter the pool.

internal/app/frontend/frontend_service.go
```diff
@@ -137,6 +137,106 @@ func doCreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest, store
...
+// UpdateBackfill updates a Backfill object, if present.
+func (s *frontendService) UpdateBackfill(ctx context.Context, req *pb.UpdateBackfillRequest) (*pb.Backfill, error) {
+	if req == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "request is nil")
+	}
+	if req.Backfill == nil {
+		return nil, status.Errorf(codes.InvalidArgument, ".backfill is required")
+	}
+
+	backfill, ok := proto.Clone(req.Backfill).(*pb.Backfill)
```
Laremere, line 149: The update here should be done defensively, currently you perform the get, but use the backfill passed in as the source of truth.  Instead start with the backfill from the get, and then update each field which should be updated.  Eg, create time would be overwritten if the wrong value were sent here.

internal/testing/e2e/backfill_test.go
```diff
@@ -65,4 +65,37 @@ func TestCreateGetBackfill(t *testing.T) {
...
+		StringArgs: map[string]string{
+			"search": "me",
+		},
+	},
+	}
+	createdBf, err := om.Frontend().CreateBackfill(ctx, &pb.CreateBackfillRequest{Backfill: bf})
+	require.Nil(t, err)
+
+	createdBf.SearchFields.StringArgs["key"] = "val"
+	updatedBf, err := om.Frontend().UpdateBackfill(ctx, &pb.UpdateBackfillRequest{Backfill: createdBf})
```
Laremere, line 86: To better test this method, send an entirely new Backfill protobuf object, with the new value but setting the ID to the one from createdBf.

As it is, this test incorrectly thinks it verifies CreateTime can't be updated.

Laremere: Also it would be a better way to show that generation doesn't need to be right.

internal/testing/e2e/backfill_test.go
```diff
@@ -65,4 +65,37 @@ func TestCreateGetBackfill(t *testing.T) {
...
+	om := newOM(t)
+	ctx := context.Background()
+
+	bf := &pb.Backfill{SearchFields: &pb.SearchFields{
+		StringArgs: map[string]string{
+			"search": "me",
+		},
+	},
+	}
+	createdBf, err := om.Frontend().CreateBackfill(ctx, &pb.CreateBackfillRequest{Backfill: bf})
```
Laremere, line 82: Test generation values?

Laremere (changes requested):

internal/app/frontend/frontend_service.go
aLekSer: Updated the code with:

```
	// Update generation here, because Frontend is used by GameServer only
	bfStored.Generation++
	bfStored.SearchFields = backfill.SearchFields
	bfStored.Extensions = backfill.Extensions
```

aLekSer:
All comments are resolved.

internal/app/frontend/frontend_service.go
```diff
@@ -122,6 +122,8 @@ func doCreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest, store
 
 	backfill.Id = xid.New().String()
 	backfill.CreateTime = ptypes.TimestampNow()
+	// Set Generation to 1
```
Laremere, line 125: Why this comment?  It's entirely redundant with the line below?

Laremere (approved):
Extra comment should be removed, but a few things are depending on this so I'll merge it if the tests pass.

internal/app/frontend/frontend_service.go
aLekSer: That's true updated

Laremere:
Tests are still failing....

PR #1293, opened 2020-11-25
Frontend acknowledge backfill

aLekSer:
Add AcknowledgeBackfill() request handler.

Create AssignmentGroup for all associated tickets and use Redis stateStore, not backend to assign all tickets.

Work on #1240 

Contains part of #1279 as well as #1288 , so check only latest commit when viewing files changes. Namely:
https://github.com/googleforgames/open-match/pull/1293/commits/9b5cfde3f3a6cbb18c2b89f1b86ce46b8ce13fa5
AcknowledgeBackfill do not update Generation, unlike other UpdateBackfill methods (from GS and Frontend).


internal/app/frontend/frontend_service.go
```diff
@@ -181,15 +181,60 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
...
+	}
 
-// CreateBackfill creates a new Backfill object.
-func (s *frontendService) CreateBackfill(ctx context.Context, req *pb.CreateBackfillRequest) (*pb.Backfill, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+	bf, associatedTickets, err := s.store.GetBackfill(ctx, req.GetBackfillId())
+	if err != nil {
+		return nil, err
+	}
+	ag := pb.AssignmentGroup{TicketIds: associatedTickets, Assignment: req.GetAssignment()}
```
hsorellana, line 231: I think this should be set inline, when setting assignments, or at least inside the len if block. But it has no use at the main level of execution

internal/app/frontend/frontend_service.go
```diff
@@ -181,15 +181,60 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
...
+	if len(associatedTickets) != 0 {
+		_, _, err = s.store.UpdateAssignments(ctx, &pb.AssignTicketsRequest{
+			Assignments: []*pb.AssignmentGroup{&ag},
+		})
```
hsorellana, line 235: Why not return the err here, if err != nil??

aLekSer:
CreateBackfill() should also query AcknowledgeBackfill in Redis, just to set initial initial timestamp.
Frontend AcknowledgeBackfill request, should first check if Backfill exists in Redis, otherwise AcknowledgeBackfill can create a new record in sorted set for non-existent Backfill.

internal/testing/e2e/backfill_test.go
```diff
@@ -321,14 +302,19 @@ func TestBackfillGenerationMismatch(t *testing.T) {
...
 	require.Equal(t, io.EOF, err)
-}
 
-func mustAny(m proto.Message) *any.Any {
-	result, err := ptypes.MarshalAny(m)
-	if err != nil {
-		panic(err)
-	}
-	return result
+	actual, err := om.Frontend().GetBackfill(ctx, &pb.GetBackfillRequest{BackfillId: bfID})
```
akremsa, line 312: Why do you test GetBackfill here?
The name of the function is createMatchWithBackfill and you are passing backfill as a parameter.
I think if you want to test GetBackfill you should do it out of this method in order to not overload one with a lot of code. createMatchWithBackfill does more than a simple create match step here.

aLekSer: This is a refactor of @yeukovichd test. Now it is used in 3 place, so that we can associate ticket with backfill using a match proposal.

aLekSer: One more additional check is not bad. So I will not remove this line.

internal/app/frontend/frontend_service.go
```diff
@@ -354,7 +354,26 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
 // AcknowledgeBackfill is used to notify OpenMatch about GameServer connection info.
 // This triggers an assignment process.
 func (s *frontendService) AcknowledgeBackfill(ctx context.Context, req *pb.AcknowledgeBackfillRequest) (*pb.Backfill, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+	if req.GetBackfillId() == "" {
+		return nil, status.Errorf(codes.InvalidArgument, ".BackfillId is required")
+	}
+	if req.GetAssignment().Connection == "" {
```
akremsa, line 360: If Assignment is nil you will fail here with `invalid memory address or nil pointer dereference` error

internal/testing/e2e/backfill_test.go
```diff
@@ -132,20 +131,39 @@ func TestBackfillFrontendLifecycle(t *testing.T) {
...
-	require.Nil(t, err)
+	},
+	}
+	createdBf, err := om.Frontend().CreateBackfill(ctx, &pb.CreateBackfillRequest{Backfill: bf})
+	require.NoError(t, err)
+	require.Equal(t, int64(1), createdBf.Generation)
+
+	ticketID := createMatchWithBackfill(ctx, om, createdBf, t)
+	conn := "127.0.0.1:4242"
+	getBF, err := om.Frontend().AcknowledgeBackfill(ctx, &pb.AcknowledgeBackfillRequest{BackfillId: createdBf.Id, Assignment: &pb.Assignment{Connection: conn, Extensions: map[string]*any.Any{
```
akremsa, line 150: tests with parameters validation are missing. seems you are testing only a happy path.

akremsa (changes requested):

internal/app/frontend/frontend_service.go
```diff
@@ -354,7 +354,26 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
...
+	}
+	if len(associatedTickets) != 0 {
+		_, _, err = s.store.UpdateAssignments(ctx, &pb.AssignTicketsRequest{
+			Assignments: []*pb.AssignmentGroup{&pb.AssignmentGroup{TicketIds: associatedTickets, Assignment: req.GetAssignment()}},
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
+	return bf, nil
```
akremsa, line 376: Seems AcknowledgeBackfill is missing?

internal/app/frontend/frontend_service.go
```diff
@@ -354,7 +354,32 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
...
+	if req.GetBackfillId() == "" {
+		return nil, status.Errorf(codes.InvalidArgument, ".BackfillId is required")
+	}
+	if req.GetAssignment().Connection == "" {
+		return nil, status.Errorf(codes.InvalidArgument, ".Assignment.Connection should be set")
+	}
+
+	bf, associatedTickets, err := s.store.GetBackfill(ctx, req.GetBackfillId())
+	if err != nil {
+		return nil, err
```
hsorellana, line 366: Shouldn't these errors be returned with status.Error??

akremsa: GetBackfill already returns exact status codes

internal/testing/e2e/backfill_test.go
```diff
@@ -132,20 +131,39 @@ func TestBackfillFrontendLifecycle(t *testing.T) {
 	require.Nil(t, get)
 }
 
-func TestProposedBackfillCreate(t *testing.T) {
-	ctx := context.Background()
+func TestAcknowledgeBackfill(t *testing.T) {
```
akremsa, line 134: Actually, this test does not cover AcknowledgeBackfill at all, otherwise, you would notice that you do not call the Acknowledge backfill method from a statestore. This test should be completely rewritten.

internal/testing/e2e/backfill_test.go
aLekSer: Well I think parameters validation can form a new test - `frontend_service_test.go`

internal/app/frontend/frontend_service.go
aLekSer: Added

internal/testing/e2e/backfill_test.go
aLekSer: This is E2E test, so it is hard to test this.
I know that there is an additional Redis state change - this would be covered in E2E test.
It covers one of main effects of acknowledgment - all tickets get assigned.

akremsa: Ok, I see why you can not cover AcknowledgeBackfill step in e2e, you still can do that in unit test to check the redis state. What do you think?

aLekSer:
@akremsa Addressed your comments. Tests which are flaky on CI are TestProposedBackfillCreate and TestProposedBackfillUpdate

internal/app/frontend/frontend_service_test.go
```diff
@@ -340,6 +340,88 @@ func TestDoWatchAssignments(t *testing.T) {
...
 
+// TestAcknowledgeBackfill - test input validation
+func TestAcknowledgeBackfillValidation(t *testing.T) {
+	cfg := viper.New()
+	tests := []struct {
+		description string
+		request     *pb.AcknowledgeBackfillRequest
+	}{
+		{
+			description: "check backfillId exists",
```
akremsa, line 351: Please, provide a more informative test case description.
Follow the pattern - "what you test, what you expect"
Example: "no backfillID, err expected"

internal/app/frontend/frontend_service_test.go
```diff
@@ -340,6 +340,88 @@ func TestDoWatchAssignments(t *testing.T) {
...
+	for _, test := range tests {
+		test := test
+		t.Run(test.description, func(t *testing.T) {
+			ctx := context.Background()
+
+			store, closer := statestoreTesting.NewStoreServiceForTesting(t, cfg)
+			defer closer()
+			fs := frontendService{cfg, store}
+			bf, err := fs.AcknowledgeBackfill(ctx, test.request)
+			require.Equal(t, codes.InvalidArgument.String(), status.Convert(err).Code().String())
```
akremsa, line 373: Check not only an error code but also an error description. This way you ensure that you receive exactly that error which you expect. Check tests in statestore for reference.

internal/app/frontend/frontend_service_test.go
```diff
@@ -340,6 +340,88 @@ func TestDoWatchAssignments(t *testing.T) {
...
+			require.Nil(t, bf)
+		})
+	}
+}
+
+// TestAcknowledgeBackfill - CreateBackfill,
```
akremsa, line 379: What is the point of this comment? It doesn't give any extra information.

aLekSer: I extended this one

internal/app/frontend/frontend_service_test.go
```diff
@@ -340,6 +340,88 @@ func TestDoWatchAssignments(t *testing.T) {
...
+	require.NoError(t, err)
+	fs := frontendService{cfg, store}
+	ids, err := store.GetExpiredBackfillIDs(ctx)
+	require.NoError(t, err)
+	require.Len(t, ids, 0)
+
+	time.Sleep(cfg.GetDuration("pendingReleaseTimeout"))
+	ids, err = store.GetExpiredBackfillIDs(ctx)
+	require.NoError(t, err)
+	require.Len(t, ids, 1)
```
akremsa, line 405: these 2 calls of GetExpiredBackfillIDs can be safely removed because here you test GetExpiredBackfillIDs actually and they are not related to TestAcknowledgeBackfill. In this test, we assume that they work as expected, so please don't waste test execution time using sleep on these steps. This functionality is properly covered in statestore tests.

internal/app/frontend/frontend_service_test.go
```diff
@@ -340,6 +340,88 @@ func TestDoWatchAssignments(t *testing.T) {
...
+			fs := frontendService{cfg, store}
+			bf, err := fs.AcknowledgeBackfill(ctx, test.request)
+			require.Equal(t, codes.InvalidArgument.String(), status.Convert(err).Code().String())
+			require.Nil(t, bf)
+		})
+	}
+}
+
+// TestAcknowledgeBackfill - CreateBackfill,
+func TestAcknowledgeBackfill(t *testing.T) {
```
akremsa, line 380: Please add a case where you pass a backfill id that doesn't exist and check a proper error.

internal/testing/e2e/backfill_test.go
aLekSer: Added UT.

internal/app/frontend/frontend_service_test.go
akremsa: ?

internal/app/frontend/frontend_service_test.go
```diff
@@ -340,6 +340,100 @@ func TestDoWatchAssignments(t *testing.T) {
...
+			request:         &pb.AcknowledgeBackfillRequest{BackfillId: "", Assignment: &pb.Assignment{Connection: "10.0.0.1"}},
+			expectedMessage: ".BackfillId is required",
+		},
+		{
+			description:     "no Assignment, error is expected",
+			request:         &pb.AcknowledgeBackfillRequest{BackfillId: "1234", Assignment: nil},
+			expectedMessage: ".Assignment is required",
+		},
+		{
+			description:     "no Assignment.Connection, error is expected Assignment",
```
akremsa, line 362: Assignment", - typo?

aLekSer: yes

internal/app/frontend/frontend_service_test.go
aLekSer: You can see in a latest commit

internal/testing/e2e/backfill_test.go
```diff
@@ -238,6 +206,26 @@ func TestProposedBackfillUpdate(t *testing.T) {
 			Score: 10,
 		}),
 	}
+	createMatchWithBackfill(ctx, om, b, t)
+	client, err := om.Query().QueryTickets(ctx, &pb.QueryTicketsRequest{Pool: &pb.Pool{
+		StringEqualsFilters: []*pb.StringEqualsFilter{{StringArg: "field", Value: "value"}},
+	}})
+
+	require.Nil(t, err)
```
hsorellana, line 214:
```suggestion
	require.NoError(t, err)

```

aLekSer: I was testing a task on VS Code. And this is a leftover. changed back already.

akremsa: good catch. 
@aLekSer please update it everywhere in all test files you are working with

akremsa (approved):

internal/app/frontend/frontend_service.go
```diff
@@ -354,7 +354,35 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
...
 // This triggers an assignment process.
 func (s *frontendService) AcknowledgeBackfill(ctx context.Context, req *pb.AcknowledgeBackfillRequest) (*pb.Backfill, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+	if req.GetBackfillId() == "" {
+		return nil, status.Errorf(codes.InvalidArgument, ".BackfillId is required")
+	}
+	if req.GetAssignment() == nil {
+		return nil, status.Errorf(codes.InvalidArgument, ".Assignment is required")
+	}
+	if req.GetAssignment().Connection == "" {
```
Laremere, line 363: Connection is not a required field on assignments.

internal/app/frontend/frontend_service.go
```diff
@@ -354,7 +354,35 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
...
+		return nil, status.Errorf(codes.InvalidArgument, ".Assignment.Connection should be set")
+	}
+
```
Laremere, line 366: Shouldn't this be locking the backfill before running?

internal/app/frontend/frontend_service.go
```diff
@@ -354,7 +354,35 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
...
+	if len(associatedTickets) != 0 {
+		_, _, err = s.store.UpdateAssignments(ctx, &pb.AssignTicketsRequest{
+			Assignments: []*pb.AssignmentGroup{&pb.AssignmentGroup{TicketIds: associatedTickets, Assignment: req.GetAssignment()}},
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
+	return bf, nil
 }
```
Laremere, line 386: After the tickets are assigned, they need to be removed from the associated ticket list?  Otherwise they'll be assigned every acknowledge call?

aLekSer:
All comments are resolved

thisisnotapril:
@HazWard @calebatwd would love to have y'all review

internal/app/frontend/frontend_service.go
aLekSer: Updated: now removing only assigned tickets from Associated tickets list.

internal/app/frontend/frontend_service.go
aLekSer: Added the lock

internal/app/frontend/frontend_service.go
aLekSer: And inside the mutex

davidedmonds (approved):
Looking good from over here - seems all the comments have been addressed as well.

andrewgrundy (approved):

internal/app/frontend/frontend_service.go
sawagh: Curious if the entire logic (Assign Tickets, Remove the assigned ones from backfill, update acknowledged state on backfill) should live in the store layer itself? Having a AcknowledgeBackfill here that does all this individually and then calls AcknowledgeBackfill on the store which just updates the state is confusing. Any reason to have this logic separated from store? Is the store.AcknowledgeBackfill called other places where the rest of pieces of acknowledging are not needed?

sawagh (changes requested):

internal/app/frontend/frontend_service.go
```diff
@@ -354,7 +354,62 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
...
 func (s *frontendService) AcknowledgeBackfill(ctx context.Context, req *pb.AcknowledgeBackfillRequest) (*pb.Backfill, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+	if req.GetBackfillId() == "" {
+		return nil, status.Errorf(codes.InvalidArgument, ".BackfillId is required")
+	}
+	if req.GetAssignment() == nil {
+		return nil, status.Errorf(codes.InvalidArgument, ".Assignment is required")
+	}
+
+	m := s.store.NewMutex(req.GetBackfillId())
```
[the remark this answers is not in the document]
aLekSer: Note that on line 406 we execute `UpdateBackfill` and Backfill internal protobuf format includes also ticketed associated with this Backfill. So this lock is necessary here.
https://github.com/googleforgames/open-match/pull/1293/files#diff-fbf6a8aad66158048250071262ad0c5fd8c08c9034840c5802784888408881f3R406

internal/app/frontend/frontend_service.go
```diff
@@ -354,7 +354,62 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
...
+
+	if len(associatedTickets) != 0 {
+		_, tickets, err := s.store.UpdateAssignments(ctx, &pb.AssignTicketsRequest{
+			Assignments: []*pb.AssignmentGroup{&pb.AssignmentGroup{TicketIds: associatedTickets, Assignment: req.GetAssignment()}},
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		// all assigned tickets should be removed from the associatedTickets
```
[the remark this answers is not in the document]
aLekSer: Note that `UpdateAssignment` can return not the full list and some tickets are left unassigned.

internal/app/frontend/frontend_service.go
aLekSer: Actually I have chosen probably not the best name for statestore function `AcknowledgeBackfill`. I will rename statestore function to `UpdateAcknowledgmentTimetstamp`.

aLekSer:
Actually I have chosen probably not the best name for statestore function AcknowledgeBackfill. I will rename statestore function to UpdateAcknowledgmentTimetstamp.

internal/statestore/backfill_test.go
```diff
@@ -411,7 +411,7 @@ func TestAcknowledgeBackfillLifecycle(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func TestAcknowledgeBackfill(t *testing.T) {
+func TestUpdateAcknowledgmentTimestamptTimestamp(t *testing.T) {
```
yeukovichd, line 414: Double timestamp?

internal/statestore/backfill.go
```diff
@@ -200,18 +200,18 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
-		return status.Errorf(codes.Unavailable, "AcknowledgeBackfill, id: %s, failed to connect to redis: %v", id, err)
+		return status.Errorf(codes.Unavailable, "UpdateAcknowledgmentTimestamp, id: %s, failed to connect to redis: %v", id, err)
 	}
 	defer handleConnectionClose(&redisConn)
-	return acknowledgeBackfill(redisConn, id)
+	return updateAcknowledgmentTimestamp(redisConn, id)
 }
 
-func acknowledgeBackfill(conn redis.Conn, backfillID string) error {
+func updateAcknowledgmentTimestamp(conn redis.Conn, backfillID string) error {
```
yeukovichd, line 214: Maybe it is better to use prefix 'do' like in other place?

aLekSer: Thanks for the hint will add `do`

internal/statestore/backfill_test.go
aLekSer: Sorry resolved

aLekSer:
Simplified the `e2e/backfill_test.go` review process, refactoring of the same `createMatchWithBackfill` I moved out of this PR.

aLekSer:
@SaWagh I have applied all your comments. PTAL

internal/app/frontend/frontend_service.go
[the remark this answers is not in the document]
sawagh: Looking at the code in UpdateAssignment, any ticket left unassigned is because it was not found. (See: AssignmentFailure_TICKET_NOT_FOUND). So currently, any ticket that is returned as not found from UpdateAssignment is still left on the Backfill? We should atleast log the tickets that we leave unassigned here as UpdateAssignment returns a list of errors along with a list of assigned tickets (currently we ignore the errors list)

internal/app/frontend/frontend_service.go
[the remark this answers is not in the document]
sawagh: I do agree that the lock is necessary here  prevent the backfill from being modified from concurrent calls. I am not sure if it is sufficient. Can you explain how it protects the UpdateAssignments call from modifying the tickets while they are being updated from the AcknowledgeBackfill call?

internal/app/frontend/frontend_service.go
sawagh: Plan to do this in this PR or a followup PR?

aLekSer: I have already applied this change to StateStore.

internal/app/frontend/frontend_service.go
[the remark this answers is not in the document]
aLekSer: I have added logging of the errors on assignTickets request.

internal/app/frontend/frontend_service.go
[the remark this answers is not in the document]
aLekSer: Tickets are immutable, so we don't need to keep track of their state. They can only be assigned, but searchFields could not be updated, this is the main difference with Backfill which has generation (version) number.

aLekSer:
@sawagh I have added error logging with Test case containing one of two tickets got deleted before the AcknowledgeBackfill call.

internal/app/frontend/frontend_service.go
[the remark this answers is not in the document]
sawagh: I did not mention updating the searchFields at all. 

TicketA is a part of BackfillA. You receive a call to AcknowledgeBackfill for BackfillA - and are updating the assignment on the TicketA by making UpdateAssignments call - Line 389.

Concurrently, the Backend receives AssignTickets method setting Assignment for TicketA to a different assignment. 

Is there anything that prevents the AssignTickets method from calling  UpdateAssignments call on TicketA concurrently with AcknowledgeBackfill calliong UpdateAssignments for TicketA?

sawagh (approved):
If we dont resolve the concurrency comment, I am ok checking this in and tracking that in a separate issue.

internal/app/frontend/frontend_service.go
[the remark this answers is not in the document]
aLekSer: Actually this is not possible to have one ticket being associated to two backfills or used in a normal match. Evaluator would not make this happen. Only one match would be selected with a ticketA and later on it would be moved into `pendingRelease` tickets not visible by `QueryTickets` later on. We are working on scale tests it will let us know if concurrent issues like this would arise.

PR #1297, opened 2020-11-30
Clean up expired backfills

akremsa:
Clean up expired backfills. This functionality should be used by synchronizer.


aLekSer:
@googlebot I consent.

internal/statestore/backfill_test.go
```diff
@@ -117,7 +120,7 @@ func TestUpdateBackfill(t *testing.T) {
 			expectedMessage: "",
 		},
 		{
-			description:     "create existing backfill, no err expected",
+			description:     "create existing backfill, no err expecteds",
```
hsorellana, line 123: This is a typo, should be undo to what it was

akremsa: thanks a lot

akremsa:
@Laremere, please review

internal/statestore/backfill_test.go
```diff
@@ -158,7 +158,7 @@ func TestUpdateBackfill(t *testing.T) {
 				require.NoError(t, err)
 			} else {
 				require.Error(t, err)
-				require.Equal(t, tc.expectedCode.String(), status.Convert(err).Code().String())
+				require.Equal(t, tc.expectedCode.String(), status.Convert(err).Code().String(), "QQQ")
```
hsorellana, line 161: Debugging??

akremsa: Seems yes, thanks, Hugo. 👍

internal/testing/e2e/backfill_test.go
```diff
@@ -325,6 +325,64 @@ func TestBackfillGenerationMismatch(t *testing.T) {
 	require.Equal(t, io.EOF, err)
 }
 
+func TestCleanUpBackfills(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	om := newOM(t)
+
+	t1, err := om.Frontend().CreateTicket(ctx, &pb.CreateTicketRequest{Ticket: &pb.Ticket{}})
+	require.Nil(t, err)
```
hsorellana, line 334: probably use require.NoError??

akremsa: agree, changing it right now.

internal/statestore/backfill.go
```diff
@@ -22,12 +22,20 @@ import (
...
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"open-match.dev/open-match/internal/ipb"
 	"open-match.dev/open-match/pkg/pb"
 )
 
+var (
+	logger = logrus.WithFields(logrus.Fields{
+		"app":       "openmatch",
+		"component": "statestore",
```
hsorellana, line 35: what about using statestore.Redis?? so logs declares which implementation is using

internal/statestore/backfill.go
```diff
@@ -143,6 +151,48 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+	if err != nil {
+		return err
+	}
+
+	for _, id := range expiredBfIDs {
+		_, tickets, err := rb.GetBackfill(ctx, id)
+		if err != nil {
+			logger.WithFields(logrus.Fields{
+				"error":       err.Error(),
+				"backfill id": id,
```
hsorellana, line 172: probably this key should be underscored, i don't know how it outputs

internal/statestore/backfill.go
akremsa: ok, why not.

internal/statestore/backfill.go
akremsa: agree

aLekSer:
Some E2E tests are failing due to update in `common.go`:
```
	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -53,3 +53,3 @@
        	            	  },
        	            	- (string) (len=21) "pendingreleasetimeout": (string) (len=6) "1000ms",
        	            	+ (string) (len=21) "pendingreleasetimeout": (string) (len=5) "200ms",
        	            	  (string) (len=26) "proposalcollectioninterval": (string) (len=5) "200ms",
```

aLekSer (approved):
Thanks for adding Cleanup based on Acknowledged Backfills.

internal/app/frontend/frontend_service.go
```diff
@@ -354,7 +354,29 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
 // AcknowledgeBackfill is used to notify OpenMatch about GameServer connection info.
 // This triggers an assignment process.
 func (s *frontendService) AcknowledgeBackfill(ctx context.Context, req *pb.AcknowledgeBackfillRequest) (*pb.Backfill, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+	// WARNING: THIS CODE WILL BE REPLCAES BY PR "Frontend acknowledge backfill #1293"
```
aLekSer, line 357:
```suggestion
	// WARNING: THIS CODE WILL BE REPLACED BY PR "Frontend acknowledge backfill #1293"
```

akremsa: it will be replaced anyway 🤣

internal/statestore/backfill.go
```diff
@@ -62,7 +70,7 @@ func (rb *redisBackend) CreateBackfill(ctx context.Context, backfill *pb.Backfil
 		return status.Errorf(codes.AlreadyExists, "backfill already exists, id: %s", backfill.GetId())
 	}
 
-	return acknowledgeBackfill(redisConn, backfill.GetId())
+	return nil
```
aLekSer, line 73: Can you please return this back?

internal/app/frontend/frontend_service.go
```diff
@@ -354,7 +354,62 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
 // AcknowledgeBackfill is used to notify OpenMatch about GameServer connection info.
 // This triggers an assignment process.
 func (s *frontendService) AcknowledgeBackfill(ctx context.Context, req *pb.AcknowledgeBackfillRequest) (*pb.Backfill, error) {
```
sawagh, line 356: I see this method also implemented in https://github.com/googleforgames/open-match/pull/1293/ - Is this accidentally in this PR too? or is this PR dependent on the other one and they need to be merged in order?

internal/app/synchronizer/synchronizer_service.go
```diff
@@ -292,6 +292,11 @@ Registration:
 	<-closedOnCycleEnd
 	stats.Record(ctx, iterationLatency.M(float64(time.Since(cst)/time.Millisecond)))
 
+	err := s.store.CleanupBackfills(ctx)
```
sawagh, line 295: for a single returned err, prefer using the more concise pattern: 

if err := s.store.CleanupBackfills(ctx); err != nil {
    logger.Errorf("Failed to clean up backfills, %s", err.Error())
}

internal/statestore/backfill.go
```diff
@@ -143,6 +151,48 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
 	return nil
 }
 
+// CleanupBackfills removes expired backfills
+func (rb *redisBackend) CleanupBackfills(ctx context.Context) error {
+	redisConn, err := rb.redisPool.GetContext(ctx)
+	if err != nil {
```
sawagh, line 157: use similar shorter error handling pattern as suggested above (also in other places in the code)

internal/statestore/backfill.go
```diff
@@ -143,6 +151,48 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+		return status.Errorf(codes.Unavailable, "CleanupBackfills, failed to connect to redis: %v", err)
+	}
+	defer handleConnectionClose(&redisConn)
+
+	expiredBfIDs, err := rb.GetExpiredBackfillIDs(ctx)
+	if err != nil {
+		return err
```
sawagh, line 164: consider adding logging here?

internal/statestore/backfill.go
```diff
@@ -143,6 +151,48 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+	}
+
+	for _, id := range expiredBfIDs {
+		_, tickets, err := rb.GetBackfill(ctx, id)
+		if err != nil {
+			logger.WithFields(logrus.Fields{
+				"error":       err.Error(),
+				"backfill_id": id,
+			}).Error("cleanup failed to get the backfill")
```
sawagh, line 173: does the logger already add the function name to the log? in that case, the logs could be cleaner and only state what failed ("failed to get the backfill" here)

internal/statestore/backfill.go
```diff
@@ -143,6 +151,48 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+			continue
+		}
+
+		err = rb.DeleteBackfill(ctx, id)
+		if err != nil {
+			logger.WithFields(logrus.Fields{
+				"error":       err.Error(),
+				"backfill_id": id,
+			}).Error("cleanup failed to delete the backfill")
```
sawagh, line 182: if you fail to delete the actual backfill (so it still has Tickets), do you still want to go ahead and delete tickets from Pending Release? would that not leave the system in inconsistent state?

internal/statestore/backfill_test.go
```diff
@@ -604,3 +604,57 @@ func generateBackfills(ctx context.Context, t *testing.T, service Service, amoun
...
+	defer service.Close()
+	ctx := utilTesting.NewContext(t)
+
+	rc, err := redis.Dial("tcp", fmt.Sprintf("%s:%s", cfg.GetString("redis.hostname"), cfg.GetString("redis.port")))
+	require.NoError(t, err)
+
+	bfID := "mockBackfill-1"
+	ticket1ID := "t1"
+	ticket2ID := "t2"
+	ticketIDs := []string{ticket1ID, ticket2ID}
```
sawagh, line 622: nit - since ticket1ID, ticket2ID are not used anywhere else, you might as well do []string{"t1", "t2"}

internal/statestore/backfill_test.go
```diff
@@ -604,3 +604,57 @@ func generateBackfills(ctx context.Context, t *testing.T, service Service, amoun
...
+	bfLastAck := "backfill_last_ack_time"
+	proposedTicketIDs := "proposed_ticket_ids"
+
+	// ARRANGE
```
sawagh, line 626: nit: inconsistent comments through the test. Prefer using complete sentences - you may just choose to describe a comment blurb for the test rather than commenting each method as the names are verbose anyways.

internal/statestore/backfill_test.go
```diff
@@ -604,3 +604,57 @@ func generateBackfills(ctx context.Context, t *testing.T, service Service, amoun
...
+	require.NoError(t, err)
+
+	err = service.AddTicketsToPendingRelease(ctx, ticketIDs)
+	require.NoError(t, err)
+
+	// ACT
+	err = service.CleanupBackfills(ctx)
+	require.NoError(t, err)
+
+	// ASSERT
```
sawagh, line 644: missing any actual assert here?

internal/statestore/backfill_test.go
```diff
@@ -604,3 +604,57 @@ func generateBackfills(ctx context.Context, t *testing.T, service Service, amoun
...
+
+	// backfill doesn't exist anymore
+	_, _, err = service.GetBackfill(ctx, bfID)
+	require.Error(t, err)
+	require.Equal(t, "Backfill id: mockBackfill-1 not found", status.Convert(err).Message())
```
sawagh, line 649: simply checking error message equality is sufficient? it would fail anyways if there is no error.

internal/testing/e2e/backfill_test.go
```diff
@@ -177,31 +178,23 @@ func TestProposedBackfillCreate(t *testing.T) {
...
 	actual, err := om.Frontend().GetBackfill(ctx, &pb.GetBackfillRequest{BackfillId: bfID})
-	require.Nil(t, err)
+	require.NoError(t, err)
 	require.NotNil(t, actual)
 
 	b.Id = actual.Id
 	b.CreateTime = actual.CreateTime
 	require.True(t, proto.Equal(b, actual))
-
-	client, err := om.Query().QueryTickets(ctx, &pb.QueryTicketsRequest{Pool: &pb.Pool{
```
sawagh, line 198: curious why this blurb was removed? seems unrelated to this change itself?

internal/testing/e2e/backfill_test.go
```diff
@@ -318,13 +303,69 @@ func TestBackfillGenerationMismatch(t *testing.T) {
...
+		Config:  om.MMFConfigGRPC(),
+		Profile: &pb.MatchProfile{},
+	})
+	require.NoError(t, err)
+
+	_, err = stream.Recv()
+	require.Error(t, err)
+	e, ok := status.FromError(err)
+	require.True(t, ok)
+	require.Contains(t, e.Message(), "error(s) in FetchMatches call. syncErr=[failed to handle match backfill: 1: rpc error: code = NotFound desc = Backfill id:")
```
sawagh, line 366: Curious about the test logic here. Given your sleep is even before the FetchMatches call - how is that sleep causing backfill expiration? I would expect FetchMatches call to trigger MMF and have a backfill generated and so shouldnt the sleep be after the call?

internal/testing/e2e/common.go
```diff
@@ -162,15 +162,15 @@ func (om *om) MMFConfigHTTP() *pb.FunctionConfig {
 // that parsing bugs can't hide logic bugs.
 const registrationInterval = time.Millisecond * 200
 const proposalCollectionInterval = time.Millisecond * 200
-const pendingReleaseTimeout = time.Millisecond * 200
+const pendingReleaseTimeout = time.Second * 1
```
sawagh, line 165: why change these from 200ms to 1s?

sawagh (changes requested):

internal/app/frontend/frontend_service.go
akremsa: yes, the whole frontend_service.go file was taken from #1293

internal/testing/e2e/common.go
akremsa: This was introduced in #1310 and I needed this change too.

internal/statestore/backfill.go
akremsa: I suggest leaving this one as is, and in other places in statestore too. If I change it to the shorter pattern you suggest then redisConn won't be visible out of this if-scope. Then I need to create a separate variable for that. So I don't think it's worth changing, we follow this pattern everywhere here. This is a matter of personal preference.

sawagh: Sorry. For methods that return value along with error (like here), it is reasonable to have the error check outside (like you have already). 

For methods that return just err, I think the shorter format is concise and readable. However, a quick check tells me we are using either format in the code - not uniform but I am ok with you not changing the format.

internal/statestore/backfill.go
```diff
@@ -200,6 +208,48 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
 	return nil
 }
 
+// CleanupBackfills removes expired backfills
+func (rb *redisBackend) CleanupBackfills(ctx context.Context) error {
```
scosgrave, line 212: Should this function mirror the code in frontend_service.go's doDeleteBackfill? If so, I think you are missing the DeindexBackfill call.  We might also want to find a way to write the code once and re-use it.

internal/testing/e2e/backfill_test.go
scosgrave: I'm pretty sure the wait prior to FetchMatches is valid. The Backfill is created right after the call to CreateTicket at the start of the test. The Backfill is acknowledged right when it is created, and that sets its expiration timestamp, so waiting prior to calling FetchMatches can cause the Backfill to age.
I think this test mirrors the case where a game server has a player leave a game, and it creates the Backfill on its own. Maybe we need another test where the MMF is not able to completely fill a Match, and requests a Backfill to be created?

internal/statestore/backfill.go
scosgrave, line 212: Sorry for asking this here, and not in backfill.go, but I think it is related closely enough:

Why do we expire Backfills at 80% of the pendingReleaseTimeout? I think it isn't very clear that setting pendingReleaseTimeout will have an effect on when Backfills are deleted. I think it would be better for Backfills to have their own explicit expiration timeout in the config, and to update the ticket release process to disassociate the ticket from a Backfill, if it was associated with one.

internal/testing/e2e/backfill_test.go
sawagh: Thanks. That case makes sense - yes I was confusing it with the MMF created backfill case - and +1 to having a test case to cover that.

internal/statestore/backfill.go
akremsa: In case of **returning** an error it is better to log one outside of statestore package.

internal/statestore/backfill.go
akremsa: It does not.
Example of the current output:
`cleanup failed to get the backfill            app=openmatch backfill_id=mockBackfill-1 component=statestore.redis error=QQQ
`

internal/statestore/backfill_test.go
akremsa: ok

internal/statestore/backfill_test.go
akremsa: "Arrange act assert" is a common pattern of writing tests and I've just made the test a bit more readable. You clearly see that `err = service.AddTicketsToPendingRelease(ctx, ticketIDs)`, for example, is a part of an ARRANGE section, which means it is used to prepare data for an actual test. Then I clearly separate a function under the test with ACT section, and so on. So I find these comments quite useful here.

internal/statestore/backfill_test.go
akremsa: see comment above

internal/statestore/backfill_test.go
akremsa: Checking for an error explicitly provides more precise message.
Example:
`Error:      	An error is expected but got nil.`
I clearly see that there should be an error but the function returned nil.

If I leave only comparing error messages I will get:
```        	
Error:      	Not equal: 
expected: "Backfill id: mockBackfill-1 not found"
actual  : ""
```
And here I'm not sure - is error nil or is its description just empty for some reason?

internal/testing/e2e/backfill_test.go
akremsa: Had issues with this test, agreed with @yeukovichd to remove these lines at some point, but this test is updated in #1318 so I will return this code back.

internal/statestore/backfill.go
akremsa: We had a conversation with @Laremere in one of the PRs thread and agreed on using this 80%. 
@aLekSer we do plan to add this value to a separate config parameter, don't we?

akremsa: good catch. I will try to reuse that code.

scosgrave: Ok. Thanks for discussing it with @Laremere. If this was the agreed-upon approach, then I'm fine with it.

internal/statestore/backfill.go
akremsa: we try to delete as much as we can

akremsa:
@scosgrave @sawagh I've updated this PR, could you please check it again?

internal/statestore/backfill.go
```diff
@@ -200,6 +208,77 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
...
+
+	// 1. deindex backfill
+	err = rb.DeindexBackfill(ctx, id)
+	if err != nil {
+		return err
+	}
+
+	// just log errors and try to perform as mush actions as possible
+
+	go func() {
```
yeukovichd, line 227: Why is new goroutine required here?

akremsa: removed

internal/statestore/backfill_test.go
sawagh: I like the test pattern itself - but did not think the comments were necessary. Comments are only useful when they add value not obvious when reading the code. This pattern seems clean enough and the comments // ARRANGE // ACT // ASSERT do not seem to add value. If anything, they do seem misplaced compared to any other comment in production code. 

Did find an interesting blog that goes through this exact pattern and why you might add comments - and why it is a bad idea. Interesting read: https://medium.com/@pjbgf/title-testing-code-ocd-and-the-aaa-pattern-df453975ab80

Readability is subjective. You may leave them in place if you think it makes it more readable. It did not make it any more readable for me.

internal/statestore/backfill_test.go
sawagh: In your sample output, the the actual error being "" is clearly indicative of the error returned being empty right? Still not sure what extra value the extra check adds. The test expects an error - and the returned error is not that - and thats the failure. Unless this error does not catch some output, I do not see the value for the extra check.

internal/testing/e2e/backfill_test.go
sawagh: Will this be returned back in this same PR? Please resolve the comment when it is reverted back. Thanks!

sawagh (approved):
I still have some pending comments - but comments are not blocking. Feel free to check in once Scott signs off as well.

internal/statestore/backfill_test.go
akremsa: Message `actual  : ""` doesn't give an information where it was nil error returned or actual error with an empty string description. `status.Convert(err).Message()` converts nil error to an empty string.

internal/testing/e2e/backfill_test.go
akremsa: This test is already updated in #1318

internal/statestore/backfill_test.go
sawagh: Does it matter if it returned nil - or "" - both of which are unexpected when the expected value is "Backfill id: mockBackfill-1 not found"? What debugging value is added by understanding that it returned nil - over not being sure if it was nil or empty string? I wont block subject my approval over this - but am not convinced that the added check adds any value to the test.

hsorellana:
LGTM 👍

internal/statestore/backfill.go
```diff
@@ -200,6 +208,77 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
 	return nil
 }
 
+// DeleteBackfillCompletely using goroutine performs a set of operations to remove backfill and all related entities.
```
aLekSer, line 211: Does it actually uses go routines?

internal/testing/e2e/backfill_test.go
```diff
@@ -533,6 +534,65 @@ func TestBackfillGenerationMismatch(t *testing.T) {
...
+		require.False(t, ok)
+
+		out <- m.MatchId
+		return nil
+	})
+
+	// wait until backfill is expired, then try to get it
+	time.Sleep(pendingReleaseTimeout * 2)
+
+	// statestore.CleanupBackfills is called at the end of each syncronizer cycle after fetch matches call, so expired backfill will be removed
```
aLekSer, line 579:
```suggestion
	// statestore.CleanupBackfills is called at the beginning of each syncronizer cycle after fetch matches call, so expired backfill will be removed
```

aLekSer:
I have created #1324 ticket to track some contention issues which would arise if we move this logic into go routine. Mostly FetchMatches can return the Backfill to the Director, but there would be no Backfills already if cleanup occurred before this event. We would investigate this deeper in Scale tests.

internal/testing/e2e/backfill_test.go
```diff
@@ -533,6 +534,65 @@ func TestBackfillGenerationMismatch(t *testing.T) {
...
+	stream, err := om.Backend().FetchMatches(ctx, &pb.FetchMatchesRequest{
+		Config:  om.MMFConfigGRPC(),
+		Profile: &pb.MatchProfile{},
+	})
+	require.NoError(t, err)
+
+	_, err = stream.Recv()
+	e, ok := status.FromError(err)
+	require.True(t, ok)
+	require.Contains(t, e.Message(), "error(s) in FetchMatches call. syncErr=[failed to handle match backfill: 1: rpc error: code = NotFound desc = Backfill id:")
```
aLekSer, line 589: @akremsa this should not return an error.

aLekSer: We can make 2 FetchMatches requests.

PR #1299, opened 2020-12-01
Create, Update backfill after MMF run

yeukovichd:
Each time when MMF proposes a match with backfill, the backfill should be saved or updated in the store.


internal/app/backend/backend_service.go
```diff
@@ -330,6 +339,64 @@ func (s *backendService) AssignTickets(ctx context.Context, req *pb.AssignTicket
...
+	}()
+
+	bf, ids, err := store.GetBackfill(ctx, backfill.Id)
+
+	if err != nil {
+		return err
+	}
+
+	if bf.Generation != backfill.Generation {
+		return errors.New("backfill generation mismatch")
```
akremsa, line 391: I think if you specify exact generations here in a log message it will be a bit more informative. 
Like: fmt.Errorf("backfill generation mismatch, current %s, incoming %s", ...)

akremsa (approved):

internal/app/backend/backend_service.go
```diff
@@ -330,6 +339,64 @@ func (s *backendService) AssignTickets(ctx context.Context, req *pb.AssignTicket
 	return resp, nil
 }
 
+func handleMatch(ctx context.Context, match *pb.Match, store statestore.Service) error {
```
Laremere, line 342: `func createOrUpdateBackfill`?

internal/app/backend/backend_service.go
```diff
@@ -330,6 +339,64 @@ func (s *backendService) AssignTickets(ctx context.Context, req *pb.AssignTicket
...
+func doCreateBackfill(ctx context.Context, backfill *pb.Backfill, ticketIds []string, store statestore.Service) error {
+	backfill.Id = xid.New().String()
+	backfill.CreateTime = ptypes.TimestampNow()
+	return store.CreateBackfill(ctx, backfill, ticketIds)
+}
+
+func doUpdateBackfill(ctx context.Context, backfill *pb.Backfill, ticketIds []string, store statestore.Service) error {
+	m := store.NewMutex(backfill.Id)
+	err := m.Lock(ctx)
+
```
Laremere, line 371: nit: removing empty line between `err := ` and `if err != nil` clauses.

internal/testing/e2e/backfill_test.go
```diff
@@ -66,3 +68,134 @@ func TestCreateGetBackfill(t *testing.T) {
...
+				"field": "value",
+			},
+		},
+	}})
+	require.Nil(t, err)
+
+	m := &pb.Match{
+		MatchId:  "1",
+		Tickets:  []*pb.Ticket{t1},
+		Backfill: b,
```
Laremere, line 164: Please test the mmf changing a value on the backfill.

internal/app/backend/backend_service.go
```diff
@@ -330,6 +339,64 @@ func (s *backendService) AssignTickets(ctx context.Context, req *pb.AssignTicket
...
+	}()
+
+	bf, ids, err := store.GetBackfill(ctx, backfill.Id)
+
+	if err != nil {
+		return err
+	}
+
+	if bf.Generation != backfill.Generation {
+		return fmt.Errorf("backfill generation mismatch - expecting: %d, actual: %d", bf.Generation, backfill.Generation)
```
Laremere, line 391: The logic is that we should drop the match, not throw an error.  It would be good to include in the debug logs, but nothing more.

internal/testing/e2e/backfill_test.go
```diff
@@ -66,3 +68,134 @@ func TestCreateGetBackfill(t *testing.T) {
...
+
+	actual, err := om.Frontend().GetBackfill(ctx, &pb.GetBackfillRequest{BackfillId: b.Id})
+	require.Nil(t, err)
+	require.NotNil(t, actual)
+	require.True(t, proto.Equal(b, actual))
+
+	client, err := om.Query().QueryTickets(ctx, &pb.QueryTicketsRequest{Pool: &pb.Pool{
+		StringEqualsFilters: []*pb.StringEqualsFilter{{StringArg: "field", Value: "value"}},
+	}})
+
```
Laremere, line 197: We're also going to want to include making sure the backfill updates in query.  (having queried before so we know it has the old cached version.)  This isn't going to work until several pieces are in place though... just make sure it's on the work items that get done.  (either your tracker, or just an issue on github)

Laremere:
From the CI failure:
`    TestProposedBackfillUpdate: backfill_test.go:186: 
        	Error Trace:	backfill_test.go:186
        	Error:      	Expected nil, but got: &status.Error{e:(*status.Status)(0xc0005b6ea0)}
        	Test:       	TestProposedBackfillUpdate
    TestProposedBackfillUpdate: common.go:52: Evaluator set but never called.
--- FAIL: TestProposedBackfillUpdate (0.06s)`

internal/app/backend/backend_service.go
```diff
@@ -330,6 +339,64 @@ func (s *backendService) AssignTickets(ctx context.Context, req *pb.AssignTicket
...
+		return nil
+	}
+
+	ticketdIds := make([]string, len(match.Tickets))
+
+	for _, t := range match.Tickets {
+		ticketdIds = append(ticketdIds, t.Id)
+	}
+
+	if backfill.Id != "" {
```
Laremere, line 355: This leans in to style and it's totally optional, but consider:

Often when an if statement has both clauses used, (in this case, the rest of this function could be in an "else" clause with no change to behavior) it's better to have the if statement be the "positive" clause.  That is, `if backfill.Id == ""`.  That way when you get to the "else" clause, you're not dealing with a double negative mentally.

Also, when doing that on this function, I think `doCreateBackfill` and `doUpdateBackfill` should be inlined.  It makes for a long function, but that's ok because it's just doing a bunch of stuff that needs to be done in order.   A comment works just as well as a function name, and inlining removes overall complexity (fewer functions.)

(It helps in this case `doCreateBackfill` is relatively short, doing the swap on `if` and then inlining would happen to be relatively clean.  If the length of the results were flipped, I'd probably prefer inlining over simplifying the if statements.  Here we can get best of both worlds, I think...)

yeukovichd: Interesting point of view. Inlined functions and used positive checks.

internal/testing/e2e/backfill_test.go
yeukovichd: You are talking about querying tickets or backfills? Using such code I just want to ensure that ticket associated with backfill will disappear from query results.

internal/testing/e2e/backfill_test.go
yeukovichd: I will check changes to search fields and extensions fields will be applied.

internal/testing/e2e/backfill_test.go
Laremere: Oh actually this just made me realize, when a backfill is rejected because of an update from the game server - we need to release the tickets because they were already marked as pending by the synchronizer.  (or move the marking as pending to the backend after verifying the backfill would work too.)

I was talking about the backfill having a new field value set by the mmf.

Laremere: The work re: unpending tickets because a backfill changed, can be done in a followup PR.

Laremere (approved):

PR #1306, opened 2020-12-10
Release 1.1

jonfoust:
Release-1.1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@
 # If you want information on how to edit this file checkout,
 # http://makefiletutorial.com/
 
-BASE_VERSION = 1.1.0-rc.1
+BASE_VERSION = 1.1.0
 SHORT_SHA = $(shell git rev-parse --short=7 HEAD | tr -d [:punct:])
 BRANCH_NAME = $(shell git rev-parse --abbrev-ref HEAD | tr -d [:punct:])
 VERSION = $(BASE_VERSION)-$(SHORT_SHA)
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@
 # If you want information on how to edit this file checkout,
 # http://makefiletutorial.com/
 
-BASE_VERSION = 1.1.0
+BASE_VERSION = 1.1.0-rc.1
 SHORT_SHA = $(shell git rev-parse --short=7 HEAD | tr -d [:punct:])
 BRANCH_NAME = $(shell git rev-parse --abbrev-ref HEAD | tr -d [:punct:])
 VERSION = $(BASE_VERSION)-$(SHORT_SHA)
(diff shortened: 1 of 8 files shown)


thisisnotapril:
@jonfoust we doing this?

jonfoust:
Should we? The release requires at least one reviewer.

sawagh:
Lets discuss this in the community meeting tomorrow. I am curious as to how the ongoing backfill feature work lines up with this (would prefer cutting 1.1 now so backfill feature has ample of time to bake in 1.2)

Makefile
```diff
@@ -668,16 +668,11 @@ api/api.md: third_party/ build/toolchain/bin/protoc-gen-doc$(EXE_EXTENSION)
 	$(PROTOC) api/*.proto \
 		-I $(REPOSITORY_ROOT) -I $(PROTOC_INCLUDES) \
   		--doc_out=. \
-  		--doc_opt=markdown,api.md
+  		--doc_opt=markdown,api_temp.md
 # Crazy hack that insert hugo link reference to this API doc -)
```
calebatwd, line 672: it seems like this may not be true anymore?

go.mod
```diff
@@ -45,6 +45,7 @@ require (
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
 	github.com/pkg/errors v0.8.1
 	github.com/prometheus/client_golang v1.2.1
+	github.com/pseudomuto/protoc-gen-doc v1.3.2 // indirect
```
aLekSer, line 48: `go mod tidy` invocation removes this indirect dependency.

calebatwd (approved):

go.mod
jonfoust: I'll explore where this is used. I'll make sure if it is not part of the make release target it gets added in a future PR

Makefile
jonfoust: This does look like an additional step. will reproduce and be sure. I believe this change was made and merged in previous PR. If so, I'll reopen and make another change

jonfoust: Reproduced and it seems to hold true.

go.mod
jonfoust: I will either add this to a future PR for either the Release process or make release target. Could not locate any use of it.

jonfoust: Actually I found where it is used. 

https://github.com/googleforgames/open-match/blob/4b3147511bdcc292b711390532fd10ea6a5da8eb/Makefile#L696

jonfoust: This looks to be used for generating documentation for the docs repository

aLekSer: Seems that without go.mod changes should work with this make command:
https://github.com/googleforgames/open-match/blob/fcd590eca689311fa31b5a68c0935c2b11aa913a/Makefile#L561

jonfoust: I see. I will submit a PR with the changes.

PR #1300, opened 2020-12-05
Backfill query implementation

yeukovichd:
**What this PR does / Why we need it**:
PR allows querying and caching backfills.

PR #1310, opened 2020-12-16
Implement backfill querying

yeukovichd:
PR allows querying and caching backfills.

A take over of #1300 and #1280 also has @Laremere comments.

### DIFF
--- a/internal/testing/e2e/common.go
+++ b/internal/testing/e2e/common.go
@@ -162,7 +162,7 @@ func (om *om) MMFConfigHTTP() *pb.FunctionConfig {
 // that parsing bugs can't hide logic bugs.
 const registrationInterval = time.Millisecond * 200
 const proposalCollectionInterval = time.Millisecond * 200
-const pendingReleaseTimeout = time.Millisecond * 200
+const pendingReleaseTimeout = time.Second * 1
 const assignedDeleteTimeout = time.Millisecond * 200
 
 // configFile is the "cononical" test config.  It exactly matches the configmap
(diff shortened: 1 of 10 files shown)


aLekSer:
Coping over the gist of the diff, part of the code where moved into a separate file `cache.go` so it is hard to see that no previous functionality was changed (on github view):
https://gist.github.com/aLekSer/392cf3b4cf8cb2c754b632a635c3e7f4

aLekSer (approved):
Approving as it is the same as previous #1300.

internal/testing/e2e/common.go
```diff
@@ -162,15 +162,15 @@ func (om *om) MMFConfigHTTP() *pb.FunctionConfig {
 // that parsing bugs can't hide logic bugs.
 const registrationInterval = time.Millisecond * 200
 const proposalCollectionInterval = time.Millisecond * 200
-const pendingReleaseTimeout = time.Millisecond * 200
```
sawagh, line 165: Do the backfill tests need a larger timeout? Given that this behavior may impact both backfill / non backfill scenarios - whats the impact of this on the regular e2e tests etc.? 

(I see this being used in TestTicketReleaseByTimeout / TestReleaseTickets etc.)

aLekSer: @yeukovichd addressed this in Slack before:
>The reason why TestProposedBackfillUpdate is failing time to time because sometimes it takes a little bit longer than pendingReleaseTimeout (200ms in tests) to create\update backfill after tickets added to pending release in synchroniser.

PR #1314, opened 2020-12-17
Update location for stable and incubator charts

### LINKED ISSUE #1313: Recent builds failing. Need to update location for stable and incubator charts
**What happened**:
Looking at recent builds that have failed, it seems most of them are failing at the same point. Looking deeper, it seems the charts are not being downloaded.
No need, most builds are failing at the moment

- Here is a link detailing the EOL of the current chart locations used in Open Match:
https://helm.sh/blog/new-location-stable-incubator-charts/

- An update to the minimum Helm version will auto-detect if the repos are correct.

yeukovichd:
Update location for stable and incubator charts.

Closes #1313


aLekSer (approved):
Thanks for fixing CI.

PR #1316, opened 2020-12-17
Release 1.1 fix

jonfoust:
To avoid pulling commits after the 1.1.0-rc release, a fix was needed to update the chart locations in Makefile and several chart files.

Fixes #
#1306 

Changes fixed on master by #1314


aLekSer (approved):
That should work.

As a side note there is a need to clean all tgz archives in `install/helm/open-match`.

PR #1317, opened 2020-12-17
MMF backfill example

yeukovichd:
Shows how MMF can deal with backfills.


yeukovichd:
@googlebot I consent.

aLekSer (approved):
Now we have working MMF using Backfill example 👍

examples/functions/golang/backfill/mmf/matchfunction.go
```diff
@@ -0,0 +1,283 @@
...
+	openSlotsKey    = "open-slots"
+	matchName       = "backfill-matchfunction"
+)
+
+type matchFunctionService struct {
+	grpc               *grpc.Server
+	queryServiceClient pb.QueryServiceClient
+	port               int
+}
+
```
sawagh, line 42: Given this is an example, I would strongly recommend adding comment block on the top giving an overview of whats the behavior of this method. How does it return matches, when does it return backfill matches, when does it populate more tickets into a backfill etc.?

It would make it much simpler to review if you also add small comments on top of each of the helpers - so it is easy to understand how they piece together.

PR #1322, opened 2020-12-22
Fix install-scale-chart target

yeukovichd:
Has fixes required for successful run of install-scale-chart.


aLekSer (approved):
Thanks for fixing `serviceAccount` and `configName` .

PR #1325, opened 2020-12-30
Make redis lock expiration configurable

hsorellana:
Adds new helm vars for feature configuration

### DIFF
--- a/install/helm/open-match/values-production.yaml
+++ b/install/helm/open-match/values-production.yaml
@@ -188,10 +188,10 @@ open-match-core:
   assignedDeleteTimeout: 10m
   # Maximum number of tickets to return on a single QueryTicketsResponse.
   queryPageSize: 10000
+  # Duration for redis locks to expire.
+  backfillLockTimeout: 1m
 
   redis:
-    # Duration for redis locks to expire.
-    mutexTimeout: 5m
     enabled: true
     # If open-match-core.redis.enabled is set to false, have Open Match components talk to this redis address instead.
     # Otherwise the default is set to the om-redis instance.
--- a/install/helm/open-match/values-production.yaml
+++ b/install/helm/open-match/values-production.yaml
@@ -190,6 +190,8 @@ open-match-core:
   queryPageSize: 10000
 
   redis:
+    # Duration for redis locks to expire.
+    mutexTimeout: 5m
     enabled: true
     # If open-match-core.redis.enabled is set to false, have Open Match components talk to this redis address instead.
     # Otherwise the default is set to the om-redis instance.
--- a/internal/statestore/testing/fake_redis.go
+++ b/internal/statestore/testing/fake_redis.go
@@ -45,7 +45,7 @@ func New(t *testing.T, cfg config.Mutable) func() {
 	cfg.Set("redis.pool.maxActive", PoolMaxActive)
 	cfg.Set("redis.pool.idleTimeout", PoolIdleTimeout)
 	cfg.Set("redis.pool.healthCheckTimeout", PoolHealthCheckTimeout)
-	cfg.Set("redis.mutexTimeout", "5m")
+	cfg.Set("backfillLockTimeout", "5m")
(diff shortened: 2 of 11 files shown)


internal/statestore/ticket_test.go
```diff
@@ -901,6 +901,7 @@ func createRedis(t *testing.T, withSentinel bool, withPassword string) (config.V
 	cfg.Set("redis.pool.idleTimeout", time.Second)
 	cfg.Set("redis.pool.healthCheckTimeout", 100*time.Millisecond)
 	cfg.Set("redis.pool.maxActive", 5)
+	cfg.Set("redis.mutexTimeout", "5m")
```
hsorellana, line 904: Once the test passes, will try deleting this change, because i'm not sure if it is useful here

internal/testing/e2e/cluster_test.go
```diff
@@ -22,6 +22,8 @@ import (
 	"os"
 	"testing"
```
aLekSer, line 24: please remove this blank line

internal/statestore/ticket_test.go
hsorellana: Backfill also uses this, so leaving it as it is

akremsa (approved):

install/helm/open-match/values-production.yaml
```diff
@@ -190,6 +190,8 @@ open-match-core:
   queryPageSize: 10000
 
   redis:
+    # Duration for redis locks to expire.
```
yeukovichd, line 193: I think mutextTimeout should follow enabled parameter

internal/testing/e2e/common.go
```diff
@@ -218,6 +218,7 @@ api:
     httpport: "51509"
 
 redis:
+  mutexTimeout: "5m"
```
yeukovichd, line 221: Is it better to decrease 5m value to just 1m?

hsorellana: Ok, i'll change it

install/helm/open-match/values-production.yaml
hsorellana: and when disabled, how should it be handled in the code??

yeukovichd: Yes, you are right. I suppose if section is not enabled that means all parameters defined in it are not used at all? Maybe it is better to move that parameter to open-match-core? and name parameter like backfillLockTimeout?

hsorellana: Maybe put this configuration before enabled is kind of misleading. On the other hand, i like your suggestion, will change it

internal/statestore/testing/fake_redis.go
```diff
@@ -45,6 +45,7 @@ func New(t *testing.T, cfg config.Mutable) func() {
 	cfg.Set("redis.pool.maxActive", PoolMaxActive)
 	cfg.Set("redis.pool.idleTimeout", PoolIdleTimeout)
 	cfg.Set("redis.pool.healthCheckTimeout", PoolHealthCheckTimeout)
+	cfg.Set("backfillLockTimeout", "5m")
```
yeukovichd, line 48: Hugo could you set 1m everywhere please?

hsorellana: oh i missed this one, fixing it

hsorellana:
@yeukovichd @aLekSer this PR is ready, for one last check and merge 👍

yeukovichd (approved):

Laremere:
This PR of yours adds `backfillLockTimeout`, but #1326 implements the same behavior with `mutexTimeout` while also adding `backfillTimeToLive`.  What's the overall intention here?

hsorellana:
@Laremere both PRs adds two variables: `backfillLockTimeout` (previously `mutexTimeout`) and `backfillTimeToLive`. #1326  is dependent on this PR, since it is outdated with the newest changes done in this PR

Laremere (approved):

PR #1326, opened 2021-01-04
Add backfill TTL config variable

hsorellana:
This PR adds a new variable in Helm charts for configuring the timeout for Backfill TTL

This PR should be merged after #1325 is merged and resolving all conflicts, if any.


Laremere:
I prefer the timeout being a fraction of the ticket pending timeout, as it is requires less configuration, and it removes a possible misconfiguration causing bugs.  Is there a compelling reason to add this?

hsorellana:
as i understood it, this variable goes in replacement because, right now, it is used another variable for this use case, which wasn't meant to be, in the first place

hsorellana:
@Laremere i was thinking about adding the `backfillTimeToLive` variable in a way that it's use like this:

```go
func getBackfillReleaseTimeout(cfg config.View) time.Duration {
	return cfg.GetDuration("pendingReleaseTimeout") * cfg.GetFloat64("backfillTimeToLive")
}
```

so that way, in either values.yaml files, can be defined how much percentage of `pendingReleaseTimeout` should be applied to `backfillTimeToLive`, but everything configurable can be source of bugs.

What do you think about this??

aLekSer:
@hsorellana there might be `backfillTTLPortion` parameter which is in between `(0; 80%)` interval. If something bigger than this  you can default it to `80%` with warning message, that one of parameters was ignored.

hsorellana:
@aLekSer nice, i like that approach

Laremere:
Sorry - still don't see the need for this variable?

It's okay if the ticket pending timeout is used for this as well - they're both related to how long OM should wait before giving up on game servers.  Perhaps we should document the config file a bit better, but other than that I don't see any value in having this as a configuration option - when would someone ever want to change it from the default?

aLekSer:
This parameter was requested by @scosgrave Sean and the team.

internal/statestore/ticket_test.go
```diff
@@ -903,6 +903,7 @@ func createRedis(t *testing.T, withSentinel bool, withPassword string) (config.V
 	cfg.Set("redis.pool.maxActive", 5)
 	cfg.Set("backfillLockTimeout", "1m")
 	cfg.Set("pendingReleaseTimeout", "200ms")
+	cfg.Set("backfillTimeToLive", "800ms")
```
aLekSer, line 906: @hsorellana I thought you would change this `800ms` time to some per cent value and rename this `backfillTimeToLive` to something like `backfillTTL` .

aLekSer (changes requested):
Changes in this PR are inconsistent

aLekSer:
This should hold: `AcknowledgeBackfill time < cleanupBackfillTTL < pendingRelease timeout`.
Otherwise backfill would be expired, or for the second we have some regular tickets associated but already returned to the pool of available tickets.

scosgrave:
> This parameter was requested by @scosgrave Sean and the team.

I think we requested this because we did not know tickets would be released from the backfill when pendingRelease timeout happened. Having seen the work you have done, and the discussion here, I agree that backfillTTL should not be a independent config setting. We should make it a percentage of the pendingRelease timeout, and make sure we document that there is a relationship between the two. In addition, we should make it easy for those running Open Match to determine how often their game servers should acknowledge backfills. Maybe we should print out the backfillTTL at startup?

PR #1328, opened 2021-01-04
Backfill scenario in scale tests

yeukovichd:
Extends current scale tests by new backfill scenario.


examples/scale/scenarios/backfill/backfill.go
```diff
@@ -0,0 +1,414 @@
...
+}
+
+func runAcknowledgeBackfills(fe pb.FrontendServiceClient, matchesToAcknowledge <-chan *pb.Match, logger *logrus.Entry) {
+	for m := range matchesToAcknowledge {
+		ids := []string{}
+		for _, t := range m.Tickets {
+			ids = append(ids, t.GetId())
+		}
+
+		err := doAcknowledgeBackfill(fe, m.Backfill.Id)
```
aLekSer, line 203: We have only one `AcknowledgeBackfill` per Backfill returned by FetchMatches. In the real case there would be series of `AcknowledgeBackfill`. Say every minute.

aLekSer: This is the fastest way to feel the match. So probably OK for now.

examples/scale/backend/backend.go
```diff
@@ -39,17 +40,6 @@ var (
 	})
 
 	activeScenario = scenarios.ActiveScenario
-
```
Laremere, line 42: What's the reason for moving this into a separate package?  Just so the backfill scenario can access it?

Laremere (changes requested):
The intent behind the code structure of the scale tests is that the game logic is separated from infrastructure logic.  For backfill I would like that to continue.  You can see in the scenario there's a "GameScenario", the only thing added there should be what a backfill created on the frontend looks like, I think?

Specifically, extending the interface to have different backend and frontend implementations is not the desired way to go with this code.  The scenario object created should turn on or off functionality in the frontend/backend.

For the backend, I don't think there needs to be any new parameters - the director knows whether it was a backfill match based on the presence of backfill, and can choose behavior based on that.

Backfill acknowledgement should work on a loop for each backfill.  Though if there's a scenario where the backfill fills up, the "game server" can choose to delete the backfill instead of acknowledging it.  It should only trigger this loop if "allocate server"" is true on the match, otherwise the backend can basically do nothing (other than note the values for the metrics.)

Laremere:
Oh, and willing to discuss more if that doesn't make sense, or if the existing structure isn't allowing my suggestions to work - we may need to refactor this a little more to handle the new scenarios.

examples/scale/backend/backend.go
yeukovichd: Yes, I moved metrics into separate package because I wanted to reuse them.

yeukovichd:
> Oh, and willing to discuss more if that doesn't make sense, or if the existing structure isn't allowing my suggestions to work - we may need to refactor this a little more to handle the new scenarios.

examples/scale/scenarios/firstmatch/firstmatch.go
```diff
@@ -50,7 +51,11 @@ func (_ *FirstMatchScenario) Ticket() *pb.Ticket {
 	return &pb.Ticket{}
 }
 
-func (_ *FirstMatchScenario) MatchFunction(p *pb.MatchProfile, poolTickets map[string][]*pb.Ticket) ([]*pb.Match, error) {
+func (_ *FirstMatchScenario) Backfill() *pb.Backfill {
```
hsorellana, line 54: in this case, if the "instance" for FirstMatchScenario is not used, simply set (*FirstMatchScenario) as owner of the func. I think is cleaner

Laremere:
Yes the logic shouldn't be in the game scenario.
I'm not convinced that a new interface is necessary.  If everything for, eg, the frontend were in the frontend package, then you wouldn't need to move frontend metrics to another package.  Also most of the logic is the same, so I'm worried about duplicating too much with different infrastructure

Issue #1142, opened 2020-03-03
Several redis items still installed with redis disabled


Laremere:
Reporting by Luna on slack:

> I also noticed even with redis disabled I'm still getting the following objects:
om-redis-podsecuritypolicy
om-redis-role
om-redis-role-binding

yfei1:
This issue has been there for a while, I haven't figured out a neat way to resolve this yet.

PR #1330, opened 2021-01-10
Fix for Helm chart to use custom Redis

SealTV:
It is currently not possible to use Helm Chart with a custom Redis instance.

The error happens when I try to deploy Helm Chart with the following config:
```yaml values-open-match.yaml
open-match-core:
  enabled: true

  redis:
    enabled: false
    hostname: my_redis_host
    port: 6379
    user: ""
```

I get the error:
```bash
$ helm upgrade open-match ./open-match/install/helm/open-match --install --namespace default --create-namespace -f ./values-open-match.yaml --wait --timeout 10m
#Release "open-match" does not exist. Installing it now.
#Error: template: open-match/templates/service-account.yaml:107:11: executing "open-match/templates/service-account.yaml" at <include "call-nested" (list . "redis" "redis.serviceAccountName")>: error calling include: template: open-match/templates/_helpers.tpl:225:4: executing "call-nested" at <include $template (dict "Chart" (dict "Name" (last $subchart)) "Values" $values "Release" $dot.Release "Capabilities" $dot.Capabilities)>: error calling include: template: no template "redis.serviceAccountName" associated with template "gotpl"
```

This PR fix the error.


Laremere (approved):

aLekSer (approved):

aLekSer:
This fix is part of fixing this issue: https://github.com/googleforgames/open-match/issues/1142

PR #1331, opened 2021-01-11
Fix proto comments for better markdown output

hsorellana:
Improves the definition in protobuf objects so when the autogenerated documentation has an improved display of it


api/frontend.proto
```diff
@@ -84,25 +84,31 @@ message WatchAssignmentsResponse {
...
 // to possible change or removal.
 message CreateBackfillRequest {
+  // An empty Backfill object.
   Backfill backfill = 1;
 }
 
 // BETA FEATURE WARNING: This Request message is not finalized and still subject
 // to possible change or removal.
 message DeleteBackfillRequest {
+  // An existing Backfill ID.
```
HazWard, line 104: Just a personal preference but `ID of backfill ticket to delete`. The other could follow the same pattern but otherwise it looks good to me.

hsorellana: nice, i like it. Will change it

api/messages.proto
```diff
@@ -239,9 +236,10 @@ message Match {
   reserved 5, 6;
 }
 
+// Represents a backfill entity which is used to fill partially full matches.
+//
```
Laremere, line 240: You removed this newline between the comment and the beta feature warning elsewhere, why add it here?

hsorellana: My initial intention was to keep a separate line between object description and the beta feature warning but i realized this only works at top level. If i put a separate line in an inner field, the markdown generated has tables broken visually

Laremere (approved):

PR #1332, opened 2021-01-11
Add help comments in Makefile

hsorellana:
Adds description to make command and displays it when running make help


Makefile
```diff
@@ -739,6 +746,10 @@ define fast_test_folder
 	$(foreach dir, $(wildcard $(1)/*/.), $(call fast_test_folder, $(dir)))
 endef
 
+## ####################################
```
aLekSer, line 749: This should be before `define test_folder`

aLekSer (changes requested):

Makefile
aLekSer: Even before `build:`

Makefile
```diff
@@ -699,6 +699,13 @@ api/%.swagger.json: api/%.proto third_party/ build/toolchain/bin/protoc$(EXE_EXT
 		-I $(REPOSITORY_ROOT) -I $(PROTOC_INCLUDES) \
 		--swagger_out=logtostderr=true,allow_delete_body=true:$(REPOSITORY_ROOT)
 
+## ####################################
+## # Utilities
+##
```
aLekSer, line 704: No need for this since it already belongs to:
``` 
## # Protobuf
```
section
```suggestion
```

Makefile
hsorellana: ohh you're right!

aLekSer (approved):
Thanks, now it is better.

PR #1334, opened 2021-01-13
Move Cleanup Backfills after main SynchronizerCycle & add workers pool

### LINKED ISSUE #1324: Consider using go routine for CleanupBackfills function

In #1297 PR `CleanupBackfills()` function call was added at the beginning of the Synchronizer cycle. If number of backfills is 0 - if we are not using backfills there should be no difference to the normal Synchronizer flow. But if something goes wrong and number of Backfills would grow rapidly and `CleanupBackfills` will be slow, this could lead to problems with invoking other go routines in Synchronizer, pre-backfill one.

This should be done in a go routine:
```
	err := s.store.CleanupBackfills(ctx)
	if err != nil {
		logger.Errorf("Failed to clean up backfills, %s", err.Error())
	}
```
Similar to all other calls:
```
	go func() {
		fanInFanOut(m2c, m3c, m6c)
		// Close response channels after all responses have been sent.
		for _, r := range registrations {
			close(r.m7c)
		}
	}()
```
Locks on Backfills should help to avoid issues when Backend would update Backfill simultaneously, as it is acquiring the lock first and `GetBackfill` would return `NotFound` error in this case - no `Backfill` would be created (in update which performs a set) after cleanup.

Putting CleanupBackfills at the end of the synchronizer cycle.

**Additional context**
There could be some issues with E2E tests after this changes. @akremsa can comment more on what kind of errors were met on CI.

akremsa:
During the contributor's meeting, we agreed with @Laremere to follow this pattern in the cleanup backfills process.
Move CleanupBackfills after main SynchronizerCycle. 

There are some benchmarks:

**Using workers pool**
workers=5; expired backfills=50
```BenchmarkCleanupBackfills-12    	      40	  28685354 ns/op	15493765 B/op	   90618 allocs/op```

workers=3; expired backfills=50
```BenchmarkCleanupBackfills-12    	      34	  29910474 ns/op	15493394 B/op	   90614 allocs/op```

workers=1; expired backfills=50
```BenchmarkCleanupBackfills-12    	      26	  40829001 ns/op	15492384 B/op	   90604 allocs/op```

workers=5; expired backfills=200
```BenchmarkCleanupBackfills-12    	       9	 114774808 ns/op	61991684 B/op	  362357 allocs/op```

workers=3; expired backfills=200
```BenchmarkCleanupBackfills-12    	       8	 130279702 ns/op	61985000 B/op	  362315 allocs/op```

workers=1; expired backfills=200
```BenchmarkCleanupBackfills-12    	       7	 200287135 ns/op	61978501 B/op	  362230 allocs/op```

**Previous straightforward implementation**
expired backfills=50
```BenchmarkCleanupBackfills-12    	      25	  41920697 ns/op	15491571 B/op	   90599 allocs/op```

expired backfills=200
```BenchmarkCleanupBackfills-12    	       6	 186510753 ns/op	61971938 B/op	  362200 allocs/op```

As you can see, results with worker=1 are pretty close to not using the worker at all.
Also, there is no big difference between 5 and 3 workers, taking into account that all benchmarks have a margin of error. 

Closes #1324 .

### DIFF
--- a/internal/app/synchronizer/synchronizer_service.go
+++ b/internal/app/synchronizer/synchronizer_service.go
@@ -212,11 +212,6 @@ func (s *synchronizerService) runCycle() {
 	/////////////////////////////////////// Initialize cycle
 	ctx, cancel := contextcause.WithCancelCause(context.Background())
 
-	err := s.store.CleanupBackfills(ctx)
-	if err != nil {
-		logger.Errorf("Failed to clean up backfills, %s", err.Error())
-	}
-
 	m2c := make(chan mAndM6c)
 	m3c := make(chan *pb.Match)
 	m4c := make(chan *pb.Match)
@@ -296,6 +291,11 @@ Registration:
 	})
 
 	<-closedOnCycleEnd
+	err := s.store.CleanupBackfills(ctx)
+	if err != nil {
+		logger.Errorf("Failed to clean up backfills, %s", err.Error())
+	}
+
 	stats.Record(ctx, iterationLatency.M(float64(time.Since(cst)/time.Millisecond)))
 
 	// Clean up in case it was never needed.
--- a/internal/app/synchronizer/synchronizer_service.go
+++ b/internal/app/synchronizer/synchronizer_service.go
@@ -291,15 +291,16 @@ Registration:
 	})
 
 	<-closedOnCycleEnd
-	err := s.store.CleanupBackfills(ctx)
-	if err != nil {
-		logger.Errorf("Failed to clean up backfills, %s", err.Error())
-	}
 
 	stats.Record(ctx, iterationLatency.M(float64(time.Since(cst)/time.Millisecond)))
 
 	// Clean up in case it was never needed.
 	cancelProposalCollection.Stop()
+
+	err := s.store.CleanupBackfills(ctx)
+	if err != nil {
+		logger.Errorf("Failed to clean up backfills, %s", err.Error())
+	}
 }
 
 ///////////////////////////////////////
--- a/internal/statestore/backfill.go
+++ b/internal/statestore/backfill.go
@@ -306,7 +306,6 @@ func (rb *redisBackend) CleanupBackfills(ctx context.Context) error {
 	wg.Add(len(expiredBfIDs))
 	backfillIDsCh := make(chan string, len(expiredBfIDs))
 
-	// TODO: move workers amount to a config parameter
 	for w := 1; w <= 3; w++ {
 		go rb.cleanupWorker(ctx, backfillIDsCh, &wg)
 	}
--- a/internal/statestore/backfill.go
+++ b/internal/statestore/backfill.go
@@ -17,6 +17,7 @@ package statestore
 import (
 	"context"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by akremsa** [comment]
@Laremere, @yeukovichd  let's discuss pros and cons of putting cleanup method at the beginning or at the end of a sync cycle.
We agreed to put it at the beginning because this approach guarantees that we work only with relevant backfills and non of them will be deleted during MMF execution.
We had lots of issues with tests where cleanup, when it was put at the end, is involved. It's really hard to predict how long will it take to delete everything and introducing lots of sleep calls brings more complexity and doesn't make this whole sync and testing process stable.

**Comment by aLekSer** [comment]
Also this is discussed here #1324 in more details.

**Comment by Laremere** [comment]
The reason to move it to after the cycle is because that's when the synchronizer isn't being waited upon, and is doing little work.  The backfill will have been a part of many cycles before the cleanup, so I don't see improving by one cycle to be that helpful.

The testing problem can be solved by:
1. Create backfill
2. Wait till it should expire
3. Call fetch matches, having mmf and evaluator do nothing.
4. Call fetch matches again
5. backfill should be gone.

Since backfill cleanup will finish before the next fetch matches can run, running fetch matches a second time guarantees that the cleanup which should clean up the backfill has finished.

File: internal/app/synchronizer/synchronizer_service.go [hunk 1]
```diff
@@ -296,6 +291,11 @@ Registration:
...
 
 	<-closedOnCycleEnd
+	err := s.store.CleanupBackfills(ctx)
+	if err != nil {
+		logger.Errorf("Failed to clean up backfills, %s", err.Error())
+	}
+
 	stats.Record(ctx, iterationLatency.M(float64(time.Since(cst)/time.Millisecond)))
 
 	// Clean up in case it was never needed.
```
yeukovichd, line 301: What will happen if cancelProposalCollection fires before cleanup finishes its work?

internal/statestore/backfill.go
```diff
@@ -280,22 +281,42 @@ func (rb *redisBackend) DeleteBackfillCompletely(ctx context.Context, id string)
...
 	expiredBfIDs, err := rb.GetExpiredBackfillIDs(ctx)
 	if err != nil {
 		return err
 	}
 
+	var wg sync.WaitGroup
+	wg.Add(len(expiredBfIDs))
+	backfillIDsCh := make(chan string, len(expiredBfIDs))
+
+	// TODO: move workers amount to a config parameter
```
aLekSer, line 309:
```suggestion
	// TODO TicketNumber: move workers amount to a config parameter
```

internal/app/synchronizer/synchronizer_service.go
akremsa: moved CleanupBackfills to the very end of the function

internal/statestore/backfill.go
```diff
@@ -280,22 +281,41 @@ func (rb *redisBackend) DeleteBackfillCompletely(ctx context.Context, id string)
 	return nil
 }
 
+func (rb *redisBackend) cleanupWorker(ctx context.Context, backfillIDsCh <-chan string, wg *sync.WaitGroup) {
+	var err error
+	for id := range backfillIDsCh {
```
aLekSer, line 286: I think no need to add the type into the name here:
```suggestion
	for id := range backfillIDs {
```

aLekSer: Seems that this is minor I will remove

internal/testing/e2e/backfill_test.go
```diff
@@ -581,17 +592,27 @@ func TestCleanUpExpiredBackfills(t *testing.T) {
 	// wait until backfill is expired, then try to get it
 	time.Sleep(pendingReleaseTimeout * 2)
 
-	// statestore.CleanupBackfills is called at the beginning of each syncronizer cycle after fetch matches call, so expired backfill will be removed
+	// statestore.CleanupBackfills is called at the end of each syncronizer cycle after fetch matches call, so expired backfill will be removed
```
aLekSer, line 595:
```suggestion
	// statestore.CleanupBackfills is called at the end of each synchronizer cycle after fetch matches call, so expired backfill will be removed
```

akremsa:
@Laremere please review this PR again

aLekSer (approved):
Now regular Synchronizer cycle and CleanupBackfills are performed in parallel.

PR #1335, opened 2021-01-14
Expired backfills can not be updated or acknowledged

akremsa:
Previously it was possible to "resurrect" expired backfills. With this fix, you are not allowed to do that.

### DIFF
--- a/internal/statestore/backfill.go
+++ b/internal/statestore/backfill.go
@@ -188,6 +188,15 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
 	}
 	defer handleConnectionClose(&redisConn)
 
+	expired, err := isBackfillExpired(redisConn, backfill.Id, getBackfillReleaseTimeout(rb.cfg))
+	if err != nil {
+		return err
+	}
+
+	if expired {
+		return status.Errorf(codes.Unavailable, "can not update an expired backfill, id: %s", backfill.Id)
+	}
+
 	bf := ipb.BackfillInternal{
 		Backfill:  backfill,
 		TicketIds: ticketIDs,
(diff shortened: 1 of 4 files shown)


internal/statestore/backfill_test.go
```diff
@@ -115,57 +118,131 @@ func TestCreateBackfill(t *testing.T) {
...
 	cfg, closer := createRedis(t, false, "")
 	defer closer()
 	service := New(cfg)
 	require.NotNil(t, service)
 	defer service.Close()
 	ctx := utilTesting.NewContext(t)
 
-	bf := pb.Backfill{
-		Id:         "1",
+	// ARRANGE
```
aLekSer, line 129: Hard to understand this comment what to Arrange?

internal/statestore/backfill.go
```diff
@@ -208,6 +217,17 @@ func (rb *redisBackend) UpdateBackfill(ctx context.Context, backfill *pb.Backfil
 	return nil
 }
 
+func isBackfillExpired(conn redis.Conn, id string, ttl time.Duration) (bool, error) {
```
aLekSer, line 220: May be it is better to remove last argument as it is not changed anywhere and the ttl duration is always has one value.

aLekSer, line 220:
```suggestion
func isBackfillExpired(conn redis.Conn, id string) (bool, error) {
	ttl:=  getBackfillReleaseTimeout(rb.cfg)

```

akremsa: Then I need to link this function with redisBackend struct.
I will leave it as is.

aLekSer:
Thanks, Alexey. Code is great, but E2E tests should be fixed. When I run 20 times I see no passes, while in master E2E passes quite often:
```
for i in {0..20}; do go test -run Backfill -count 1 "open-match.dev/open-match/internal/testing/e2e" ; done >> ./test_results.log
```

aLekSer (approved):
Approving, code is needed to avoid Backfill zombies, when expired backfill would be not expired if it acknowledged before Cleanup find it.

aLekSer:
One more run with 20 iterations gives me no failed runs. This is strange.

PR #1338, opened 2021-01-15
Release process update

jonfoust:
- Update steps to move "Use build artifacts" later in the process
- Adds step to publish release notes to Open Match blog


aLekSer (approved):
Thanks for adding the blog publish step.

PR #1339, opened 2021-01-16
Add backfill scenario to scale tests

yeukovichd:
Backfill scenario implementation in scale tests.

In order to run scale tests the following should be done:
1. create GKE cluster using ```make create-gke-cluster```
2. push container images to GCR using ```make push-images```
3. update values-production.yaml in install/helm/open-match with values required for tests
4. install helm chart using ```make install-scale-chart```
5. run ```make proxy``` in order to establish connection with Grafana
6. open http://localhost:3000, log into Grafana, open Scale dashboard and see charts


examples/scale/frontend/frontend.go
```diff
@@ -89,13 +107,24 @@ func runner(fe pb.FrontendServiceClient) {
...
 	if err != nil {
 		logger.WithError(err).Error("failed to create a ticket")
 		return
 	}
 
-	_ = id
+	err = watchAssignments(ctx, fe, ticketToWatch{id: id, createdAt: createdAt})
+	if err != nil {
+		logger.WithError(err).Errorf("failed to get ticket assignment: %s", id)
+	} else {
```
hsorellana, line 120: i think this else block can be dropped and move the code to the top level, right??

examples/scale/scenarios/backfill/backfill.go
```diff
@@ -0,0 +1,254 @@
...
+
+	any, err := ptypes.MarshalAny(&wrappers.Int32Value{Value: int32(val)})
+	if err != nil {
+		panic(err)
+	}
+
+	b.Extensions[openSlotsKey] = any
+}
+
+func statelessMMF(p *pb.MatchProfile, poolBackfills map[string][]*pb.Backfill, poolTickets map[string][]*pb.Ticket, ticketsPerMatch int) ([]*pb.Match, error) {
```
Laremere, line 176: What's this method for?  Not used anywhere?

Laremere (changes requested):

examples/scale/frontend/frontend.go
yeukovichd: Can't agree with this because I measure ticket assignment time when call watchAssignments finishes successfully only.

examples/scale/scenarios/backfill/backfill.go
yeukovichd: I tried to implement 2 different backfill scenarios:
1. N backfills created initially by frontend and MMF takes that backfills and fills them. After that backfills are acknowledged by backend. (statelessMMF)
2. MMF creates backfill and fill it with N tickets on the first FetchMatches. Updates backfill's extensions with remaining open slots (= K - N where K - number of tickets for full match). After that backfill is acknowledged by backend. On the second FetchMatches MMF takes previously created backfill and tries to fill it with remaining tickets. After that backfill is acknowledged and deleted by backend. (statefullMMF)

In order to use 1 or 2 scenario I should configure Scenario struct in scenarios.go and call proper MMF implementation in MatchFunction function.

Maybe namings statefullMMF and statelessMMF are not good here.

syntxerror:
@yeukovichd I've overlooked and re-reviewed this and would like to add it to the release. Do you mind merging from current master so I may merge.

syntxerror (approved):
LGTM

PR #1341, opened 2021-01-18
Backfill: Skip not found errors on Backend

aLekSer:
There could be the case when backfill returned by the MMF was deleted
in CleanupBackfills.

We need neither return an error if UpdateBackfill failed due to deleted Backfill , nor return the match.

Relates to #1334


Laremere (approved):

aLekSer:
I have added a UT, similar to what was before (`TestCleanUpExpiredBackfills`), but now deletion of BF is performed manually.

PR #1346, opened 2021-01-21
Add seed to Ticket and Backfill ID generation

aLekSer:
Add time as seed to xid algorithm.

There could be some errors with tickets always have the same IDs after redeploying OM.


aLekSer:
Sorry, seems that this code is equivalent.
I am trying to fix this error in match function:
```
      "Status": "Encountered error: rpc error: code = Unknown desc = error(s) in FetchMatches call. syncErr=[error receiving match from synchronizer: rpc error: code = Unknown desc = error calling evaluator: multiple match functions used same match_id: \"profile-backfill-matchfunction-time-2021-01-21T17:55:27.65-num-1-c04s\"], mmfErr=[\u003cnil\u003e]",
 ```

PR #1347, opened 2021-01-21
Fix minor typos

aLekSer:
There was typos in new Backfill function matchId name.

If we are using two profiles with the same Backfill MatchFunction it would produce same matchIds and generates such error in Synchronizer:
```
      "Status": "Encountered error: rpc error: code = Unknown desc = error(s) in FetchMatches call. syncErr=[error receiving match from synchronizer: rpc error: code = Unknown desc = error calling evaluator: multiple match functions used same match_id: \"profile-backfill-matchfunction-time-2021-01-21T17:55:27.65-num-1-c04s\"], mmfErr=[\u003cnil\u003e]",
```

### DIFF
--- a/examples/functions/golang/backfill/mmf/matchfunction.go
+++ b/examples/functions/golang/backfill/mmf/matchfunction.go
@@ -254,7 +254,7 @@ func newMatch(num int, profile string, tickets []*pb.Ticket, b *pb.Backfill) pb.
 	t := time.Now().Format("2006-01-02T15:04:05.00")
 
 	return pb.Match{
-		MatchId:       fmt.Sprintf("profile-%s-time-%s-num-%d", matchName, t, num),
+		MatchId:       fmt.Sprintf("profile-%s-time-%s-num-%d", profile, t, num),
 		MatchProfile:  profile,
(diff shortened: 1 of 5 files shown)


HazWard (approved):

examples/functions/golang/backfill/mmf/matchfunction.go
```diff
@@ -254,7 +254,7 @@ func newMatch(num int, profile string, tickets []*pb.Ticket, b *pb.Backfill) pb.
 	t := time.Now().Format("2006-01-02T15:04:05.00")
 
 	return pb.Match{
-		MatchId:       fmt.Sprintf("profile-%s-time-%s-num-%d", matchName, t, num),
+		MatchId:       fmt.Sprintf("profile-%s-time-%s-num-%d", profile, t, num),
```
HazWard, line 257: Instead of profile you could also use an ID like we use for tickets.

aLekSer: As I understand from xid code, they use proximity as Unix() seconds and add global counter. So for multiple MMF pods we can have same xids generated.

PR #1350, opened 2021-01-27
Fix WatchAssignments causes memory leaks

### LINKED ISSUE #1327: WatchAssignments causes memory leaks
**What happened**:
Frontend service has WatchAssignments action to watch assignments on ticket. That action uses backoff strategy to continuously check if ticket is assigned. The problem is that WatchAssignments gets stuck in endless loop. There are no graceful ways to finish WatchAssignments request. But it can be finished if: 1) ticket is deleted 2) ticket get new assignment and client connection is closed. In other cases it will work and wasting server resources.

I think behaviour of WatchAssignments should be changed. Request should be completed right after ticket assignment is streamed to the client.

The problem can be reproduce by running the following test:

```
func TestWatchAssignments(t *testing.T) {
	om := newOM(t)
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	t1, err := om.Frontend().CreateTicket(ctx, &pb.CreateTicketRequest{Ticket: &pb.Ticket{}})
	require.NoError(t, err)

	{
		req := &pb.AssignTicketsRequest{
			Assignments: []*pb.AssignmentGroup{
				{
					TicketIds:  []string{t1.Id},
					Assignment: &pb.Assignment{Connection: "a"},
				},
			},
		}
		resp, err := om.Backend().AssignTickets(ctx, req)
		require.NoError(t, err)
		require.Len(t, resp.Failures, 0)
	}

	{
		stream, err := om.Frontend().WatchAssignments(ctx, &pb.WatchAssignmentsRequest{TicketId: t1.Id})
		require.NoError(t, err)

		var a *pb.Assignment
		for a.GetConnection() == "" {
			resp, err := stream.Recv()
			require.NoError(t, err)

			a = resp.Assignment
		}

		require.Equal(t, "a", a.Connection)
	}
}
```

yeukovichd:
Has a fix which provides client option to cancel WatchAssignments request
#1327 
Closes #1327


aLekSer (approved):
Thanks for the fix.
@Laremere can you review it also?

internal/app/frontend/frontend_service.go
```diff
@@ -294,6 +294,10 @@ func doWatchAssignments(ctx context.Context, id string, sender func(*pb.Assignme
 	var currAssignment *pb.Assignment
 	var ok bool
 	callback := func(assignment *pb.Assignment) error {
+		if ctx.Err() == context.Canceled || ctx.Err() == context.DeadlineExceeded {
```
Laremere, line 297: No need to check the error type, just if err != nil, return it.

Laremere (approved):

Issue #1354, opened 2021-02-04
API change: Return 404 Not Found on DELETE ticket/backfill ticket requests when the ticket does not exist


lucy-yuan:
## Overview

Currently, Open Match returns a success response 200 on the DELETE requests for ticket (`/v1/frontendservice/tickets/{ticket_id}`) and backfill ticket (`/v1/frontendservice/backfills/{backfill_id}`), regardless whether the ticket was actually found. Proposal to return 404 Not Found in these situations.

## Motivation

* To align with the GET requests on the same endpoints, which return 404 Not Found when the resource is not found.
* To provide more visibility on the DELETE request.

## Impact

Clients that expect a successful response to a delete ticket request when sending an ID that does not exist or when sending the same ID multiple times will be impacted.

## Change Proto

Proto is the same for successful requests. 404 Not Found response code to be returned when requested ticket not found.

Please see PR #1352 for proposed changes.

HazWard:
@jonfoust Do you know what's the process when introducing breaking changes?

jonfoust:
@HazWard There is a bunch of things to consider. Normally it would go under review and discussion before being approved if it were breaking the main branch. This change seems to be for backfill which is considered beta at this point. Happy to discuss other approaches to breaking changes.

lucy-yuan:
PR has been merged.

PR #1352, opened 2021-02-01
Return 404 when deleting ticket/backfill ticket that does not exist

lucy-yuan:
**What this PR does / Why we need it**:

Delete requests will return 404 instead of 200 when the resource (ticket) is not found. This does not break the concept of idempotency and will align with get requests as well as the frontend swagger configuration.

PR #1356, opened 2021-02-23
Return 404 when deleting ticket/backfill ticket that does not exist (frontend)

lucy-yuan:

Related to PR #1352 and issue #1354.

The linked PR returned NotFound from `statestore` for delete ticket, but does not actually handle this error on `frontend_service`. This PR is to return the Notfound error on the frontend so that a 404 will be returned as a response.

On delete backfill ticket, the NotFound from `statestore` is also returned as a 404 rather than logged and ignored.


lucy-yuan:
Not needed at the moment. Closing.

PR #1368, opened 2021-04-03
Revert "Sentinel fix"

syntxerror:
Reverts googleforgames/open-match#1367

need to revert due to unwanted changes to open-match/third_party/swaggerui/config.json


sawagh (approved):
since only the changes to the config.json are to be reverted, do you want to change just that instead of reverting the entire PR?

syntxerror:
I just did the revert once I caught the mistake after merging and thought it made the most sense is all since I would have to do another PR for the config.json. It is pretty much the same amount of PRs so didn't see the harm in either approach

PR #1394, opened 2021-07-13
Go client update

### LINKED ISSUE #1384: k8s go-client version is too old

syntxerror:

- Updates the [default GKE version ](https://cloud.google.com/kubernetes-engine/docs/release-notes#June_16_2021)
- Updates go-client version

Closes #1384


HazWard:
Open Match doesn't actually need Kubernetes so I think it would be better to have a separate `go.mod` file in [`tools/`](https://github.com/googleforgames/open-match/tree/main/tools) instead. It would allow you to more some of the dependencies on go-client from Open Match.

syntxerror:
@googlebot  I signed it!

go.mod
```diff
@@ -49,17 +47,15 @@ require (
...
 	google.golang.org/grpc v1.36.0
 	google.golang.org/protobuf v1.25.1-0.20201208041424-160c7477e0e8
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
-	k8s.io/api v0.0.0-20191004102349-159aefb8556b // kubernetes-1.14.10
-	k8s.io/apimachinery v0.0.0-20191004074956-c5d2f014d689 // kubernetes-1.14.10
-	k8s.io/client-go v11.0.1-0.20191029005444-8e4128053008+incompatible // kubernetes-1.14.10
+	k8s.io/api v0.21.2 // kubernetes-1.14.10
+	k8s.io/apimachinery v0.21.2 // kubernetes-1.14.10
+	k8s.io/client-go v0.21.2 // kubernetes-1.14.10
```
andrewgrundy, line 58: these comments are misleading

andrewgrundy (changes requested):
Although this does fix the problem (for now) I don't think k8s should be a dependency of open-match. Could the e2e tests and the tools each have their own mod files, then we can remove this dependency from open-match itself?

go.mod
syntxerror: I agree that e2e test should have their own mod files. I have been thinking of moving the e2e test out of the repo as they part of Open Match itself. We can provide it in another repo if necessary such as ecosystems. Was going to put the idea out there tomorrow during the contributors meeting

syntxerror:
Will close this PR and split into two. One for an upgraded GKE version and another for giving the e2e tests and tooling their own go.mod files

PR #1417, opened 2021-09-12
Using uuid instead time value to make unique matchId.

kongbong:
I did try tutorial matchmaker101 in open-match 1.2.0 and
I got many errors of "multiple match functions used same match_id".
```2021/09/12 03:52:37 Failed to fetch matches for profile mode_based_profile, got rpc error: code = Unknown desc = error(s) in FetchMatches call. syncErr=[error receiving match from synchronizer: rpc error: code = Unknown desc = error calling evaluator: multiple match functions used same match_id: "profile-mode_based_profile-time-2021-09-12T03:52:37.39-0"], mmfErr=[<nil>]```

I guess matchfunction is called concurrently.
matchId should be unique so I used uuid for that instead time.

Closes #

### DIFF
--- a/tutorials/matchmaker101/frontend/main.go
+++ b/tutorials/matchmaker101/frontend/main.go
@@ -63,9 +63,12 @@ func main() {
 // once it has an assignment.
 func deleteOnAssign(fe pb.FrontendServiceClient, t *pb.Ticket) {
 	for {
+		time.Sleep(time.Second * 1)
+
 		got, err := fe.GetTicket(context.Background(), &pb.GetTicketRequest{TicketId: t.GetId()})
 		if err != nil {
-			log.Fatalf("Failed to Get Ticket %v, got %s", t.GetId(), err.Error())
+			log.Printf("Failed to Get Ticket %v, got %s", t.GetId(), err.Error())
+			continue
 		}
 
 		if got.GetAssignment() != nil {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tutorials/matchmaker101/solution/matchfunction/mmf/matchfunction.go [hunk 1]
```diff
@@ -17,7 +17,7 @@ package mmf
 import (
 	"fmt"
 	"log"
-	"time"
+	"github.com/google/uuid"
```
andrewgrundy, line 20: open match uses this UUID library: github.com/rs/xid. It probably makes sense to be consistent.

andrewgrundy (changes requested):

tutorials/matchmaker101/solution/matchfunction/mmf/matchfunction.go
syntxerror: Agreed. @kongbong Can you make use of that library uuid library and we can close this out

kongbong: I changed it.

tutorials/matchmaker101/frontend/main.go
```diff
@@ -63,17 +63,19 @@ func main() {
 // once it has an assignment.
 func deleteOnAssign(fe pb.FrontendServiceClient, t *pb.Ticket) {
 	for {
+		time.Sleep(time.Second * 1)
```
syntxerror, line 66: Reason for adding this delay?

kongbong: sorry, I think this is added accidentally.

Issue #1227, opened 2020-06-12
Use Go 1.5's GOMODCACHE to improve docker build times


Laremere:
Currently, the base build docker image performs multiple copies to improve build speed:
1. Copy over the go.mod and go.sum files.
2. go get
3. copy over the rest of the files
4. <build various binaries here>

This is done so that with docker layer caching, making a small code change doesn't result in it needing to download all dependencies.  However whenever the dependencies do change, it does require a full re-download.  Plus you need to run go commands locally to get the go.mod / go.sum files to update.

Go 1.5 is adding an environmental variable which allows specifying where the go mod cache is located.

I think the current double copy can be removed, and replaced with:
In the docker build commands, specify a mount of the host machine's go mod cache, and set the environment variable for the go mod cache to that mount (say, "/gomodcache")

This makes it a little more dependent on the host machine, which isn't great for docker.  However it simplifies the docker files, plus go.sum is used to perform a checksum of the go mod cache files, so given the immutability here I'm not worried, and think the tradeoffs are worth it.

mridulji:
Looks interesting...🙂 and I'm actually new to it.
As I think, there can be two aspects of reducing docker build times.
1. In building image at developer's machine. It may be achieved by mounting disk location at the build step in docker file. It will be useful in enhancing developer's experience.
```
RUN
    --mount=target=. \
    --mount=type=cache,target=/root/.cache/go-build \
    .....
```
2. the other one is in CI,

@Laremere  please help me by letting me know if I'm getting this right.

PR #1435, opened 2022-01-16
docker build optimization by using mount cache for go dependencies

mridulji:
Reduces docker CMD build times by mounting cache for go dependencies. 

#1227 

Closes #

It requires docker 19.03 version or later which is not very much new (released on 2019-07-22) and latest verision is 20.10 so I hope it works for all.

https://www.docker.com/blog/containerize-your-go-developer-environment-part-1/#:~:text=As%20we%20are,in%20your%20environment.

Reference :- https://www.docker.com/blog/containerize-your-go-developer-environment-part-2/

### DIFF
--- a/Dockerfile.cmd
+++ b/Dockerfile.cmd
@@ -18,7 +18,9 @@ WORKDIR /go/src/open-match.dev/open-match
 
 ARG IMAGE_TITLE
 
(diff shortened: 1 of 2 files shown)


syntxerror (approved):
LGTM

PR #1437, opened 2022-01-27
Using uuid instead time value to make unique matchId.

kongbong:
I did try tutorial matchmaker101 in open-match 1.2.0 and
I got many errors of "multiple match functions used same match_id".
2021/09/12 03:52:37 Failed to fetch matches for profile mode_based_profile, got rpc error: code = Unknown desc = error(s) in FetchMatches call. syncErr=[error receiving match from synchronizer: rpc error: code = Unknown desc = error calling evaluator: multiple match functions used same match_id: "profile-mode_based_profile-time-2021-09-12T03:52:37.39-0"], mmfErr=[<nil>]

I think matchfunction is called concurrently.
matchId should be unique so I used uuid for that instead time.

Closes #

I closed old pull request and make this one as old one has other changes.
https://github.com/googleforgames/open-match/pull/1417


jmrapp1:
I also run into this error and trying this solution has fixed the problem. +1 on getting this merged in

syntxerror (approved):
LGTM

PR #1438, opened 2022-02-03
AssignTickets empty check and test cases added

### LINKED ISSUE #1266: Unhelpful '500' failure when assigning an empty list of tickets
**What happened**:

Due to a local bug in my code, our Director sent an [`AssignTicketsRequest`](https://open-match.dev/site/docs/reference/api/#openmatch.AssignTicketsRequest) via REST to the backend containing an empty list of tickets.

We got a 500 error back from the RPC gateway without any useful information.

The om-backend logs contained only:
```
time="2020-10-09T12:38:09Z" level=error msg="failed to update assignments" app=openmatch component=app.backend error="ERR wrong number of arguments for 'mget' command"
time="2020-10-09T12:38:09Z" level=error msg="failed to update assignments for requested tickets" app=openmatch component=app.backend error="ERR wrong number of arguments for 'mget' command"
```

Either:
* No tickets were assigned, and life continues on; or
* A failure indicating that 'no tickets' is not a rational `AssignTicketsRequest`.

In this case, the latter would have been more useful, and pointed me towards the real issue; but the former is quite acceptable, and fits the more-general use-case.

Untested: Add a test similar to [`TestEmptyReleaseTicketsRequest`](https://github.com/googleforgames/open-match/blob/6f05e526fb258d390fb277de0d53623ce5194316/internal/testing/e2e/ticket_test.go#L190-L202) to test an empty `AssignTicketsRequest`.

The actual error message being logged is because we call the redis `mget` command with the ticket list, but `mget` *fails* when given an empty array, e.g., https://github.com/NodeRedis/node-redis/issues/125.

I would suggest either this case be caught and short-cut to "success" before passing down to Redis, or the API be respecified to be 1-or-more ticket IDs so that code-gen won't let me call this sort of thing, and the server can send a 400-series error instead, so I know the call was incorrect.

(I'm not sure how the GRPC/REST bridge works, so I don't know if perhaps you would get a 400-series error automatically from the definition, or if perhaps you cannot actually specify "1-or-more ticket IDs" in the API spec)

```
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.11", GitCommit:"d94a81c724ea8e1ccc9002d89b7fe81d58f89ede", GitTreeState:"clean", BuildDate:"2020-03-12T21:08:59Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.9", GitCommit:"94f372e501c973a7fa9eb40ec9ebd2fe7ca69848", GitTreeState:"clean", BuildDate:"2020-09-16T13:47:43Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

Bare-metal Kubernetes cluster.

Open Match master, not sure of the exact date. 

**Install Method(yaml/helm):**:

Helm subchart.

mridulji:
**What this PR does / Why we need it**: When making a call to `AssignTicketsRequest`, To check at least one of the `AssignmentGroup` have at least one `ticket_ids` present such that redis command doesn't throw error.
```
time="2020-10-09T12:38:09Z" level=error msg="failed to update assignments" app=openmatch component=app.backend error="ERR wrong number of arguments for 'mget' command"
time="2020-10-09T12:38:09Z" level=error msg="failed to update assignments for requested tickets" app=openmatch component=app.backend error="ERR wrong number of arguments for 'mget' command"
```

To prevent the code execution from going into this situation, we can check the length of the array containing all `ticket_ids`, so that in case `AssignmentGroup` doesn't contain any value in `ticket_ids` will return error `codes.InvalidArgument`
 
**Which issue(s) this PR fixes**: #1266 

Closes #

unit test case condition added: `TestUpdateAssignments`

> description: "empty ticketIds, error expected"
> errMessage:         "AssignmentGroupTicketIds is empty"
> 

e2e test case added: `TestAssignTicketsEmpty`

### DIFF
--- a/internal/statestore/ticket.go
+++ b/internal/statestore/ticket.go
@@ -255,6 +255,10 @@ func (rb *redisBackend) UpdateAssignments(ctx context.Context, req *pb.AssignTic
 		}
 	}
 
+	if len(idsI) == 0 {
+		return nil, nil, status.Error(codes.InvalidArgument, "AssignmentGroupTicketIds is empty")
+	}
+
 	ticketBytes, err := redis.ByteSlices(redisConn.Do("MGET", idsI...))
 	if err != nil {
 		return nil, nil, err
--- a/internal/statestore/ticket_test.go
+++ b/internal/statestore/ticket_test.go
@@ -229,6 +229,23 @@ func TestUpdateAssignments(t *testing.T) {
 			},
 		},
 		{
+			description: "empty ticketIds, error expected",
+			request: &pb.AssignTicketsRequest{
+				Assignments: []*pb.AssignmentGroup{
+					{
+						TicketIds:  []string{},
+						Assignment: &pb.Assignment{Connection: "2"},
+					},
+				},
+			},
+			expected: expected{
+				resp:               nil,
+				errCode:            codes.InvalidArgument,
+				errMessage:         "AssignmentGroupTicketIds is empty",
+				assignedTicketsIDs: nil,
+			},
+		},
+		{
 			description: "nil assignment, error expected",
 			request: &pb.AssignTicketsRequest{
 				Assignments: []*pb.AssignmentGroup{
--- a/internal/testing/e2e/ticket_test.go
+++ b/internal/testing/e2e/ticket_test.go
@@ -74,6 +74,27 @@ func TestAssignTickets(t *testing.T) {
 	require.Equal(t, "b", get.Assignment.Connection)
 }
 
+// TestAssignTicketsEmpty covers calls to assign when empty TicketIds
+func TestAssignTicketsEmpty(t *testing.T) {
+	om := newOM(t)
+	ctx := context.Background()
+
+	req := &pb.AssignTicketsRequest{
+		Assignments: []*pb.AssignmentGroup{
+			{
+				TicketIds:  []string{},
+				Assignment: &pb.Assignment{Connection: "a"},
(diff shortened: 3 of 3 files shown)


andrewgrundy (approved):
looks good to me

internal/statestore/ticket.go
```diff
@@ -255,6 +255,10 @@ func (rb *redisBackend) UpdateAssignments(ctx context.Context, req *pb.AssignTic
 		}
 	}
 
+	if len(idsI) == 0 {
+		return nil, nil, status.Error(codes.InvalidArgument, "AssignmentGroupTicketIds is empty")
```
TBBle, line 259: Isn't this test only going to match if _all_ `AssignmentGroup`s had empty `TicketIds`? It would be clearer if the message referred to that, as the closest I can see to `AssignmentGroupTicketIds` would be `AssignmentGroup.TicketIds` which would be an error appropriate for _any_  `AssignmentGroup`s having empty `TicketIds`.

mridulji: since variable `idsI` contains all `TicketIds` so this check will prevent from throwing error when there is not a single `TicketIds` is present and if any `AssignmentGroup`s having empty `TicketIds` will be silently ignored.

PR #1457, opened 2022-04-27
Fix tutorial images not building due to locally specified open-match version

narthur157:
Specifies the tutorial project images to use the remote open-match versions instead of the local project one
Fixes the issue I had described here: https://github.com/googleforgames/open-match-docs/issues/252
Similar to https://github.com/googleforgames/open-match/pull/1220
Closes https://github.com/googleforgames/open-match-docs/issues/252.

### DIFF
--- a/tutorials/matchmaker101/director/Dockerfile
+++ b/tutorials/matchmaker101/director/Dockerfile
@@ -15,8 +15,10 @@
 FROM golang:alpine as go
 WORKDIR /app
 ENV GO111MODULE=on
-
+RUN apk add --no-cache git
 COPY . .
+RUN go mod edit -replace open-match.dev/open-match@v0.0.0-dev=open-match.dev/open-match@latest
+RUN go mod tidy
 RUN go build -o director .
 
 CMD ["/app/director"]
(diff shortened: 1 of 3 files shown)


tutorials/matchmaker101/director/Dockerfile
```diff
@@ -15,8 +15,10 @@
 FROM golang:alpine as go
 WORKDIR /app
 ENV GO111MODULE=on
-
+RUN apk add --no-cache git
 COPY . .
+RUN go mod edit -replace open-match.dev/open-match@v0.0.0-dev=open-match.dev/open-match@latest
```
syntxerror, line 20: By adding the latest tag you're pulling the most recent versions image (e.g 1.4.0-rc.1). For the main branch we try to avoid this and ask devs to switch to the version branch of their choosing. If there is a larger reason for this to happen, happy to listen to your suggestions/concerns.

narthur157: No reason - is there a version you'd suggest? Or a tag that represents something like "stable"

PR #1475, opened 2022-07-07
Added persistent field to store any config/metadata in ticket and backfill

### LINKED ISSUE #1426: assignment should not overwrite ticket metadata during backfill
I use ticket for metadata storage (Extensions field). When ticket gets  assignment via back, it lost is metadata field. I put metadata from back fill into assignment. But I want ticket metadata in Extensions still to be retained, if it is not overwritten by Extensions from Assignment.

1. Create ACK with Assignment and ACK.
2. Get ticket after Assignment.
Expected: All Extensions of ticket retained
Actual: Ticket looses its metadata in Extensions.

Use some other storage to store ticket metadata. 

**Additional context**
I assume assignment should not clear metadata.

mridulji:
**What this PR does / Why we need it**: The data in extension field is overwrite during assignment. So to store some config data which necessarily needs to persist , `*pb.Backfill.PersistentField` is added which will could be changed by invoking `FrontendService.UpdateBackfill`.

Closes #1426 


syntxerror (approved):

eltone:
I know this is an old PR but FYI this change modifies tag numbers which goes against protobuf schema best practise. This would break deserialisation for old messages.

PR #1476, opened 2022-07-13
Update WatchAssignment function

### LINKED ISSUE #1472: Confusion in the codes of WatchAssignments function
Here is the codes of WatchAssignments:
```golang
func (s *frontendService) WatchAssignments(req *pb.WatchAssignmentsRequest, stream pb.FrontendService_WatchAssignmentsServer) error {
	ctx := stream.Context()
	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		default:
			sender := func(assignment *pb.Assignment) error {
				return stream.Send(&pb.WatchAssignmentsResponse{Assignment: assignment})
			}
			return doWatchAssignments(ctx, req.GetTicketId(), sender, s.store)
		}
	}
}
```

Here is my question: Since the loop will quit in the first loop anyway, what's the meaning of the "for" here?

mridulji:
**What this PR does / Why we need it**: This PR modifies the `FrontendService.WatchAssignment` function and listens to the `ctx.Done` functionality of GRPC stream in callback function so that in every retry of callback function, it is checked whether the user has terminated this API call or not.

Closes #1472 

- Updated the GKE version for regular channel support (i.e. 1.22.8-gke.202) [Reference](https://cloud.google.com/kubernetes-engine/docs/release-notes)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -636,7 +636,7 @@ delete-kind-cluster: build/toolchain/bin/kind$(EXE_EXTENSION) build/toolchain/bi
 create-cluster-role-binding:
 	$(KUBECTL) create clusterrolebinding myname-cluster-admin-binding --clusterrole=cluster-admin --user=$(GCLOUD_ACCOUNT_EMAIL)
 
-create-gke-cluster: GKE_VERSION = 1.21.11-gke.900 # gcloud beta container get-server-config --zone us-west1-a
+create-gke-cluster: GKE_VERSION = 1.22.8-gke.202 # gcloud beta container get-server-config --zone us-west1-a
 create-gke-cluster: GKE_CLUSTER_SHAPE_FLAGS = --machine-type n1-standard-8 --enable-autoscaling --min-nodes 1 --num-nodes 6 --max-nodes 10 --disk-size 50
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -636,7 +636,7 @@ delete-kind-cluster: build/toolchain/bin/kind$(EXE_EXTENSION) build/toolchain/bi
 create-cluster-role-binding:
 	$(KUBECTL) create clusterrolebinding myname-cluster-admin-binding --clusterrole=cluster-admin --user=$(GCLOUD_ACCOUNT_EMAIL)
 
-create-gke-cluster: GKE_VERSION = 1.21.11-gke.900 # gcloud beta container get-server-config --zone us-west1-a
+create-gke-cluster: GKE_VERSION = 1.22.8-gke.202 # gcloud beta container get-server-config --zone us-west1-a
```
syntxerror, line 639: I would not update the default cluster as part of this PR and do it in a separate PR

syntxerror: I apologize. This version is no longer within the stable release channel. See [here](https://cloud.google.com/kubernetes-engine/docs/release-notes-stable#May_26_2022)

syntxerror (approved):
LGTM

PR #1481, opened 2022-08-08
shifted e2e tests to project root

### LINKED ISSUE #1302: Making e2e testing packages public
When developing a library with Open Match, we want to do automated E2E testing. However, the package for e2e testing in Open Match is located in an internal package and cannot be used.

For reference, [Agones publishes an e2e testing package](https://github.com/googleforgames/agones/tree/master/test/e2e) that can be used externally, and it would be great if Open Match could do the same.

### LINKED ISSUE #1448: Remove dependencies for e2e tests by creating independent go mod files.
**What happened**:
An issue (#1384) identified the go-client is out of date but it was identified that go-client is only used for tooling. After discussion it would be best to move all dependencies for e2e test such as go-client into a separate go mod file since Open Match does not depend of several of these.

mridulji:
**What this PR does / Why we need it**: This PR separates the testing folder from other important implementations of open-match core and shifts it to project root location so that this package can have it's own go.mod file to manage its dependencies and not strictly linked with root's go.mod file.  

Closes #1448 
Closes #1302 


syntxerror (approved):
LGTM

PR #1490, opened 2022-09-04
Add open-match-override setting

kemurayama:

Currently, ConfigMap (open-match-override) should be installed via `--set open-match-override.enabled=true`.   

This PR enables users to set `open-match-override.enabled` with `values.yaml`.

Closes #

### DIFF
--- a/install/helm/open-match/templates/om-configmap-override.yaml
+++ b/install/helm/open-match/templates/om-configmap-override.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-{{- if index .Values "open-match-override" }}
+{{- if index .Values "open-match-override" "enabled" }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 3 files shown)


mridulji (commented):
I think you may also need to change [here](https://github.com/googleforgames/open-match/blob/main/install/helm/open-match/templates/om-configmap-override.yaml#L15) to mention `"enabled"`. So it will be 

`{{- if index .Values "open-match-override" "enabled" }}`

But are you able to successfully install open-match (by yaml/helm both ways) by the changes you made 🤔 ?

mridulji:
Hey @kemurayama, Please let us know if you are able to test the [suggested changes](https://github.com/googleforgames/open-match/pull/1490#pullrequestreview-1170615816) or if there is something you need help with.

kemurayama:
Hi @mridulji, 

I've confirmed that I needed to add "enabled" in `om-configmap-override.yaml` as you mentioned to install open-match-configmap-override.

mridulji (approved):
LGTM👍🏻

PR #1505, opened 2022-10-31
Fix: struct fields order revert

mridulji:
**What this PR does / Why we need it**: In a recent feature request, we added a field in messages protobuf file and changed the order of the field which may result in need to update integration with open-match. By following the [documentation](https://developers.google.com/protocol-buffers/docs/proto), we are reverting back to its previous order and new field is assigned with next number in the sequence.

Closes #


andrewgrundy:
This seems like a bad idea. The previous breaking change is already out there, by doing this Open Match will have introduced another breaking change, which feels worse than just leaving it alone.

PR #1506, opened 2022-10-31
Specify HPA for individual service

### LINKED ISSUE #1390: Be able to config HPA for each individual service

The current helm chart allows to config HPA in the global level, which means all Open Match services inherit from the same global HAP configuration.

In our use case, we may generate huge amount of requests to Open Match Query service in a short period of time, which overloads the existing Open Match Query service pods. Below is the output of the `kubectl top pods`

```
kepler-loadtest-open-match-query-cd55b976-c5jch            1016m        631Mi
kepler-loadtest-open-match-query-cd55b976-vz5x5            466m         696Mi
kepler-loadtest-open-match-query-cd55b976-w4kqk            993m         501Mi
```

The existing pods are killed for OOM. I can see that some new pods are started by HPA, but it takes time to start up new pods. And the longer it takes, the more workload our system will generate. I can see the pods can no longer start up like below.

```
kepler-loadtest-open-match-query-cd55b976-455ld                   0/1     OOMKilled           1          3m10s
kepler-loadtest-open-match-query-cd55b976-bmgjw                   0/1     Running             1          3m10s
kepler-loadtest-open-match-query-cd55b976-c5jch                   1/1     Running             2          26m
kepler-loadtest-open-match-query-cd55b976-pbj5x                   0/1     CrashLoopBackOff    1          2m55s
kepler-loadtest-open-match-query-cd55b976-vhdxw                   1/1     Running             1          3m10s
kepler-loadtest-open-match-query-cd55b976-vz5x5                   0/1     CrashLoopBackOff    3          26m
kepler-loadtest-open-match-query-cd55b976-w4kqk                   0/1     OOMKilled           2          26m
kepler-loadtest-open-match-query-cd55b976-xrnkn                   0/1     Running             1          2m55s
```

We want to be able to configure HPA for each individual service, so that we can set the minimum pod number of Open Match Query service to be 10 while others still keep the value as 3. We want to reserve enough pods for Open Match query service to handle some pulse workload while other services do not need to keep that many pods to save cost.

Be able to overwrite the default global HPA configuration for each individual Open Match service using value.yaml in the helm chart.

**Additional context**

mridulji:
**What this PR does / Why we need it**: This PR implements a feature to provide the functionality of maintaining the HPA for the microservices individually rather than a general config applicable for all microservices so that minimum number of replicas could be set higher than others for the some particular microservice. In the concerned issue mentioned, There is need to keep minimum replicas of query microservice ready to serve.

Closes #1390 

### DIFF
--- a/install/helm/open-match/subcharts/open-match-customize/templates/evaluator.yaml
+++ b/install/helm/open-match/subcharts/open-match-customize/templates/evaluator.yaml
@@ -54,7 +54,7 @@ spec:
     apiVersion: apps/v1
     kind: Deployment
     name: {{ include "openmatch.evaluator.hostName" . }}
-  {{- include "openmatch.HorizontalPodAutoscaler.spec.common" . | nindent 2 }}
+  {{- include "openmatch.HorizontalPodAutoscaler.evaluator.spec.common" . | nindent 2 }}
 ---
 apiVersion: apps/v1
 kind: Deployment
--- a/install/helm/open-match/subcharts/open-match-customize/templates/matchfunctions.yaml
+++ b/install/helm/open-match/subcharts/open-match-customize/templates/matchfunctions.yaml
@@ -54,7 +54,7 @@ spec:
     apiVersion: apps/v1
     kind: Deployment
     name: {{ include "openmatchcustomize.function.hostName" . }}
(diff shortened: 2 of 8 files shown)


syntxerror (approved):
LGTM

PR #1508, opened 2022-11-04
Default values of configs

### LINKED ISSUE #1013: Refactor internal/config package and have it checks the necessary config values
This issue tracks the progress on refactoring the `internal/config` package. Our current config package directly read the config files from viper and wont check what fields are missing at all. It would be great if we could refactor/add a tool to check what fields are missing from the config and document the default config values for the optional fields instead.

mridulji:
**What this PR does / Why we need it**:  This PR sets the default values of various configs in matchmaker_config_override.yaml file of open-match override configmap to ensure smooth start-up of all deployments in case any config is missing or omitted. Below is the list of default values of respective configs:

- registrationInterval: 1s
- proposalCollectionInterval: 10s
- pendingReleaseTimeout: 1m
- assignedDeleteTimeout: 10m
- queryPageSize: 1000
- backfillLockTimeout: 1m

Closes #1013 

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -58,6 +58,30 @@ func Read() (*viper.Viper, error) {
 		return nil, fmt.Errorf("fatal error reading override config file, desc: %s", err.Error())
 	}
 
+	if !cfg.IsSet("registrationInterval") {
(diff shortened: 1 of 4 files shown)


syntxerror (approved):
LGTM

PR #1530, opened 2023-01-18
Fix: add Content-Type and Transfer-Encoding to matchfunction:run POST request

seesemichaelj:

The HTTP POST request to run the Match Function (`/v1/matchfunction:run`) currently has no `Content-Type` header on it. This prevents parsing utils (such as https://www.npmjs.com/package/body-parser) from automatically being able to parse the JSON body. This PR adds the missing `Content-Type` header as well as the `Transfer-Encoding` header.

There is no associated issue for this.

Here is the Slack conversation about this issue: https://open-match.slack.com/archives/CCD3H0L2D/p1673486707254609


mridulji (approved):
LGTM 👍🏻

PR #1540, opened 2023-02-21
Removing deprecation PSP

### LINKED ISSUE #1449: PodSecurityPolicy deprecated in k8s 1.25
**What happened**:
Installing on GKE using the currently supported k8s versions displays a deprecation notice for this resource type.

**What you expected to happen**
Need to update these to use the new PodSecurity admission controller.
https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement

Start a new GKE cluster, follow the OM installation instructions.

GKE (although any platform using k8s will display this, it's a k8s deprecation)

Latest (altthough affects previous versions too)

**Install Method(yaml/helm):**:
Both

mridulji:
**What this PR does / Why we need it**: This PR contains the changes required for removing PodSecurityPolicy been deprecated since v1.21 of kubernetes and removed from v1.25+. So in order to be able to deploy open-match on GKE with version 1.25 or up, this change is necessary. 

As a solution this is in-built Pod Security Admission helps us to enforce such security. When following the [documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/migrate-podsecuritypolicy) for migration of PSP to PSA, we are able to enforce baseline level of security using [pspmigrator](https://github.com/kubernetes-sigs/pspmigrator).

Closes #1449 

- Respective changes tested on GKE version 1.25.5-gke.2000


joeholley (approved):
LGTM

benji56:
I want to install it in Digital ocean. The version is 1.25.4-do.0. Unfortunately, whatever I do I get this error. What is the problem? I tried installing with helm and also with the 1.7.0 yaml file. Unfortunately the result is the same. 

`helm install open-match --create-namespace --namespace open-match open-match/open-match
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: [resource mapping not found for name: "open-match-redis-podsecuritypolicy" namespace: "open-match" from "": no matches for kind "PodSecurityPolicy" in version "policy/v1beta1"
ensure CRDs are installed first, resource mapping not found for name: "open-match-core-podsecuritypolicy" namespace: "open-match" from "": no matches for kind "PodSecurityPolicy" in version "policy/v1beta1"
ensure CRDs are installed first]`

mridulji:
Hi @benji56, may I know which version of OM are you trying to install. Because till 1.7.0 version we are unable to support on kubernetes version 1.25+. However this PR's changes will be in 1.8.0 release so it would be possible from next versions onward. Please keep following [open-match-discuss](https://groups.google.com/g/open-match-discuss) group to get the latest updates.

PR #1578, opened 2023-06-12
Unlock redsync locks using context.Background()

### LINKED ISSUE #1576: Backfill mutex not unlocked when context is canceled in `CleanupBackfills`
**What happened**:

A backfill stopped being acknowledged since it was no longer needed. After a while it expired and was scheduled for cleanup. The `CleanupBackfills` routine was interrupted through the context being cancelled. This led to a mutex lock on the backfill not being unlocked, until the `backfillLockTimeout` was reached.

Within this period the game server needed backfilling again started trying to re-acknowledge the backfill, but ran into continuous errors of `rpc error: code = Unknown desc = redsync: failed to acquire lock` until the lock timeout was reached and we finally got a response that it had expired. Only then could it react to that state and trigger creation of a new backfill. This is in part an issue with our current implementation, which is quite early stage, and we don't implement deleting of backfills right now and just let them expire when they're not needed.

However I do believe there's a bug with the mutex handling in `DeleteBackfillsCompletely` where the defer func to release the lock is using the same context that is provided, which in the case of a cancellation will be canceled, so I think it will never properly release the lock in that case: https://github.com/googleforgames/open-match/blob/main/internal/statestore/backfill.go#L245

It might be better to give it `context.Background()` or a new context with a timeout.

Logs:
image
image

Cancelling the context should not leave mutexes in a locked state.

Not quite sure what caused the context cancellation in the first place, looks like it happens when all the callers (director FetchMatches calls) are done.

Client Version: v1.27.1
Kustomize Version: v5.0.1
Server Version: v1.24.11-gke.1000

GKE

1.5.0

**Install Method(yaml/helm)**:

Helm to generate yaml applied with kubectl.

glindstedt:

Context: https://github.com/googleforgames/open-match/issues/1576

We observed an issue where `CleanupBackfills` would leave backfill mutex locks in a locked state if the context was canceled, which prevented us from doing anything with the backfill until the `backfillLockTimeout`. Even if the backfill had expired we had to wait for the timeout to find out about it.

This PR ensures that the deferred `Unlock` is performed using a background context, so if the original context was canceled we will still try to unlock the mutex.

Closes #1576

### DIFF
--- a/internal/app/backend/backend_service.go
+++ b/internal/app/backend/backend_service.go
@@ -388,7 +388,7 @@ func createOrUpdateBackfill(ctx context.Context, backfill *pb.Backfill, ticketId
 	}
 
 	defer func() {
-		_, unlockErr := m.Unlock(ctx)
+		_, unlockErr := m.Unlock(context.Background())
 		if unlockErr != nil {
 			logger.WithFields(logrus.Fields{"backfill_id": backfill.Id}).WithError(unlockErr).Error("failed to make unlock")
 		}
--- a/internal/app/frontend/frontend_service.go
+++ b/internal/app/frontend/frontend_service.go
@@ -171,7 +171,7 @@ func (s *frontendService) UpdateBackfill(ctx context.Context, req *pb.UpdateBack
 		return nil, err
 	}
 	defer func() {
-		if _, err = m.Unlock(ctx); err != nil {
+		if _, err = m.Unlock(context.Background()); err != nil {
 			logger.WithError(err).Error("error on mutex unlock")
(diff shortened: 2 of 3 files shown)


joeholley:
Note: Backfills is a community-contributed feature, so we just merge PRs related to it into the next version provided tests pass and there's no obvious issues with the code.

This will be in 1.8. Thanks for the contribution!

joeholley (approved):

joeholley:
Manually merged into release-1.8 branch commit 74d84a26

Issue #1601, opened 2023-10-05
add ability to set uo log output 


lvoby:
**Is your feature request related to a problem? Please describe.**
Add ability to set up logging output. We run project on google cloud and we use google log explorer for logging collection. When log appear in stderr it marks like severity:ERROR

**Describe the solution you'd like**
By default golang log write to stderr. To change output we can use log.SetOutput

**Describe alternatives you've considered**
I'm open for any other solutions, thanks

![image](https://github.com/googleforgames/open-match/assets/25271538/64a28811-3195-44d1-8b5e-8e581aba14fb)

PR #1602, opened 2023-10-05
add ability to set up log output

lvoby:
Simple solution to set up log output.
In most cases we don't need this, so where will be only two options (stderr/stdout)

**Which issue(s) this PR fixes**: https://github.com/googleforgames/open-match/issues/1601

Closes #


simao-silva:
Any expected deadline for this?

PR #1674, opened 2023-11-07
Skip expired backfills instead of returning error

glindstedt:

If a single backfill has expired it would cause the whole FetchMatches call to error out. Now instead the backfill will be skipped, similar to the `NotFound` case.

The status code returned for attempting to acknowledge or update an expired backfill is changed from Unavailable since the operation is not retryable. I opted for `FailedPrecondition`, but maybe `InvalidArgument` could make more sense since you can't "revive" the backfill, only create a new one  if it expired.

There are some formatting changes from format-on-save, let me know if I should remove them to make the PR clearer.


testing/e2e/backfill_test.go
```diff
@@ -613,6 +616,62 @@ func TestBackfillSkipNotfoundError(t *testing.T) {
...
+	om.SetEvaluator(func(ctx context.Context, in <-chan *pb.Match, out chan<- string) error {
+		p, ok := <-in
+		require.True(t, ok)
+
+		out <- p.MatchId
+		return nil
+	})
+
+	// Depends on pendingReleaseTimeout
+	time.Sleep(1 * time.Second)
```
glindstedt, line 657: This is a bit ugly, would be nice if we could have something like [k8s.io/utils/clock/testing](https://pkg.go.dev/k8s.io/utils/clock/testing) to pass the time

internal/app/backend/backend_service.go
```diff
@@ -183,7 +183,7 @@ func synchronizeRecv(ctx context.Context, syncStream synchronizerStream, m *sync
 				err = createOrUpdateBackfill(ctx, backfill, ticketIds, store)
 				if err != nil {
 					e, ok := status.FromError(err)
-					if err == errBackfillGenerationMismatch || (ok && e.Code() == codes.NotFound) {
+					if err == errBackfillGenerationMismatch || (ok && (e.Code() == codes.NotFound || e.Code() == codes.FailedPrecondition)) {
```
glindstedt, line 186: This is the main change

lonefreak (approved):

PR #1687, opened 2024-02-20
fix metrics for backfills per query

qua-yanase:
There is currently an incorrect measure set in the "open-match.dev/query/backfills_per_query" view. This is the fix.

Closes #


joeholley (approved):
LGTM

govargo:
Thank you for your contribution!!

PR #1789, opened 2025-12-02
Release tickets after backfill assignments

fumlig:
After tickets are assigned to new matches, they are deindexed and also released from the pending release set.

After backfill assignments, tickets are deindexed but never released. This has led to us accumulating ticket ids in the pending release set and therefore leaking memory.


fumlig:
Mistakenly opened in the wrong repository

